PR #7, opened 2017-01-12
2.11.1

edjubuh:
- Fix broken fault reporting (#1)
- Makefile can generate PROS templates
- Support for general purpose libraries in the firmware folder. All `firmware/*.a` files will get included in the linking process


edjubuh:
Need to verify `make template` works as intended before approval

include/watchdog.h
```diff
@@ -19,11 +19,11 @@
  * Public API Function to enable watchdog.
  * Must be called in initializeIO
  */
-void iwdgEnable();
+void watchdogEnable();
```
HotelCalifornia: Is this fine as `watchdogEnable` as opposed to `watchdogInit` (which conforms to the naming standards a little better)?

PR #17, opened 2017-06-20
Add standalone operational mode

HotelCalifornia:
This needs testing. I had a dubious rig set up, but then I lost the toolchain. Oh 🐳.

### DIFF
--- a/src/supervisor.c
+++ b/src/supervisor.c
@@ -43,6 +43,8 @@ static uint8_t svIndex;
 static uint8_t svPacket;
 // Team name to report when asked for configuration
 char svTeamName[8];
+// Initial SV flags
+char svInitFlags = 0x0;
 
 // Supervisor transfer complete, check the flags sent
 static INLINE void _svComplete() {
@@ -133,6 +135,11 @@ void _svNextByte() {
 	SPI1->DR = spiBufferTX[idx];
 }
 
+// standaloneModeEnable - enable standalone operation
+void standaloneModeEnable() {
+	svInitFlags = 0x1;
+}
+
 // svInit - Initialize supervisor communications
 void svInit() {
 	uint8_t i;
@@ -154,6 +161,7 @@ void svInit() {
 		spiBufferTX[i] = (uint16_t)0x0000;
 	SV_OUT->key = SV_MAGIC;
 	SV_OUT->mode = 2;
+	SV_OUT->flags = svInitFlags;
 	SV_OUT->version = 1;
 	// Load in empty motor values
 	svSetAllData((uint8_t)0x7F);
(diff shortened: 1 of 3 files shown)


HotelCalifornia:
Testing checklist:

- [ ] code runs in auto with tether
- [ ] code runs without waiting for vexnet (without tether)

src/supervisor.c
```diff
@@ -194,7 +194,7 @@ void svSynchronize() {
 	// HELLO message!
 	kwait(50);
 	print("\r\nPowered by PROS " FW_VERSION "\r\n" FW_DISCLAIMER
-		"\r\nPROS (C)2011-2014 Purdue ACM SIGBOTS\r\n");
+		"\r\nPROS (C)2011-2017 Purdue ACM SIGBOTS\r\n");
```
HotelCalifornia: I know it's not strictly related, but while we're here and editing `supervisor.c`...

src/supervisor.c
```diff
@@ -137,7 +137,9 @@ void _svNextByte() {
 
 // standaloneModeEnable - enable standalone operation
 void standaloneModeEnable() {
-	svInitFlags = 0x1;
```
edjubuh: This didn't work because svInit runs before initIO, so the actual TX SPI buffer never gets updated.

baylessj:
The final version of this feature implements standalone mode by setting a flag in the outgoing SPI communication buffer between the user and master processors to specify to the master processor that the controller should run in standalone mode. Because this process occurs during the `initializeIO()` function, there is a very brief moment of time between the SPI communication's initialize at boot-up and when the `initializeIO()` function runs where the controller has not yet entered standalone mode (evidenced by typical "searching for a VEXNet connection" lights), but as soon as the `initializeIO()` function runs then the VEXNet light drops out and the controller enters standalone mode.

Succesful test cases include:
 
- `standaloneModeEnable()` called in `initializeIO()`, running tethered.
- `standaloneModeEnable()` called in `initializeIO()`, running standalone (no VEXNet/USB connection)
- no call to `standaloneModeEnable()`, running tethered

edjubuh:
@baylessj, can you test no standalone mode with a joystick (one of USB A-A or VEXNet - don't care which)

HotelCalifornia:
Bonus points if you test them both

edjubuh:
[kernel-2.11.2.zip](https://github.com/purduesigbots/pros/files/1093394/kernel-2.11.2.zip)

To test, extract the files to `%AppData%\PROS\pros-mainline\kernel-2.11.2` such that the file `%AppData%\PROS\pros-mainline\kernel-2.11.2\template.pros` exists. For Linux and MacOS, the same applies but the folder is `~/.pros` instead of `%AppData%\pros`

src/supervisor.c
```diff
@@ -133,6 +133,12 @@ void _svNextByte() {
 	SPI1->DR = spiBufferTX[idx];
 }
 
+// standaloneModeEnable - enable standalone operation
+void standaloneModeEnable() {
+	// set flag in outgoing SPI buffer to enable standlone mode
+	SV_OUT->flags |= 1;
```
jbuschjr: Are we aware currently of what the remaining bits of this `flags` byte (I assume it to be `uint8_t`) do?

edjubuh: No, we'll have to take a look at the other bit values during the fall semester.

Issue #23, opened 2017-06-24
Broken links in readme


tehkhop:
Ironically, the links under "I think I found a bug!" are broken.

PR #24, opened 2017-06-24
fix links in readme (#23)

HotelCalifornia:
I swear this better be the last time I do this.

Also I deserve credit because I did this all from my phone.


HotelCalifornia:
@edjubuh you can squash all these if you want

Issue #5, opened 2016-12-31
C++ newlib compatibility


PixelToast:
PROS and newlib have a conflicting FILE symbol
`In file included from ../include/main.h:46:0,
                 from lcd.cpp:4:
../include/API.h:744:13: error: conflicting declaration 'typedef int FILE'
 typedef int FILE;
             ^
In file included from /usr/include/newlib/c++/5.4.1/cwchar:44:0,
                 from /usr/include/newlib/c++/5.4.1/bits/postypes.h:40,
                 from /usr/include/newlib/c++/5.4.1/bits/char_traits.h:40,
                 from /usr/include/newlib/c++/5.4.1/string:40,
                 from /usr/include/newlib/c++/5.4.1/stdexcept:39,
                 from /usr/include/newlib/c++/5.4.1/array:38,
                 from /usr/include/newlib/c++/5.4.1/tuple:39,
                 from lcd.cpp:1:
/usr/include/newlib/wchar.h:53:16: note: previous declaration as 'typedef __FILE FILE'
 typedef __FILE FILE;
                ^
`
this can be solved by either changing the name of `FILE` or changing all the `#define`s to use `static const` instead and putting all the pros API functions into a namespace

edjubuh:
Maintaining symbol compatibility with every library is quite an undertaking. The official PROS 2.0 kernel will respect libc and GCC constraints. I'll leave it open if someone from the community decides to modify the filesystem, but the PROS development team don't currently have plans to address this for PROS 2.0.

iuyte:
I fixed this with CTRL+F and replacing all instances of FILE with PROS_FILE in API.h, as per https://blog.pxtst.com/using-c-in-pros/.

PR #26, opened 2017-06-30
Add C++ compatability with FILE symbol

edjubuh:
FILE was renamed to PROS_FILE. If the FILE symbol wasn't defined, then backwards compatability is maintained with #def'ing FILE to PROS_FILE.

### DIFF
--- a/include/API.h
+++ b/include/API.h
@@ -748,7 +748,7 @@ typedef int PROS_FILE;
 /**
  * For convenience, FILE is defined as PROS_FILE if it wasn't already defined. This provides 
  * backwards compatability with PROS, but also allows libraries such as newlib to be incorporated
- * into PROS projects. If you're not using C++/newlib, you can use FILE as previously
+ * into PROS projects. If you're not using C++/newlib, you can disregard this and just use FILE.
  */
 #define FILE PROS_FILE
 #endif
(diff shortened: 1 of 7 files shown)


edjubuh:
RFC from @iuyte and/or @PixelToast for addressing #5

include/API.h
```diff
@@ -736,12 +736,23 @@ bool i2cWrite(uint8_t addr, uint8_t *data, uint16_t count);
...
  */
-typedef int FILE;
+typedef int PROS_FILE;
+
+
+#ifndef FILE
+/**
+ * For convenience, FILE is defined as PROS_FILE if it wasn't already defined. This provides 
+ * backwards compatability with PROS, but also allows libraries such as newlib to be incorporated
+ * into PROS projects. If you're not using C++/newlib, you can use FILE as previously
```
HotelCalifornia: I take it back, "as previously" sort of falls flat. Let's say

 > If you're not using C++/newlib, you can use FILE as [normal? as you would per the docs?]

I'm having trouble coming up with wording that I like.

edjubuh:
[kernel-2.12.0-beta1.zip](https://github.com/purduesigbots/pros/files/1114044/kernel-2.12.0-beta1.zip)

iuyte:
I am yet to run this, but it does compile with my C++ projects perfectly fine.

iuyte:
Can confirm this works :ok_hand:

include/API.h
iuyte: "If you're not using C++/newlib, you can use FILE as per usual"?

edjubuh:
Hmm.. GitHub didn't auto-close the PR. Well, it's merged into 2.12.0 so closing it

Issue #32, opened 2017-07-27
Ultrasonic sensor seems to return zero periodically


bradamiller:
#### Expected Behavior:
The VEX Ultrasonic sensor should always return a reasonable value for the distance
#### Actual Behavior:
I'm seeing a zero value about every 50ms in between the correct values.
#### Steps to reproduce:
I wrote this test program that gets the time of  a zero value and subtracts the last time we saw a zero
to look for a repeatable period:
```
 Ultrasonic ultra;
 const int leftMotor = 1;
 const int rightMotor = 10;

void operatorControl() {
		ultra = ultrasonicInit(4, 3);
		int lastTime = 0;
		while (1) {
			int distance = ultrasonicGet(ultra);
			if (distance == 0) {
				// if this is a zero range value
				unsigned long thisTime = millis();
				int time = thisTime - lastTime;  // time since last zero
				printf("millis: %d\n", time);
				while (ultrasonicGet(ultra) == 0); // wait for the distance to become nonzero
				lastTime = thisTime; // update the last time we saw a zero
			}
		}
}
```
#### System information:
mac os 10.12.6

Kernel Version:
2.12

HotelCalifornia:
Have you tried running this test with multiple ultrasonic sensors?

In our experience, these sensors have been the most troublesome to get working on the kernel side of things, so our first instinct is to chalk this up as a hardware issue.

bradamiller:
I had some students report that they had to exclude zero values so I tried it myself with a different robot and ended up writing that test program. I’ve seen those sensor return outlier values occasionally with other libraries, but the consistent 50ms thing made me think of some kind of race condition in the code.  I’ll try a few other sensors and see if the results are the same or it is a bad sensor.

I wrote the easyC library for the PIC vex controller and I remember those sensors being pretty problematic.

stcarlso:
https://github.com/purduesigbots/pros/blob/master/src/ultrasonic.c#L39 sets the value to zero when the sensor starts pinging, and https://github.com/purduesigbots/pros/blob/master/src/ultrasonic.c#L53 sets the correct value when the ping returns. Since the ping is broadcast every 50ms, checking the value repeatedly would be expected to return zero while the ping is outbound and the sensor waiting for the reply.

[...]

PR #34, opened 2017-08-13
Define return values for Ultrasonic error cases

baylessj:
In response to Issue #32 , this pull request defines two types of error values for the Ultrasonic sensor, one for when the ultrasonic sensor is polled while it is pinging and waiting for a value (ULTRASONIC_TIMEOUT), and one when the ultrasonic is reading a value that is outside of its range (ULTRASONIC_BEYOND_RANGE).

### DIFF
--- a/include/API.h
+++ b/include/API.h
@@ -659,8 +659,8 @@ typedef void * Ultrasonic;
 /**
  * Gets the current ultrasonic sensor value in centimeters.
  *
- * If no object was found, -2 (ULTRASONIC_BEYOND_RANGE) is returned. If the ultrasonic sensor is
- * polled while it is pinging and waiting for a response, -1 (ULTRASONIC_TIMEOUT) is returned.
+ * If no object was found or if the ultrasonic sensor is polled while it is pinging and waiting
+ * for a response, -1 (ULTRA_BAD_RESPONSE) is returned.
  * If the ultrasonic sensor was never started, the return value is undefined. Round and fluffy
  * objects can cause inaccurate values to be returned.
  *
@@ -752,7 +752,7 @@ typedef int PROS_FILE;
 
 #ifndef FILE
 /**
- * For convenience, FILE is defined as PROS_FILE if it wasn't already defined. This provides 
+ * For convenience, FILE is defined as PROS_FILE if it wasn't already defined. This provides
  * backwards compatability with PROS, but also allows libraries such as newlib to be incorporated
  * into PROS projects. If you're not using C++/newlib, you can disregard this and just use FILE.
  */
--- a/include/API.h
+++ b/include/API.h
@@ -645,6 +645,18 @@ void encoderReset(Encoder enc);
 void encoderShutdown(Encoder enc);
 
 /**
+ * This value is returned if the sensor is polled while the ultrasonic sensor's ping is outbound
+ * and the sensor is waiting for a response.
+ */
+#define ULTRASONIC_TIMEOUT -1
+/**
+ * This value is returned if the ultrasonic sensor cannot find a surface within its range.
+ *
+ * The VEX Ultrasonic Sensor's range is 3.81cm to 292.1cm.
+ */
+#define ULTRASONIC_BEYOND_RANGE -2
+
+/**
  * Reference type for an initialized ultrasonic sensor.
  *
  * Ultrasonic information is stored as an opaque pointer to a structure in memory; as this is a
@@ -654,9 +666,10 @@ typedef void * Ultrasonic;
 /**
  * Gets the current ultrasonic sensor value in centimeters.
  *
- * If no object was found, zero is returned. If the ultrasonic sensor was never started, the
- * return value is undefined. Round and fluffy objects can cause inaccurate values to be
- * returned.
+ * If no object was found, -2 (ULTRASONIC_BEYOND_RANGE) is returned. If the ultrasonic sensor is
+ * polled while it is pinging and waiting for a response, -1 (ULTRASONIC_TIMEOUT) is returned.
+ * If the ultrasonic sensor was never started, the return value is undefined. Round and fluffy 
+ * objects can cause inaccurate values to be returned.
  *
  * @param ult the Ultrasonic object from ultrasonicInit() to read
  * @return the distance to the nearest object in centimeters
--- a/include/API.h
+++ b/include/API.h
@@ -645,16 +645,9 @@ void encoderReset(Encoder enc);
 void encoderShutdown(Encoder enc);
 
 /**
- * This value is returned if the sensor is polled while the ultrasonic sensor's ping is outbound
- * and the sensor is waiting for a response.
+ * This value is returned if the sensor cannot find a reasonable value to return.
  */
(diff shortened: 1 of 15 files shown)


baylessj:
The changes were tested by polling a single ultrasonic in the operator control loop at the default time interval of 20 ms. 

The conversation in Issue #32 seemed to indicate that the sensor would return ULTRASONIC_TIMEOUT every 50ms or so, but this value was returned very rarely, roughly once per 10-15 seconds of use. ULTRASONIC_BEYOND_RANGE values were improperly returned in situations where there was an object within the sensor's range much more often than the ULTRASONIC_TIMEOUT values were returned, at roughly once per second. This improper return value might be the result of testing the sensor with surfaces that are do not reflect the sound perfectly, thus causing the sensor to fail to detect the objects.

This change made it much easier to identify the difference in error values and also recognize the source of error in ultrasonic data when debugging by printing the values to the terminal.

HotelCalifornia:
> This change made it much easier to identify the difference in error values and also recognize the source of error in ultrasonic data when debugging by printing the values to the terminal.
Do you have an example of the outputs for each case?

<sup>if only to satisfy my curiosity 😉</sup>

baylessj:
```
distance: -2
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: 26
distance: 26
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: -1
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 25
distance: 26
distance: 26
distance: 26
distance: 26
distance: -1
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: -1
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 26
distance: 27
distance: 27
distance: 27
distance: 27
distance: 27
distance: 27
distance: 27
distance: 29
distance: 29
distance: 29
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -2
distance: -1
distance: -1
distance: -2
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -2
distance: -1
distance: -1
distance: -2
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
distance: -1
distance: -2
distance: -1
distance: -1
distance: -1
```

Well it would appear that my recollection of my previous data collection was incorrect (or the ultrasonic has started behaving differently since then, but that's unlikely), and the timeout values were returned every now and then, even when a flat surface was placed in front of it.

The test was conducted in the same manner as before, with the ultrasonic returning values every 20ms in the operator control loop. There was no surface in the ultrasonic's range at first, then a flat surface was introduced (the lid to the box for the Battleship board game, fwiw), and then the flat surface was removed. One particularly interesting thing to note is that the ping timed out more often than it returned the ULTRASONIC_BEYOND_RANGE value, but it's still clear that at that point in time it was garbage data because the Ultrasonic sensor was out of range.

include/API.h
```diff
@@ -654,9 +659,10 @@ typedef void * Ultrasonic;
 /**
  * Gets the current ultrasonic sensor value in centimeters.
  *
- * If no object was found, zero is returned. If the ultrasonic sensor was never started, the
- * return value is undefined. Round and fluffy objects can cause inaccurate values to be
- * returned.
+ * If no object was found, -2 (ULTRASONIC_BEYOND_RANGE) is returned. If the ultrasonic sensor is
+ * polled while it is pinging and waiting for a response, -1 (ULTRASONIC_TIMEOUT) is returned.
```
edjubuh: This documentation needs to be updated

baylessj:
Updated the pros-docs documentation in commit purduesigbots/pros-docs@a5cb8854d452afa3fd9fa32037c417664a8317fc

baylessj:
Previous commit to master was reverted, the changes can now be found in purduesigbots/pros-docs#26

Octogonapus:
Inside of `fireUltrasonic` why do we write `ULTRA_BAD_RESPONSE` to `value` before the ping? The best guess we can make for the next ping is the value of the previous ping, not an invalid response. What about not writing to `value` at all before the ping?

baylessj:
That point of debate was covered more thoroughly in Issue #32 , where the case of an object moving from the sensor's range to beyond its range was considered. If a default value is set, then the user would be aware that the object had moved out of the ultrasonic's range by the first ping that was not properly returned, but if the old value is kept, then it would appear as though the object never moved out of the ultrasonic's range. 

Keeping the old value removes the need for the user to filter the ultrasonic's returned values, but at the cost of not providing all of the available information about what the ultrasonic is observing.

Octogonapus:
Okay, if we are going with that then it all looks good 👍

HotelCalifornia:
Just want to get another test in, with a new cortex and a different ultrasonic sensor so we can confirm this will work properly

Octogonapus:
Tested using this code: https://hastebin.com/amusegonoc.cpp
Got this output: https://hastebin.com/lupecelaji.scala

Everything looks as expected. I tested the normal range and then moved into what should be outside of the range, and the return values look good.

HotelCalifornia:
@Octogonapus your test code looks like it's using an older revision of this PR, before we moved from two defined values to just one (`ULTRA_BAD_RESPONSE`)

Octogonapus:
@HotelCalifornia That's correct, my bad. I tested again using the correct version and got similar results (everything still looks good).

PR #35, opened 2017-08-26
Edit Makefile to support compiling code in subdirectories

Octogonapus:
Added Pixel's makefile to support subdirectories.


HotelCalifornia:
Is this confirmed to be backwards-compatible? I.e. have you confirmed that this will properly compile code without subdirectories?

Octogonapus:
Yes, I have tested a simple project with no subdirectories on the same robot I tested the more complex project on, and confirmed it is backwards-compatible.

edjubuh:
What happens when I have something like `libblrs/main.c` and `libwpi/main.c`? How do object files get resolved? Pretty sure at a glance it uses the relative path with directory separators as underscores but just want to make sure (so `libblrs_main.o` and `libwpi_main.o`)

Octogonapus:
All object files get generated with their full path as their name, so you can have files in multiple directories with the same name.

edjubuh:
What was your more complicated project structure like?

Octogonapus:
https://github.com/OkapiLib/OkapiLib

edjubuh:
Nice library! Could you theoretically remove the Makefile inside src? I'm leading to asking if the wildcard `**/*.c` matches `src/libwpi/hello/main.c`, which I can't remember from Makefile documentation.

Octogonapus:
You might be able to, but I'm not skilled enough in make to provide a definitive answer. I've poked this for a few days and haven't come up with anything.

template/src/Makefile
```diff
@@ -39,12 +39,14 @@ $(ASMOBJ): $(BINDIR)/%.o: %.$(ASMEXT)
...
 
 # Object management
-$(COBJ): $(BINDIR)/%.o: %.$(CEXT) $(HEADERS) $(HEADERS_2)
+$(COBJ): $(BINDIR)/$(subst ..,,$(subst /,_,%.o)): %.$(CEXT) $(HEADERS) $(HEADERS_2)
 	@echo CC $(INCLUDE) $<
-	@$(CC) $(INCLUDE) $(CFLAGS) -o $@ $<
+	@$(CC) $(INCLUDE) $(CFLAGS) -o $(BINDIR)/$(subst ..,,$(subst /,_,$@)) $<
 
-$(CPPOBJ): $(BINDIR)/%.o: %.$(CPPEXT) $(HEADERS) $(HEADERS_2)
+KEKSRC:=$(dir $(wildcard **/*/))
```
edjubuh: Is this being used?

NoMod-Programming:
Not working with c++ files in subdirectories of src.

PR #40, opened 2017-11-06
New makefile

edjubuh:
- Add makefile verbosity V=1 will show compile commands, V=2 will show every significant command (basically everything except echo)
- The default output of the Makefile should now be more friendly
- Size output is more human readable
- Support arbitrary subdirectories in the src folder
- Add library support into user Makefile
- Add user configurable parameters for things like extra CFLAGS and
  library parameters
- Makefile now has color output


HotelCalifornia:
@edjubuh can you take a couple minutes to update this at some point with what we know now?

@community: anyone care to test?

tehkhop:
Have been using this for 3+ months with no issues found. The colours are pretty too.

PR #61, opened 2018-06-14
Changed gyroGet to return a double with sub-degree accuracy.

tehkhop:
Does what it says on the tin.
Here's a graph showing int values in blue (before), and the double values in red (after). 

Values also round correctly.


stcarlso:
The addition of 128 in the original code was to correctly round the answer to the nearest integer after dividing. Since the PR casts the numerator to a double, the entire division is performed in floating point, making the addition unnecessary.

A few things to note:
* If you make this change, multiply by the pre-calculated reciprocal of 1.0 / 0.256 instead of dividing by 256.0. It will be significantly faster.
* Double precision math on the answer (including simple tasks like comparison with a constant) will be much slower. Consider having a second function with a different name to report a double answer. This also avoids an API break.
* When I last saw the code, PROS never used floating point anywhere else. If you make this change, floating point support functions will be pulled into all programs, even those which never used it otherwise, which may cause unwanted side effects. Consider reporting the answer times 100 or 1000 (can be done entirely in integer) to let the user choose if they want to divide to a float.

baylessj:
@stcarlso You make some very good points, and if this PR is to get merged in, I would certainly like to see the first and second points addressed.

In regard to your third point, are there any particular parts of the kernel that you can point to as having a higher likelihood of having errors induced by this addition of floating point operations? I'm assuming that such errors would likely be difficult to track down, and as we do not currently have test procedures set up for the PROS 2 kernel, there is a possibility that this sort of difficult to spot errors may not be noticed by our developers.

baylessj:
If we can ensure that no side-effects will come from the addition of floating point operations, then I will update my review to request that the changes for multiplication instead of division and an additional API function get made.

If there will be a non-trivial risk for the rest of the code as a result of this change, my recommendation would be that we scale up the gyro value by a factor of 100; if I remember correctly, RobotC scales the value by a factor of 10, so 100 would give us a reasonable increase in precision.

tehkhop:
What name would you recommend? gyroGetHR?

baylessj:
`gyroGetHR` sounds like the best option to me.

HotelCalifornia:
@tehkhop if you want to reopen this, you'll have to make a new PR targeting the cortex-master branch

HotelCalifornia:
JK, you can make the PR on purduesigbots/pros2

PR #81, opened 2018-12-16
Rebuild when dependent headers change

PixelToast:
Uses the gcc/g++ -M flag to generate header dependancies which get
thrown in deps/ and included next build

This is the most efficient solution where ONLY the source files that
depend on a header are recompiled versus doing a clean build on the
entire project

Should fix a LOT of random bugs people are getting when modifying
headers even including segfaults from discontinuity between function
signatures

Right now only tested on Linux with GCC 5.4.1
Both C and C++ tested
Subdirectories tested

<3


HotelCalifornia:
You need to make sure that gnu++17 is specified for c++ standard and I'm not sure why `version` is part of the `all` rule

HotelCalifornia:
And test with gcc 7.2.1 or higher

PixelToast:
I'm a bit confused, I did not change any of the flags or the all rule
https://github.com/purduesigbots/pros/blob/master/Makefile#L51
https://github.com/purduesigbots/pros/blob/master/common.mk#L17

PixelToast:
Tested on windows
GNU Make 4.2.1
GCC 7.2.1

HotelCalifornia:
@PixelToast if you want to reopen this, you'll have to make a new PR and have it target the cortex-master branch

HotelCalifornia:
JK, you can make the PR on purduesigbots/pros2

Issue #88, opened 2018-11-13
File I/O works with C functions but not with C++ std::fstream


3038922:
#### Expected Behavior:
Using STL to read and write files
#### Actual Behavior:
1 if i use "std::fstream" ,file can't open
2 If I use C language, I can open and read files. But if there are Chinese characters, PYTHON3 will make mistakes. Maybe it's the encoding of files.
3 If I set my configRead class as a global class, it will not read correctly, because the global class construction is disorderly, and it only works after initialize (). Because there is no parametric constructor for pros::Motor class, all of my robot project classes can only be constructed as local variables.
#### System information:
Operating System:  (Windows, OS X, Ubuntu)
windows 10 1803
PROS Version: (run `pros --version`)
pros, version 3.1.2
#### Additional Information
kernel 3.3.4
oklib 3.3.7
#### Screenshots/Output Dumps/Stack Traces
![1](https://user-images.githubusercontent.com/33926386/48387118-62b04b00-e72f-11e8-8c9f-8f14c0cc603e.png)
<img width="719" alt="2" src="https://user-images.githubusercontent.com/33926386/48387100-53c99880-e72f-11e8-86d3-beb08aff8cc7.png">

HotelCalifornia:
It seems that the problem is specifically with input streams (`std::ifstream`). Output streams work, but they will only flush when closed (either explicitly or when the dtor is called)

HotelCalifornia:
closed by #112

PR #89, opened 2019-01-31
✨ Add 🔥/❄ linking

edjubuh:
#### Summary:
Adds new form of linking that I call hot/cold. This method of linking creates two binary images that get uploaded to the V5. One is the "cold" image that contains any archives (e.g. libpros and okapilib) the project is using and libstdc++ and contains all functions the user could invoke from these libraries. User code is compiled and linked against this "cold" binary to form a "hot" binary. Using V5 file links, the cold binary can be uploaded once and only the hot binary can be uploaded at every compilation.

For a base project with okapilib that was originally ~471KB, the cold binary was approximately 1MB and the hot binary was 1.5 KB. When applying compression, the cold binary gets cut to 500 KB and the hot binary to <1KB (0.2% of the original size for a typical upload). The cold binary would take a few seconds to upload when wired and the hot binary would take less than a second even when uploaded wirelessly (ignoring channel transfer time).

#### Motivation:
This will make upload times shorter, especially wirelessly.

#### Impact:
 - The default binary name has changed from output.bin -> monolith.bin
 - Hot/cold linking will start off disabled by default and can be enabled with `USE_PACKAGE=1` in the Makefile (or at the command line)

##### References (optional):
purduesigbots/pros-cli#46

#### Test Plan:

- [x] Kernel
  - [X] Compile hot/cold
  - [X] Upload hot/cold
  - [x] Compile monolith
  - [x] Upload monolith
- [x] Base project
  - [x] Compile hot/cold (1.5KB/983KB)
  - [x] Upload hot/cold
  - [x] Compile monolith (425KB)
  - [x] Upload monolith
- [x] Okapi Clawbot
  - [x] Compile hot/cold (3.71KB/983KB)
  - [x] Upload hot/cold
  - [x] Compile monolith (425KB)
  - [x] Upload monolith
- [x] Rogers
  - [x] Compile hot/cold
  - [x] Upload hot/cold
  - [x] Compile monolith
  - [x] Upload monolith
- [x] Forkner
  - [x] Compile hot/cold (55.1KB/983KB w/o libforkner; 26.9KB/0.99MB w/libforkner)
  - [x] Upload hot/cold
  - [x] Compile monolith (537KB)
  - [x] Upload monolith
- [x] okapilib@4.0.0
   - [x] Compile hot/cold
   - [x] Run hot/cold
* Binary sizes are uncompressed


src/system/hot.c
```diff
@@ -0,0 +1,75 @@
...
+          "    it      lt\n"
+          "    strlt   r2, [r0], #4\n"
+          "    blt     zero_loop2");
+
+  __libc_init_array();
+}
+
+void invoke_install_hot_table() {
+  printf("%s %p %p %x %x\n", __FUNCTION__, (void*)install_hot_table, (void*)HOT_TABLE, MAGIC_ADDR[0], MAGIC_ADDR[1]);
+  if(MAGIC_ADDR[0] == 0x52616368 && MAGIC_ADDR[1] == 0x8CEF7310) {
```
Octogonapus: Instead of copy-pasting these two addresses could they be `#define`'s instead?

edjubuh:
I've now gone through newlib's crt0 to cross reference how the program is initialized with what we're doing.

https://github.com/bminor/newlib/blob/master/libgloss/arm/crt0.S

Indeed, all we need to do is zero (s)bss and call `__libc_init_array`. We don't need to worry about any of the stack initialization or hardware initialization - both of which FreeRTOS/kernel needs.

HotelCalifornia:
> ... both of which FreeRTOS/kernel needs.

to clarify, this is because FreeRTOS and the kernel will live in cold storage, so said initializations will happen there

common.mk
```diff
@@ -118,23 +118,39 @@ GETALLOBJ=$(sort $(call ASMOBJ,$1) $(call COBJ,$1) $(call CXXOBJ,$1))
...
+COLD_ELF:=$(basename $(COLD_BIN)).elf
+
+DEFAULT_BIN=$(MONOLITH_BIN)
+ifeq ($(USE_PACKAGE),1)
+DEFAULT_BIN=$(HOT_BIN)
+endif
+
+# Check if USE_PACKAGE is defined to check for migration steps from purduesigbots/pros#87
+ifndef USE_PACKAGE
+$(error Your Makefile must be migrated! Visit https://pros.cs.purdue.edu/v5/releases/kernel3.1.6.html to learn how)
```
edjubuh: This makes the required migration in #88 more obvious

HotelCalifornia: looks like you linked the wrong issue/PR

edjubuh:
Random comment no-one will care about... The current configuration requires the CPU to jump about as far as possible around memory space from the `0x07800000` region to the `0x0380000` region when invoking a library function. The maximum distance an immediate jump can go in armv7-a is ~3MB and this is ~60-70MB. Point is, we'd have to make cold text section < 3MB (or ensure that the whole hot image < 2MB) for this to not be a problem. I could probably lay out the memory this way, but it means that I have to flip it from cold being startup code to hot being startup code. I'm not entirely sure this'll be a problem anyway since a lot of the text should lie in nearby L1 I-cache sets; I'm fairly certain both hot and cold can be simultaneously mapped. If I really wanted, I could set up the PMCR to track this but effort of needing to read all of that documentation makes me 😢

PR #98, opened 2019-02-05
🐛 Prevent Buffer Overrun in Controller Printing

### LINKED ISSUE #97: Possibly buffer overrun in controller_print()
Could vasprintf() return buffer that is smaller than 16 bytes?

It is, probably, a good idea to check return value vasprintf() which is guaranteed to allocate only number of printed bytes + 1.

https://github.com/purduesigbots/pros/blob/bc0fa3e3ce4723f738741cd4c01be62f71634cf1/src/devices/controller.c#L248

baylessj:
#### Summary:
This moves to more secure string operations for the `controller_set_text` and `controller_print` functions. Previously, `controller_print` would overrun the buffer when printing a string smaller than the maximum allowable length, and a string longer than CONTROLLER_MAX_COLS could be written into the smaller destination buffer in `controller_set_text`.

#### Motivation:
This will prevent buffer overrun and its associated issues in the controller printing functions.

##### References (optional):
Closes #97.

#### Test Plan:

- [x] Compiles
- [x] Call `controller_set_text` with a string longer than `CONTROLLER_MAX_COLS`
- [x] Call `controller_print` with a string that is shorter than `CONTROLLER_MAX_COLS`


src/devices/controller.c
```diff
@@ -242,10 +241,8 @@ int32_t controller_print(controller_id_e_t id, uint8_t line, uint8_t col, const
 
 	va_list args;
 	va_start(args, fmt);
-	char* buf;
-	vasprintf(&buf, fmt, args);
-
-	buf[CONTROLLER_MAX_COLS] = '\0';
```
HotelCalifornia: we should still null-terminate the string. doesn't hurt, and may be safer

src/devices/controller.c
```diff
@@ -208,8 +208,7 @@ int32_t controller_set_text(controller_id_e_t id, uint8_t line, uint8_t col, con
 		col++;
 
 	char* buf = (char*)malloc(CONTROLLER_MAX_COLS + 1);
-	strcpy(buf, str);
-	buf[CONTROLLER_MAX_COLS] = '\0';
+	strlcpy(buf, str, CONTROLLER_MAX_COLS + 1);
```
HotelCalifornia: I have a feeling this won't compile. It's a nonstandard function and I'm skeptical that it is provided in our libc implementation.

That said, there shouldn't be anything wrong with just using `strncpy` here

src/devices/controller.c
baylessj: `vsnprintf()` automatically appends a null terminator to the destination string so we're covered there.

src/devices/controller.c
baylessj: That's reasonable, it's an odd one. The only real benefit of `strlcpy()` over `strncpy()` is the automatic addition of the null terminator, which could just be added manually after a `strncpy()` call if the current code doesn't compile.

edjubuh: `char* buf = strncpy(str, CONTROLLER_MAX_COLS + 1);` would do the whole thing in one line.

HotelCalifornia:
tested with this in opcontrol, no problems
```cpp
	pros::Controller master(pros::E_CONTROLLER_MASTER);
	master.set_text(0, 0, "abcdefghijklmnopqrstuvwxyz");
	pros::delay(75);
	master.print(1, 0, "%d abcdefghijklmnopqrstuvwxyz", 69);
```

PR #99, opened 2019-02-07
Fix archive compilation for kernel

edjubuh:
#### Summary:
- Fix the Makefile to use the right variable for what we want included in the archive
- Comments don't get echo'd when building

#### Motivation:
It's an annoying warning and console print

##### References (optional):
None

#### Test Plan:
- [x] Comment went away when compiling
- [x] No more warnings about including api.h and libv5rts.patched.a

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,12 +64,12 @@ template: clean-template library
 	$Dprosv5 c create-template $(TEMPLATE_DIR) kernel $(shell cat $(ROOT)/version) --system "./**/*" --user "src/opcontrol.{c,cpp,cc}" --user "src/initialize.{cpp,c,cc}" --user "src/autonomous.{cpp,c,cc}" --user "include/main.{hpp,h,hh}" --user "Makefile" --target v5 --output bin/output.bin
 
 LIBV5RTS_EXTRACTION_DIR=$(BINDIR)/libv5rts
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -64,12 +64,12 @@ template: clean-template library
 	$Dprosv5 c create-template $(TEMPLATE_DIR) kernel $(shell cat $(ROOT)/version) --system "./**/*" --user "src/opcontrol.{c,cpp,cc}" --user "src/initialize.{cpp,c,cc}" --user "src/autonomous.{cpp,c,cc}" --user "include/main.{hpp,h,hh}" --user "Makefile" --target v5 --output bin/output.bin
 
 LIBV5RTS_EXTRACTION_DIR=$(BINDIR)/libv5rts
-$(LIBAR): $(call GETALLOBJ,$(EXCLUDE_SRCDIRS) $(EXCLUDE_FROM_LIB)) $(EXTRA_LIB_DEPS)
```
Octogonapus: Why is `EXCLUDE_SRCDIRS` getting removed?

edjubuh: It's not quite being removed:
https://github.com/purduesigbots/pros/blob/f34a219cb49a6b5a8e81b6852abe4c84efa975d0/Makefile#L29-L32

HotelCalifornia:
uh ok pretty sure this was properly merged

PR #100, opened 2019-02-07
🐛 Refactor task_notify_when_deleting to support task_to_notify dying before target_task

### LINKED ISSUE #86: Extend task_notify_when_deleting to support task_to_notify dying before target_task dies
Currently, something bad will happen if `task_to_notify` dies before `target_task` dies (@edjubuh feel free to elaborate on what would happen). This means that I don't consider it usable in an OkapiLib release until nothing bad happens (nothing happening at all is a fine option, if `task_to_notify` is dead, just don't notify it!).

The code area we will need to look at is: https://github.com/purduesigbots/pros/blob/4eb9ea02eacae0872b88d6109e62718bac19f627/src/rtos/tasks.c#L916

One proposal is to run a check to see if the task is still alive and only notify it if it's in the expected state; however, an implementation problem with this is that TCBs can get reused, so we will have to implement something ourselves to track whether a task has been deleted. This could be done through a resizable data structure on the heap which stores an alive bit per task. This bit could be used in a lookup inside [the hook](https://github.com/purduesigbots/pros/blob/4eb9ea02eacae0872b88d6109e62718bac19f627/src/rtos/task_notify_when_deleting.c#L105) so it does nothing.

HotelCalifornia:
#### Summary:

- Increments `configNUM_THREAD_LOCAL_STORAGE_POINTERS` again
- Adds another linked list to thread local storage that will keep track of the tasks to whose death events `task_to_notify` is subscribed
- When a task dies, the `task_delete_hook` now iterates through the task's list of subscriptions and removes references to itself from the subscriptions' lists of subscribers

#### Motivation:

Previously, the case in which `task_to_notify` dies before `target_task` was undefined behavior, because `target_task` would attempt to notify `task_to_notify` even if it's already been cleaned up. The naive solution of checking the `task_to_notify`'s state directly would not have worked because in certain cases (as with the competition tasks), the TCB can be reused.

More information regarding the motivation behind this change can be found at #86, OkapiLib/OkapiLib#249, and OkapiLib/OkapiLib#321

##### References (optional):

Closes #86 

#### Test Plan:

- [x] execute test plan in src/tests/task_notify_when_deleting.c (regression test)
(the following are adapted from https://github.com/purduesigbots/pros/pull/100#issuecomment-461282903. thanks @Octogonapus)
- [x] delete `task_to_notify` and then delete `target_task`, verify nothing goes horrendously wrong
- [x] repeat previous test with OkapiLib where an internal task is the `task_to_notify`

### DIFF
--- a/src/rtos/task_notify_when_deleting.c
+++ b/src/rtos/task_notify_when_deleting.c
@@ -46,6 +46,10 @@ static struct notify_delete_action* _find_task(linked_list_s_t* ll, task_t task)
   return args.found_action;
 }
 
+void task_notify_when_deleting_init() {
+  task_notify_when_deleting_mutex = mutex_create_static(&task_notify_when_deleting_mutex_buf);
+}
+
 void task_notify_when_deleting(task_t target_task, task_t task_to_notify,
                                uint32_t value, notify_action_e_t notify_action) {
   task_to_notify = (task_to_notify == NULL) ? pxCurrentTCB : task_to_notify;
@@ -57,9 +61,8 @@ void task_notify_when_deleting(task_t target_task, task_t task_to_notify,
     return;
   }
 
-  // lazily initialize mutex
   if (task_notify_when_deleting_mutex == NULL) {
-    task_notify_when_deleting_mutex = mutex_create_static(&task_notify_when_deleting_mutex_buf);
+    task_notify_when_deleting_mutex = mutex_create();
   }
   mutex_take(task_notify_when_deleting_mutex, TIMEOUT_MAX);
 
--- a/src/rtos/task_notify_when_deleting.c
+++ b/src/rtos/task_notify_when_deleting.c
@@ -63,7 +63,7 @@ void task_notify_when_deleting(task_t target_task, task_t task_to_notify,
     ll_node_s_t* it = subscriptions_ll->head;
     bool found = false;
     while (it != NULL && !found) {
-      found = it->payload.data == task_to_notify;
+      found = it->payload.data == target_task;
       it = it->next;
     }
     if (!found) {
--- a/src/rtos/task_notify_when_deleting.c
+++ b/src/rtos/task_notify_when_deleting.c
@@ -126,11 +126,6 @@ static void unsubscribe_hook_cb(ll_node_s_t* node, void* task_to_remove) {
   linked_list_s_t* subscriptions_list = pvTaskGetThreadLocalStoragePointer(subscription, SUBSCRIBERS_TLSP_IDX);
   if (subscriptions_list != NULL) {
     linked_list_remove_data(subscriptions_list, task_to_remove);
-    // cleanup the list if we've removed its last member
-    if (subscriptions_list->head == NULL) {
-      linked_list_free(subscriptions_list);
-      vTaskSetThreadLocalStoragePointer(subscription, SUBSCRIBERS_TLSP_IDX, NULL);
-    }
   }
 }
 
@@ -148,6 +143,8 @@ void task_notify_when_deleting_hook(task_t task) {
   linked_list_s_t* ll = pvTaskGetThreadLocalStoragePointer(task, SUBSCRIPTIONS_TLSP_IDX);
   if (ll != NULL) {
     linked_list_foreach(ll, unsubscribe_hook_cb, task);
+    linked_list_free(ll);
+    vTaskSetThreadLocalStoragePointer(task, SUBSCRIPTIONS_TLSP_IDX, NULL);
   }
   // notify subscribed tasks of this task's deletion
   ll = pvTaskGetThreadLocalStoragePointer(task, SUBSCRIBERS_TLSP_IDX);
--- a/src/rtos/task_notify_when_deleting.c
+++ b/src/rtos/task_notify_when_deleting.c
@@ -124,12 +124,13 @@ static void unsubscribe_hook_cb(ll_node_s_t* node, void* task_to_remove) {
   task_t subscription = node->payload.data;
 
   linked_list_s_t* subscriptions_list = pvTaskGetThreadLocalStoragePointer(subscription, SUBSCRIBERS_TLSP_IDX);
-  linked_list_remove_data(subscriptions_list, task_to_remove);
-
-  // cleanup the list if we've removed its last member
-  if (subscriptions_list->head == NULL) {
-    linked_list_free(subscriptions_list);
-    vTaskSetThreadLocalStoragePointer(subscription, SUBSCRIBERS_TLSP_IDX, NULL);
+  if (subscriptions_list != NULL) {
+    linked_list_remove_data(subscriptions_list, task_to_remove);
+    // cleanup the list if we've removed its last member
+    if (subscriptions_list->head == NULL) {
+      linked_list_free(subscriptions_list);
+      vTaskSetThreadLocalStoragePointer(subscription, SUBSCRIBERS_TLSP_IDX, NULL);
+    }
   }
 }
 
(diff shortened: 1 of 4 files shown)


HotelCalifornia:
cc @octogonapus

I'm also looking for help coming up with tests to verify that this PR works as expected

Octogonapus:
In the following scenarios, task B is notified when task A is deleted.
 - [ ] Delete task A and verify task B is notified (regression check).
 - [ ] Delete task B and then delete task A and verify nothing blows up.
 - [ ] Repeat the previous test under OkapiLib where an internal task (such as CCPID's internal task) gets notified.

src/rtos/task_notify_when_deleting.c
```diff
@@ -41,15 +44,39 @@ void task_notify_when_deleting(task_t target_task, task_t task_to_notify,
...
-  linked_list_s_t* target_ll = pvTaskGetThreadLocalStoragePointer(target_task, TLSP_IDX);
+  // task_to_notify maintains a list of the tasks whose deletion it cares about.
+  // This will allow us to unsubscribe from notification if/when task_to_notify
+  // is deleted
+  linked_list_s_t* subscriptions_ll = pvTaskGetThreadLocalStoragePointer(task_to_notify, SUBSCRIPTIONS_TLSP_IDX);
+  if (subscriptions_ll == NULL) {
+    subscriptions_ll = linked_list_init();
+    vTaskSetThreadLocalStoragePointer(task_to_notify, SUBSCRIPTIONS_TLSP_IDX, subscriptions_ll);
+  }
+  if (subscriptions_ll != NULL) {
```
Octogonapus: Can this be an else block instead?

HotelCalifornia: No, the first branch initializes the list if it doesn't exist. The second branch should happen regardless of whether the first branch happened or not. All it needs is a non-null list

HotelCalifornia:
simple regression test:
```cpp
void opcontrol() {
	pros::delay(100);
	pros::Task target_task([](void* param) {
		pros::delay(500);
		printf("task 1 dead\n");
	}, nullptr, TASK_PRIORITY_DEFAULT, TASK_STACK_DEPTH_DEFAULT, "target_task");
	pros::Task task_to_notify([](void* param) {
		while (pros::c::task_notify_take(true, 0) != 1) {
			pros::delay(10);
		}
		printf("task 2 dead\n");
	}, nullptr, TASK_PRIORITY_DEFAULT, TASK_STACK_DEPTH_DEFAULT, "task_to_notify");
	pros::c::task_notify_when_deleting(target_task, task_to_notify, 1, pros::E_NOTIFY_ACTION_INCR);
	pros::delay(5000);
}
```
(executing src/tests/task_notify_when_deleting.c also passes)

Octogonapus:
I have done some testing myself and I think this works. The logs from my testing are attached.
[task_notify_when_deleting test results.txt](https://github.com/purduesigbots/pros/files/2848081/task_notify_when_deleting.test.results.txt)

src/rtos/task_notify_when_deleting.c
```diff
@@ -41,15 +44,39 @@ void task_notify_when_deleting(task_t target_task, task_t task_to_notify,
...
+  // neither task is NULL (implying that scheduler hasn't started yet)
   if (task_to_notify == target_task || !task_to_notify || !target_task) {
     return;
   }
 
-  linked_list_s_t* target_ll = pvTaskGetThreadLocalStoragePointer(target_task, TLSP_IDX);
+  // task_to_notify maintains a list of the tasks whose deletion it cares about.
+  // This will allow us to unsubscribe from notification if/when task_to_notify
+  // is deleted
+  linked_list_s_t* subscriptions_ll = pvTaskGetThreadLocalStoragePointer(task_to_notify, SUBSCRIPTIONS_TLSP_IDX);
```
edjubuh: So subscriptions_ll is the linked list representing what tasks will notify `task_to_notify` when they die. What is the payload?

src/rtos/task_notify_when_deleting.c
```diff
@@ -41,15 +44,39 @@ void task_notify_when_deleting(task_t target_task, task_t task_to_notify,
...
+  if (subscriptions_ll == NULL) {
+    subscriptions_ll = linked_list_init();
+    vTaskSetThreadLocalStoragePointer(task_to_notify, SUBSCRIPTIONS_TLSP_IDX, subscriptions_ll);
+  }
+  if (subscriptions_ll != NULL) {
+    // check whether task_to_notify is already subscribed. if so, do nothing
+    ll_node_s_t* it = subscriptions_ll->head;
+    bool found = false;
+    while (it != NULL && !found) {
```
edjubuh: And we search through that linked list for `task_to_notify`? Why?

src/rtos/task_notify_when_deleting.c
```diff
@@ -103,10 +144,16 @@ static void delete_hook_cb(ll_node_s_t* node, void* ignore) {
 }
 
 void task_notify_when_deleting_hook(task_t task) {
-  linked_list_s_t* ll = pvTaskGetThreadLocalStoragePointer(task, TLSP_IDX);
+  // if this task was subscribed to any other task deletion events, unsubscribe
+  linked_list_s_t* ll = pvTaskGetThreadLocalStoragePointer(task, SUBSCRIPTIONS_TLSP_IDX);
+  if (ll != NULL) {
+    linked_list_foreach(ll, unsubscribe_hook_cb, task);
+  }
```
edjubuh: Why are you not freeing up the subscriptions linked list as `task` is dying?

src/rtos/task_notify_when_deleting.c
HotelCalifornia:
```c
// check whether task_to_notify is already subscribed. if so, do nothing
```

src/rtos/task_notify_when_deleting.c
HotelCalifornia: payload is just the task handle

src/rtos/task_notify_when_deleting.c
HotelCalifornia: oh yeah. good point.

src/rtos/task_notify_when_deleting.c
HotelCalifornia: yeah oops. too many conflicting terminologies

src/rtos/task_notify_when_deleting.c
edjubuh: So subscriptions is a list of tasks `task_to_notify` has requested to be notified of when any one of them have died?

HotelCalifornia: yes. that way when `task_to_notify` dies, it can iterate through this list and remove itself from being notified, clearing up any issues of UB that would result from it dying before other tasks try to notify it

Octogonapus:
I ran my tests again with the latest changes. Still works.

src/rtos/task_notify_when_deleting.c
edjubuh: Below, why would `task_to_notify` ever be a member of its subscription list if we have:

https://github.com/purduesigbots/pros/blob/f34a219cb49a6b5a8e81b6852abe4c84efa975d0/src/rtos/task_notify_when_deleting.c#L45

and 

https://github.com/purduesigbots/pros/blob/76a22151b7020398b2a36e2e8c52eb73525cf0d0/src/rtos/task_notify_when_deleting.c#L70

HotelCalifornia: (discussed offline: this was a misunderstanding stemming from the mistake I made that was called out in the review comment below this one)

edjubuh:
I'll approve with a review from @Octogonapus

PR #112, opened 2019-02-17
🚨🔊 Add a better Data Abort Handler

### LINKED ISSUE #150: C++ exceptions not working with hot/cold linking
#### Expected Behavior:
To be able to throw and catch an exception without `terminate()` being called, regardless of whether hot/cold linking is enabled.

#### Actual Behavior:
On hot/cold linking, throwing a `std::runtime_error` and catching it with `catch(...)` will result in this output:
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  Exception text
```

#### Steps to reproduce:
1. Create a new PROS project
2. In `initialize()` write
```
void initialize() {
	try {
		pros::delay(5000);
		throw std::runtime_error("Exception text");
	} catch(...) {
		printf("Caught some error.\n");
	}
}
```
3. Build and run to get expected output
4. Change USE_PACKAGE to 1 in Makefile
5. Build and run to get unexpected output
#### System information:
Platform: V5
PROS Kernel Version: 3.1.6

#### Additional Information
Using `arm-none-eabi` 7.3.1.

edjubuh:
#### Summary:
Dunno how to be concise yet. But this makes the data abort handler do something interesting.
- We have our own error box on the V5 that reports the proper PC and also current task name (because we can)
- We still background process and printfs work, so I print registers of the state that caused the abort and stack trace of the running task
   - This required adding unwind tables. Some additional adjustments will need to be made once #89 lands to ensure that the unwinder can find both hot and cold tables.

There will need to be additional work in the CLI to create a frontend for doing addr2line and objdump to make it a bit easier to see where things went wrong.

#### Motivation:
I'm tired of guessing why the MEMORY ERROR 0000000 box pops up because the 0000000 is meaningless.

##### References (optional):
Closes #150 

#### Test Plan:
- [X] `int x = *(int*)0; printf("%d\n" ,x);` and get [this output](https://imgur.com/LkVwZ3f)


Octogonapus:
I have been seeing a lot of people ask for memory usage. Is it possible to get the current heap/task stack memory usage?

edjubuh:
> I have been seeing a lot of people ask for memory usage. Is it possible to get the current heap/task stack memory usage? That might also be something we want to display.

Stack usage is a bit philosophical. Do you want to see current task stack usage or peak stack usage or perhaps average stack usage? Current task stack usage at the time of the abort is easy since we know the current SP and where the top of the stack is (since we have access to the TCB). Peak stack usage is a more expensive to compute and perfect if you're trying to fool it, but FreeRTOS does have a [facility](https://www.freertos.org/uxTaskGetStackHighWaterMark.html). We don't currently measure any sort of average stack usage. Average stack usage could be probed periodically on every context switch and do a rolling average on that, but that seems like a lot of work for little tangible benefit.

edjubuh:
> I have been seeing a lot of people ask for memory usage. Is it possible to get the current heap/task stack memory usage? That might also be something we want to display.

Heap usage can be done by users today with [`mallinfo`](https://www.gnu.org/software/libc/manual/html_node/Statistics-of-Malloc.html). I'll add that to the printf'd abort report

Octogonapus:
> that seems like a lot of work for little tangible benefit

I agree. Just reporting whatever the stack usage happens to be at abort time is probably good enough for now.

edjubuh:
From some initial testing from @baylessj, we've discovered that many (if not all) of the `FILE*` family of functions (`fopen`, `fputs`) are marked as `cantunwind`. Investigating further.

edjubuh:
Turns out there's no unwinding information in newlib/libc.a that ships in the GNU Toolchain whatsoever... I'm trying to figure out if that was intentional.

```
arm-none-eabi-readelf -u /usr/arm-none-eabi/lib/libc.a | less
```

edjubuh:
It's possible this improves the state of #88

HotelCalifornia:
This still does not work entirely correctly, as far as my tests have shown. @edjubuh are you be able to take a look at this again sometime soon?

edjubuh:
> This still does not work entirely correctly, as far as my tests have shown. @edjubuh are you be able to take a look at this again sometime soon?

Have you verified the values of `__hot_exidx_end`, `__hot_exidx_start`, `start_of_hot_mem`, `start_of_cold_mem`, `end_of_hot_mem`, and `end_of_cold_mem`?

HotelCalifornia:
I believe I did, but I can double check

src/system/hot.c
```diff
@@ -71,6 +76,7 @@ void invoke_install_hot_table() {
   // printf("%s %p %p %x %x\n", __FUNCTION__, (void*)install_hot_table, (void*)HOT_TABLE, MAGIC_ADDR[0], MAGIC_ADDR[1]);
   if(vexSystemLinkAddrGet() == (uint32_t)0x03800000 && MAGIC_ADDR[0] == MAGIC0 && MAGIC_ADDR[1] == MAGIC1) {
     install_hot_table(HOT_TABLE);
+
```
edjubuh: @HotelCalifornia When you clean up the PR to remove some of the other test changes, can you rm this accidentally added line?

.arclint
```diff
@@ -7,7 +7,8 @@
 		},
 		"spelling": {
 			"type": "spelling",
-			"include": "(.*)"
+			"include": "(.*)",
+            "exclude": "(^(src|include)|(rtos/|display/lv_.*))"
```
HotelCalifornia: 🙄 
```suggestion
			"exclude": "(^(src|include)|(rtos/|display/lv_.*))"
```

PR #119, opened 2019-02-20
Don't include timestamp in cold image

edjubuh:
#### Summary:
Remove timestamp compilation from cold ELF

#### Motivation:
If you have multiple projects with the same set of libraries (e.g. same hash) but the cold ELF was compiled at different times, then the CRCs won't match even though the relevant symbols are equivalent.

#### Test Plan:
- [x] Still compiles
- [x] Create two projects, compile cold ELFs independently, and CRCs should still match
  - Validated by me and @jcgrif


HotelCalifornia:
as this is a hotfix, it should be based on master directly

PR #120, opened 2019-02-20
Correction of annotation errors in exposure

3038922:
Signed-off-by: 3038922 <31612534@qq.com>

#### Summary:

It's just a revision of the notes. My students found it.
#### Test Plan:

- [ ] test item
```
int32_t vision_set_exposure(uint8_t port, const uint8_t percent) {
	claim_port(port - 1, E_DEVICE_VISION);
	// This translation comes from VEX to match the brightness represented in vision utility
	vexDeviceVisionBrightnessSet(device->device_info, (((int)((percent * 100) + 50)) / 255));
	return_port(port - 1, 1);
}
```
if percent ==100
((100 *100)+50)/255=  39
I'm not sure if this is correct.

### DIFF
--- a/include/pros/vision.h
+++ b/include/pros/vision.h
@@ -267,7 +267,7 @@ vision_object_s_t vision_get_by_code(uint8_t port, const uint32_t size_id, const
  * \param port
  *        The V5 port number from 1-21
  *
- * \return The current exposure setting from [0,150], PROS_ERR if an error
+ * \return The current exposure setting from [0,100], PROS_ERR if an error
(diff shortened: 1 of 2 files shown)


edjubuh:
Ah, thanks for pointing this out. We need to update the purduesigbots/pros-docs to reflect the change. In 3.1.5, we changed the parameter to match the scaling used in the Vision Sensor Utility, which is [0, 150]. It was previously [0, 100] as you're noting from the online documentation.

PR #130, opened 2019-03-11
Migate kernel makefile to test_output_2

edjubuh:
#### Summary:
Migrate kernel makefile to `test_output_2` and make make make again.

#### Motivation:
It's broken because `@cd` isn't a command. `@` was supposed to suppress echoing the whole command, but `test_output` has changed.

##### References (optional):
#123 

#### Test Plan:
- [X] It build


include/api.h
```diff
@@ -41,8 +41,8 @@
 
 #define PROS_VERSION_MAJOR 3
 #define PROS_VERSION_MINOR 1
-#define PROS_VERSION_PATCH 6
-#define PROS_VERSION_STRING "3.1.6"
+#define PROS_VERSION_PATCH 7
+#define PROS_VERSION_STRING "3.1.7-dirty.6.021f8be"
```
HotelCalifornia: dunno about that

HotelCalifornia: same in version below 👇

edjubuh: whoops fixed.

PR #138, opened 2019-04-16
A college student gave me advice.

3038922:
A college student gave me advice.,I only know that this can alleviate the mutex problem.I don't understand the principle.


edjubuh:
Is there a test plan for this PR? I'm not against the adding `__sync_synchronize`, but want to make sure that you have a clean demonstration of it.

edjubuh:
Also, what version of the toolchain are you using?

```bash
arm-none-eabi-gcc --version
```

3038922:
```
arm-none-eabi-gcc.exe (GNU Tools for Arm Embedded Processors 8-2018-q4-major) 8.2.1 20181213 (release) [gcc-8-branch revision 267074]
```
Before that, if I used the official PID, his speed loop was often out of sync. The chassis boot was skewed. Just use this. I don't know what this function is for. I just think it's useful.

We are used ccls language server. so we changed .gitignore.
I'm not familiar with pull requests. Errors sent my own. gitignore. So I deleted him.

edjubuh:
I'm still not clear why you think this ought to be added. I can compile with an explicit call to `__sync_synchronize` and it properly puts a `dmb` instruction in the output. I used the the 8-2018-q4-major toolchain from Arm's website.

PR #139, opened 2019-04-21
Fix incorrect port type

TixoRebel:
#### Summary:
Fix incorrect port type in some motor functions

#### Motivation:
They were different from everywhere else with seemingly no reason.

#### Test Plan:

- [x] Still compiles


edjubuh:
Good catch, thanks!

edjubuh:
Make it so

PR #141, opened 2019-04-30
Add alternate generic serial API and fix generic serial filesystem driver

TixoRebel:
#### Summary:
- Add alternate generic serial API
- Fix filesystem generic serial driver

#### Motivation:
The alternate generic serial API was added for users who want a lower level serial API. The generic serial filesystem driver was broken so it was fixed and implemented using the new API so the VEXos functions are only called in one place.

#### Test Plan:
- [x] Compile kernel
- [x] Ensure `src/tests/generic_serial.cpp` passes
- [x] Ensure `src/tests/generic_serial_file.cpp` passes

### DIFF
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -28,8 +28,10 @@
 #pragma GCC diagnostic ignored "-Wall"
 #include "display/lvgl.h"
 #pragma GCC diagnostic pop
+#include "pros/serial.h"
 
 #ifdef __cplusplus
+#include "pros/serial.hpp"
 namespace pros::c {
 extern "C" {
 #endif
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -538,14 +538,14 @@ int32_t fdctl(int file, const uint32_t action, void* const extra_arg);
  * The extra argument is not used with this action, provide any value (e.g.
  * NULL) instead
  */
-#define DEVCTL_FIONWRITE 17
+#define DEVCTL_FIONWRITE 18
 
 /**
  * Action macro to set the Generic Serial Device's baudrate.
  *
  * The extra argument is the baudrate.
  */
-#define DEVCTL_SET_BAUDRATE 18
+#define DEVCTL_SET_BAUDRATE 17
 
 #ifdef __cplusplus
 }
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -532,11 +532,20 @@ int32_t fdctl(int file, const uint32_t action, void* const extra_arg);
 #define DEVCTL_FIONREAD 16
 
 /**
+ * Action macro to check if there is space available in the Generic Serial
+ * Device's output buffer
+ *
+ * The extra argument is not used with this action, provide any value (e.g.
+ * NULL) instead
+ */
+#define DEVCTL_FIONWRITE 17
+
+/**
  * Action macro to set the Generic Serial Device's baudrate.
  *
  * The extra argument is the baudrate.
  */
-#define DEVCTL_SET_BAUDRATE 17
+#define DEVCTL_SET_BAUDRATE 18
 
 #ifdef __cplusplus
 }
--- a/include/pros/serial.h
+++ b/include/pros/serial.h
@@ -1,0 +1,247 @@
+/**
+ * \file pros/serial.h
+ *
+ * Contains prototypes for the V5 Generic Serial related functions.
+ *
+ * Visit https://pros.cs.purdue.edu/v5/tutorials/topical/serial.html to learn
+ * more.
+ *
+ * This file should not be modified by users, since it gets replaced whenever
+ * a kernel upgrade occurs.
+ *
+ * Copyright (c) 2017-2019, Purdue University ACM SIGBots.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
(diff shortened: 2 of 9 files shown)


src/system/dev/dev_driver.c
```diff
@@ -41,19 +42,29 @@ typedef struct dev_file_arg {
...
 	dev_file_arg_t* file_arg = (dev_file_arg_t*)arg;
 	uint32_t port = file_arg->port;
-	claim_port(port - 1, E_DEVICE_GENERIC);
-	// check flags
-	int32_t result = vexDeviceGenericSerialReceive(device->device_info, buffer, len);
-	return_port(port - 1, result);
+	int32_t recv = 0;
+	do {
+		recv = serial_read(port, (uint8_t*)(buffer + recv), len - recv);
+	} while (!(file_arg->flags & O_NONBLOCK) && recv < 1);
```
HotelCalifornia: is it really expected behavior that opening a device like this in blocking mode (which is the default) can starve the calling task when reading or writing? sure, the workaround is to have users open in nonblocking mode and then poll it manually, but is that really what we want? I'd probably implement like

```c
while (
    (recv = serial_read(port, (uint8_t*)(buffer + recv), len - recv) < 1
    && !(file_arg->flags & O_NONBLOCK)
) task_delay(MINIMUM_DELAY_INTERVAL);
```
or something like that

HotelCalifornia: (same goes for writing)

include/pros/apix.h
```diff
@@ -531,12 +531,21 @@ int32_t fdctl(int file, const uint32_t action, void* const extra_arg);
...
+ */
+#define DEVCTL_FIONWRITE 17
+
 /**
  * Action macro to set the Generic Serial Device's baudrate.
  *
  * The extra argument is the baudrate.
  */
-#define DEVCTL_SET_BAUDRATE 17
+#define DEVCTL_SET_BAUDRATE 18
```
edjubuh: As annoying as it sounds, I'd prefer to leave `DEVCTL_SET_BAUDRATE` as 17 and have `FIONWRITE` be 18.

TixoRebel: Ok, that makes sense. I've changed `DEVCTL_SET_BAUDRATE` back to 17, though, I left it after `DEVCTL_FIONWRITE`. I figure that ordering makes more sense even though numerically it's out of order.

src/system/dev/dev_driver.c
TixoRebel: I couldn't find any constant to serve the purpose of `MINIMUM_DELAY_INTERVAL`, so it now sleep's for one millisecond (it won't sleep at all if it reads data the first time `serial_read` is called) . It is implemented it longform just to check `recv` for `PROS_ERR`. All the same goes for writing.

HotelCalifornia: Yeah that was just a placeholder. I think the smallest delay time we use anywhere is 2ms, so maybe make it that

TixoRebel: Ok, changed to 2ms

HotelCalifornia:
might also be good to merge `purduesigbots:develop` back into your branch to deduplicate the motors port type fix

HotelCalifornia:
I've queued a build on Azure (not sure why it didn't do it automatically), so there should be artifacts shortly

TixoRebel:
> Currently it looks like the serial API is being included in the default api header, I'm thinking that it would be best to include this in apix.h since this API would only be used by VEX U students and side projects. Would probably be an unecessary addition for most users.

That makes sense, I've moved the include over to `apix.h`. As far as my testing can tell, everything seems functional at this point. There are two tests in `src/tests`, `generic_serial.cpp` and `generic_serial_file.cpp`, which test the general functionality of the API using a loopback cable plugged into ports 1 and 2 (just a normal V5 cable connecting the ports).

include/pros/serial.h
```diff
@@ -0,0 +1,247 @@
...
+ * EINVAL - The given value is not within the range of V5 ports (1-21).
+ * EACCES - Another resource is currently trying to access the port.
+ *
+ * \param port
+ *        The V5 port number from 1-21
+ *
+ * \return The number of bytes free or PROS_ERR if the operation failed,
+ * setting errno.
+ */
+int32_t serial_write_free(uint8_t port);
```
baylessj: This function seems a bit confusing to me at first, perhaps the call should be renamed to `serial_get_num_free` or similar? I'm not a fan of having write as the action in this function since no writing occurs.

baylessj:
I didn't pay close enough attention to this PR the first time I read through it, didn't realize that I can test this without getting a separate device set up. Regardless, I have an RS485 converter on its way to me so I can hopefully test this with a raspberry pi as well, but in the meantime I'll try to get testing done with the loopback cable method.

baylessj:
Ran the test cases and everything worked as expected, I'll give an approving review once the write_free function is renamed.

include/pros/serial.h
TixoRebel: Would `serial_get_write_free` work? I'd like to keep write (or similar) in there just to specify it's returning data about the write buffer. If we're renaming `serial_write_free`, `serial_read_avail` should probably be changed to match.

baylessj: `serial_get_write_free` and `serial_get_ready_avail` sound like good names to me 👍🏻

Issue #93, opened 2019-01-31
Request to add functionality: V5 LLEMU "Center Text" Class Function


onthegrid007:
#### Desired Feature:
Class Function to center LCD text

#### System information:
Operating System: Windows 10

PROS Version: pros, version 3.1.2

#### Additional Information
As of now, I am using spaces to center text.

Jerrylum:
How about make a `pros::lcd::set_text_align()` feature?

```cpp
typedef enum lcd_text_align_e {
	E_LCD_TEXT_ALIGN_LEFT = 0,
	E_LCD_TEXT_ALIGN_CENTER = 1,
	E_LCD_TEXT_ALIGN_RIGHT = 2
} lcd_text_align_e_t;

#define ALIGN_LEFT E_LCD_TEXT_ALIGN_LEFT 
#define ALIGN_CENTER E_LCD_TEXT_ALIGN_CENTER
#define ALIGN_RIGHT E_LCD_TEXT_ALIGN_RIGHT 

void set_text_align(std::int16_t line, lcd_text_align_e_t align);
```

Use example:
```cpp
pros::lcd::print(0, "Title");
pros::lcd::set_text_align(0, ALIGN_CENTER);
```

PR #144, opened 2019-05-14
⬆️ Upgrade LVGL to 5.3

### LINKED ISSUE #92: Upgrade to LVGL 5.3
#### Expected Behavior:

LVGL is the latest version.

#### Actual Behavior:

LVGL is not the latest version.

#### Steps to reproduce:

Try using 5.2 and 5.3 features (LVGL docs have all been updated to 5.3-specific info) with PROS 😃 

#### System information:
Platform: V5 
PROS Kernel Version: any 

#### Additional Information

This was previously blocked on waiting for a full version of 5.2 to be released, which it [now has been](https://github.com/littlevgl/lvgl/releases/tag/v5.2)

Also apparently 5.3 has been released since the creation of this issue, so we'll just skip 5.2 entirely

#### Screenshots/Output Dumps/Stack Traces
N/A

HotelCalifornia:
#### Summary:
Upgrades LVGL to 5.3 release

#### Motivation:
everyone reeeealllly wants this

##### References (optional):
Should unblock #110 and by extension #93
Closes #92 

#### Test Plan:
- [x] let people play with it i guess


HotelCalifornia:
everything compiles right now, but important things (namely LLEMU) appear to be broken.

HotelCalifornia:
example of breakage:

2019-03-28-151855_480x272_pros_capture

HotelCalifornia:
attached is a kernel template built on commit bd6bb46 for anyone who's interested in helping debug (e.g. @Jerrylum)
[kernel@3.1.7-commit.13.bd6bb46.zip](https://github.com/purduesigbots/pros/files/3177340/kernel%403.1.7-commit.13.bd6bb46.zip)

Jerrylum:
Thank you @HotelCalifornia .

Before:
before3 1 7

After:
after3 1 7

There is a big difference after I updated to 3.1.7@develop about the text. I want to know why. Is the default font changed?

HotelCalifornia:
Ah, yes. One of the things we get in this upgrade is support for monospaced fonts, which I believe I set as the default. You should be able to switch back to the old (not monospaced) font with a call to `lv_set_font` or whatever it is (can't remember off the top of my head)

HotelCalifornia:
the symbol should be `lv_font_dejavu_20` (changing the number for the size you'd like. Monosapced fonts can be used with `pros_font_dejavu_mono_20` and friends.

Jerrylum:
OIC

https://github.com/purduesigbots/pros/blob/feature/upgrade-lvgl/include/display/lv_conf.h#L135

Jerrylum:
image

I fixed the problem but I was using `pros_dejavu`. What's happening?

HotelCalifornia:
per @edjubuh:

>For sake of backwards compatibility, I think the monospace should be turned on by default only when you initialize LLEMU

That's a good point. What I'll probably end up doing is just keeping the default font the same as it is, and changing LLEMU styles to explicitly use the monospaced version

Jerrylum:
You can solve the problem by adding a delay at the first line of initialize function.
image
According Lines 43 - 45 in main.c file:
https://github.com/purduesigbots/pros/blob/develop/src/system/main.c#L43-L45
It is because the user 'initialize' function run before the LVGL init and cause the problem. LVGL will display a empty screen (no object) in init function, and it is gray. This keep working in the old version and I don't know why it is malfunction now.

I think we should be initializing LVGL before the system daemon and the problem should be resolved.

HotelCalifornia:
Thanks for the investigation, I'll give it a try at some point in the next few days or so.

You're right that we should probably be initializing LVGL before running the competition tasks... I'm not entirely sure why it was working before but it seems to have been a fluke.

HotelCalifornia:
@Jerrylum thanks again for the help, changing the order of initialization does seem to have fixed it

2019-05-31-132534_480x272_pros_capture

HotelCalifornia:
incidentally, now that we actually have a monospace font, we can finally impose a hard limit on the number of characters that can be printed on the LLEMU screen (33). this also means that we can get rid of the call to `vasprintf` (lol).

before:
2019-05-31-133123_480x272_pros_capture

after:
2019-05-31-133457_480x272_pros_capture

HotelCalifornia:
I'm attaching a copy of the kernel from this PR (as of 890739f) so people can play around with it and tell me what's broken

[kernel@3.1.7-pr144.629.zip](https://github.com/purduesigbots/pros/files/3239736/kernel%403.1.7-pr144.629.zip)

HotelCalifornia:
well this is... uh... progress?

```cpp
lv_obj_t* canvas;
static void* canvas_buf;
void initialize() {
	canvas = lv_canvas_create(lv_scr_act(), NULL);
	canvas_buf = malloc((lv_img_color_format_get_px_size(LV_IMG_CF_TRUE_COLOR) * 480 * 240)/ 8);
	lv_canvas_set_buffer(canvas, canvas_buf, 480, 240, LV_IMG_CF_TRUE_COLOR);
	lv_canvas_draw_circle(canvas, 240, 120, 60, lv_color_t{{0xFF,0xFF,0x00}}); // yellow
}
```

2019-05-31-140044_480x272_pros_capture

SpencerJ21:
I tested my screenlib on this, pretty much everything was the same.
The only exception is with text, which is printing slightly differently.

The text is printing a single 5x7 character on a 10x10 obj. If I remove everything except for aligning the label to the center of the cube, I get this branch printing 4 pixels right of the existing kernel. 
Of course, it is possible this is a change with how aligning works or something with lvgl itself, I'll have to investigate more.

Existing kernel
Existing kernel

This branch
This branch

HotelCalifornia:
>The only exception is with text, which is printing slightly differently.

What font were you using?

SpencerJ21:
[This one here](https://github.com/SpencerJ21/screenlib/blob/master/src/screen/smallFont.c)
If you’re planning on testing with it, know it only has the characters for numbers.

HotelCalifornia:
Hm ok yeah. I'm pretty sure there were some changes made to the way that fonts are rendered (judging by the addition of support for monospace fonts). I wonder what would happen if you explicitly configured that font as such? See https://github.com/purduesigbots/pros/blob/890739f29170fb5591bb6ee57ea01860cd70d13f/src/display/lv_fonts/pros_font_dejavu_mono_10.c#L5380

theol0403:
First thing I noticed when I upgraded - my splashscreen image broke. Probably because the image was generated using the old convertor.

Before:
2019-05-31-225035_480x272_pros_capture
2019-05-31-225127_480x272_pros_capture
After:
2019-05-31-225931_480x272_pros_capture
2019-05-31-225949_480x272_pros_capture

Not much changed, it looks good. The only difference I noticed was under the angler page, in the button matrix on the left, the bottom button was expanded to fill the container. Most likely a bug fix.

I will get the image working, but apart from that the upgrade looks successful.

HotelCalifornia:
Hmmm... The image thing alone might warrant bumping to 3.2.0

theol0403:
I managed to get the image to work. However, it seems only `True Color with Alpha` gave me the results I needed, and it was a massive download.
`Raw` gave a "no data" error, and all the chroma-keyed ones had transparency problems.
Probably not a bug though, but it might be a good idea to suggest the best format in the docs.

theol0403:
Working splashscreen with transparency.
2019-05-31-232029_480x272_pros_capture

I don't think I have anything else to test, aside from upgrading the image format, everything else worked properly.

@HotelCalifornia do you want me to try to use the canvas?

HotelCalifornia:
I was also getting the "no data" thing when setting raw color format on a canvas. I think I saw something about needing a custom decoder for that.

That said, I had some ok results when using the indexed 8-bit format-- did you try that? Also I think @Jerrylum had some success with writing a bmp decoder that used 32-bit RGBA (which may actually be the same thing now I think about it)

theol0403:
I did not try any indexed formats, but does it provide the transparency I need? My image is white, and the container background is orange, which is why the transparency is needed.

When I have time I will test the reading from SD card. Last time I tried I couldn't get it to work (it just gave `no data` no matter the path), and I needed to be able to compile LVGL on my own to enable the debugging options.

I know you have thought about it already, but I really think it is a good idea to provide LVGL as a template, which people can build on their own, and people can uninstall it and opt to install another template which contains the vexos screen wrappers.

HotelCalifornia:
>I did not try any indexed formats, but does it provide the transparency I need? My image is white, and the container background is orange, which is why the transparency is needed.

Ah, ok. Assuming that the 8-bit indexed format is actually RGBA, then yes, the "A" refers to an alpha (read: transparency) channel.

>I know you have thought about it already, but I really think it is a good idea to provide LVGL as a template, which people can build on their own, and people can uninstall it and opt to install another template which contains the vexos screen wrappers.

Yeah I have thought about it. There are a couple problems:
- it's arguably better to have LVGL compiled in with the kernel:
  - so that it has a bit lower-level access (kernel heap, etc)
  - and requires basically zero configuration on the part of the user
- at this point removing LVGL from the kernel would be a really big breaking change

I suppose I could build a copy of the kernel with LVGL debugging turned on, if you think that would help

Jerrylum:
I tested the LVGL Canvas feature and made a BMP image display function.
The function can display RGB (24 bit) .bmp and RGBA (32 bit) .bmp file.

This is the code:
https://gist.github.com/Jerrylum/5d4f3168a596dfbcd87e8d5f81236fcc

The result:
2019-06-01-145200_480x272_pros_capture
Showing "pros-icon-rgba.bmp", the picture is bigger then the screen

The image I used in the code:
- pros-icon.bmp (with white background, R8:G8:B8)
- pros-icon-rgba.bmp (with transparent background, A8:R8:G8:B8)

Size: 280px X 280px

Download:
[images.zip](https://github.com/purduesigbots/pros/files/3243506/images.zip)

Tools to made image: Pixelformer

include/api.h
```diff
@@ -41,8 +41,8 @@
 
 #define PROS_VERSION_MAJOR 3
 #define PROS_VERSION_MINOR 1
-#define PROS_VERSION_PATCH 6
-#define PROS_VERSION_STRING "3.1.6"
+#define PROS_VERSION_PATCH 7
+#define PROS_VERSION_STRING "3.1.7-dirty.17.1598f4d"
```
baylessj: I'm assuming this isn't intentional, right?

version
```diff
@@ -1 +1 @@
-3.1.6
+3.1.7-dirty.17.1598f4d
```
baylessj: same here

src/system/startup.c
```diff
@@ -40,10 +40,10 @@ __attribute__((constructor(101))) static void pros_init(void) {
 
 	vdml_initialize();
 
-	system_daemon_initialize();
-
 	display_initialize();
 
+	system_daemon_initialize();
```
baylessj: Can you link to this issue or throw an explanation comment here for the re-ordering?

HotelCalifornia: explained in this very PR! 😀 

https://github.com/purduesigbots/pros/pull/144/files/e5783e32cc447bd0542e5e27baa3376e90c87234#issuecomment-496901942

HotelCalifornia: explained in this very PR! 😁 

https://github.com/purduesigbots/pros/pull/144/#issuecomment-496901942

baylessj: Right, sorry for not being more clear. I'm thinking it would be a good idea to put this as a comment in the code itself to signal to any future devs that they shouldn't reorder this particular section without a lot of testing

PR #147, opened 2019-06-24
🤫 Suppress ignored qualifiers when including lvgl

Octogonapus:
#### Summary:
This PR pushes `ignored "-Wignored-qualifiers"` when including the LVGL headers.

#### Motivation:
The LVGL headers cause a lot of noise via compiler warnings. This PR silences those warnings.

#### Test Plan:
- [x] Compile and verify no warnings are generated by LVGL headers


HotelCalifornia:
I'd like to see where we're at after #144 lands

PR #151, opened 2019-07-16
🐛 Fix multiple ADI bugs and inconsistencies

### LINKED ISSUE #115: Errno not set for ADI port configuration mismatch
#### Expected Behavior:
Errno will be set to `EINVAL` when a port's configuration is different than the action applied to the port would expect.

#### Actual Behavior:
For some, but not all, of the ADI configuration checks, errno is not set.

Example: https://github.com/purduesigbots/pros/blob/develop/src/devices/vdml_adi.c#L123

#### Steps to reproduce:
Configure an ADI port as one type, e.g. a gyroscope, and then perform an action on said port that requires a different configuration, e.g. analog_read.

#### System information:
Platform: V5
PROS Kernel Version: >3.1.0

#### Additional Information
Made debugging more difficult (most likely) in this forum thread: https://www.vexforum.com/t/gyro-suddenly-stopped-working/56648/20

### LINKED ISSUE #148: Missing braces in validate_twowire
This `else` block is missing braces: https://github.com/purduesigbots/pros/blob/5bb79abf27375665bbf6f0078111bea2c6e7a814/src/devices/vdml_adi.c#L156

### LINKED ISSUE #149: ADIEncoder does not store _port correctly
#### Expected Behavior:
The `ADIEncoder` dtor should set its port to undefined (technically the `ADIPort` dtor).

#### Actual Behavior:
The dtor sets `_port - 1` to undefined instead of `_port`.

#### Steps to reproduce:
```cpp
void opcontrol() {
  pros::ADIEncoder enc{'A', 'B'};
  enc = {'C', 'D'};
  std::cout << enc.get_value() << std::endl;
  std::cout << errno << std::endl;
}
```

```
kernel set config 0 to 14
errno from inside kernel 0
kernel set config 2 to 14
errno from inside kernel 0
ADIPort dtor setting port 2 to UNDEFINED
kernel set config 1 to 255
0
0
ADIPort dtor setting port 2 to UNDEFINED
kernel set config 1 to 255
```

Pay special attention to
```
ADIPort dtor setting port 2 to UNDEFINED
kernel set config 1 to 255
```

#### System information:
Platform: V5
PROS Kernel Version: 3.1.6

#### Additional Information
The problem is that `_port` is transformed *twice* because it's transformed once in the ctor `_port = adi_encoder_init(port_top, port_bottom, reversed);` and again in the dtor inside `adi_port_set_config(_port, E_ADI_TYPE_UNDEFINED);`. I recommend that `_port` be initialized as `_port = adi_encoder_init(port_top, port_bottom, reversed) + 1;` similar to how `ADIUltrasonic` works; `reset()` and `get_value()` then need to be refactoed to subtract 1 from the port number.

Separately, this hints at a larger issue: `ADIPort` subclasses are effectively not copyable because the dtor side-effects the port by changing its configuration. To fix this, either:
1. The `ADIPort` dtor should not side-effect the port configuration (I and others prefer this)
2. `ADIPort` should have copying disabled

TixoRebel:
#### Summary:
The ADI C API has been modified to be more self consistent. `adi_encoder_t`, `adi_ultrasonic_t` and `adi_gyro_t` now represent the user supplied port (1 through 8 instead of 0 through 7) so that functions like `adi_encoder_get` can accept either the return value from the device initialization or a user supplied port number. `errno` is now properly set during ADI port validation failures and other small bugs were fixed. The ADI C++ destructor was changed to not clear the port configuration to make it similar to other PROS API's where multiple instances can be created and destroyed without side effects.

#### Motivation:
Some functionality did not match what a user would expect and in some cases did not match the PROS documentation.

##### References:
closes #115 
closes #148 
closes #149 

#### Test Plan:
- [x] Run some ADI tests

### DIFF
--- a/include/pros/adi.hpp
+++ b/include/pros/adi.hpp
@@ -39,7 +39,7 @@ class ADIPort {
 	 */
 	ADIPort(std::uint8_t port, adi_port_config_e_t type = E_ADI_TYPE_UNDEFINED);
 
-	virtual ~ADIPort(void);
+	virtual ~ADIPort(void) = default;
 
 	/**
 	 * Gets the configuration for the given ADI port.
--- a/include/pros/adi.hpp
+++ b/include/pros/adi.hpp
@@ -39,6 +39,8 @@ class ADIPort {
 	 */
 	ADIPort(std::uint8_t port, adi_port_config_e_t type = E_ADI_TYPE_UNDEFINED);
 
+	virtual ~ADIPort(void);
+
 	/**
 	 * Gets the configuration for the given ADI port.
 	 *
--- a/include/pros/adi.hpp
+++ b/include/pros/adi.hpp
@@ -39,8 +39,6 @@ class ADIPort {
 	 */
 	ADIPort(std::uint8_t port, adi_port_config_e_t type = E_ADI_TYPE_UNDEFINED);
 
-	virtual ~ADIPort(void);
-
 	/**
 	 * Gets the configuration for the given ADI port.
 	 *
@@ -473,8 +471,6 @@ class ADIGyro : private ADIPort {
 	 *        supplied by the ADI
 	 */
 	ADIGyro(std::uint8_t port, double multiplier = 1);
-
-	~ADIGyro(void) override;
 
 	/**
 	 * Gets the current gyro angle in tenths of a degree. Unless a multiplier is
--- a/include/vdml/registry.h
+++ b/include/vdml/registry.h
@@ -24,7 +24,6 @@ extern "C" {
 typedef struct {
 	v5_device_e_t device_type;
 	V5_DeviceT device_info;
-	// HACK: Make pad a union of all possible types that can be stored
 	uint8_t pad[128];  // 16 bytes in adi_data_s_t times 8 ADI Ports = 128
 } v5_smart_device_s_t;
 
--- a/include/vdml/registry.h
+++ b/include/vdml/registry.h
@@ -24,6 +24,7 @@ extern "C" {
 typedef struct {
 	v5_device_e_t device_type;
 	V5_DeviceT device_info;
+	// HACK: Make pad a union of all possible types that can be stored
 	uint8_t pad[128];  // 16 bytes in adi_data_s_t times 8 ADI Ports = 128
 } v5_smart_device_s_t;
 
--- a/include/vdml/vdml.h
+++ b/include/vdml/vdml.h
@@ -33,7 +33,8 @@
  * If a mutex cannot be taken, errno is set to EACCES (access denied) and
  * returns.
  * 
- * FIXME: If this is used in a function that returns a double, PROS_ERR_F should be returned
+ * This and other similar macros should only be used in functions that return
+ * int32_t as PROS_ERR could be returned.
  *
  * \param port
  *        The V5 port number from 0-20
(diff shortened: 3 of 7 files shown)


include/pros/adi.hpp
```diff
@@ -39,8 +39,6 @@ class ADIPort {
 	 */
 	ADIPort(std::uint8_t port, adi_port_config_e_t type = E_ADI_TYPE_UNDEFINED);
 
-	virtual ~ADIPort(void);
```
Octogonapus: I think this should still have a virtual default dtor.

TixoRebel: Ya, you're right

include/pros/adi.hpp
```diff
@@ -39,6 +39,8 @@ class ADIPort {
 	 */
 	ADIPort(std::uint8_t port, adi_port_config_e_t type = E_ADI_TYPE_UNDEFINED);
 
+	virtual ~ADIPort(void);
```
Octogonapus: You should declare it as `= default`.

TixoRebel: Is there any difference between doing that and having an empty default implementation?

TixoRebel: Stackoverflow would seem to indicate that yes, there is a slight difference, so I'll mark is as `default`.

include/vdml/registry.h
```diff
@@ -24,6 +24,7 @@ extern "C" {
 typedef struct {
 	v5_device_e_t device_type;
 	V5_DeviceT device_info;
+	// HACK: Make pad a union of all possible types that can be stored
```
HotelCalifornia: I don't think this is necessarily a "hack"...

include/vdml/vdml.h
```diff
@@ -32,6 +32,8 @@
  * If a port isn't yet registered, it registered as a motor automatically.
  * If a mutex cannot be taken, errno is set to EACCES (access denied) and
  * returns.
+ * 
+ * FIXME: If this is used in a function that returns a double, PROS_ERR_F should be returned
```
HotelCalifornia: Is this necessary to note? Either make versions that return floating point error values, or just leave in the workaround you had below for the gyro.

HotelCalifornia:
Also pending tests of course

include/vdml/registry.h
TixoRebel: I guess so; the reason I flagged it as such was because if the size of the ADI structure ever grows or if anything else that's bigger is ever stored in it, it'll have to be manually adjusted.

Octogonapus: Can't the compiler pad it for you?

TixoRebel: What do you mean?

Octogonapus: With the alignment attribute: https://gcc.gnu.org/onlinedocs/gcc/Common-Variable-Attributes.html#Common-Variable-Attributes

TixoRebel: The note has been removed, it's probably best just to leave it unchanged as at least what's in there now works and shouldn't break in the future.

include/vdml/vdml.h
TixoRebel: I've changed the message to a simple note that it should only be used by a function that returns `int32_t` to hopefully reduce the chances of someone making that mistake in the future.

PR #152, opened 2019-07-27
🏗 Combine template source files

Octogonapus:
#### Summary:
This PR merges `initialize.cpp`, `autonomous.cpp`, and `opcontrol.cpp` into one file named `main.cpp`. To avoid confusion, `main.c` has been renamed to `startup.c`.

#### Motivation:
The three template source files have been merged because many users struggle to use `extern` in their programs. Combining these files will give novice users a better experience.

#### Test Plan:

- [x] Make a new project using the new template and verify it compiles, links, and runs correctly


theol0403:
I agree with this. When setting up a project, one of the first things I usually do is merge the files together. I also changed the comments a bit and added ascii art.
Here is what I used: 
https://gist.github.com/theol0403/944f6e69ce19835b0a743b00b9ba2e3a
You are welcome to use that formatting.

Also, it would be good if a simple extern explanation or link could be added to the FAQ.

edjubuh:
I think this PR should be split into 2: one to change the default behavior and one to create the combined files.

I'm not opposed to changing hot/cold to being the default behavior. We haven't been receiving many reports of issues.

I'm somewhat concerned about merging user files into one. It's all too common to find really messy one-file projects that are tough to work through when providing user support. Splitting into multiple files encourages thinking about code organization.

Octogonapus:
> I think this PR should be split into 2: one to change the default behavior and one to create the combined files.

Ok, I can split it.

> Splitting into multiple files encourages thinking about code organization.

Users are struggling with multiple files. I see at least one person per day not knowing how to "share variables". Users that already understand this or feel the need to split up their program into multiple source files will still be able to.

HotelCalifornia:
> > Splitting into multiple files encourages thinking about code organization.
>
> Users are struggling with multiple files. I see at least one person per day not knowing how to "share variables". Users that already understand this or feel the need to split up their program into multiple source files will still be able to.

Then we'll just add something to the docs to explain that they can split things up if they want to, because it may be less clear after this.

Octogonapus:
Can we merge this?

PR #154, opened 2019-08-04
🔧 Enable hot/cold by default

Octogonapus:
#### Summary:
This PR enables hot/cold by default.

#### Motivation:
Wireless uploading is optimized on other platforms by default. PROS should match this behavior.

#### Test Plan:

- [x] Make a new project and verify hot/cold is enabled by default.


Octogonapus:
Can we merge this?

Issue #146, opened 2019-06-16
Can't compile on Arch Linux


IrregularWombat:
#### Expected Behavior:

For the code to compile with out any errors and give me a binary which can be uploaded to a V5 Brain.

#### Actual Behavior:

Fails to compile giving me the error `Linking project with libpros,okapilib [ERRORS]`

#### Steps to reproduce:

Install latest version of Arch Linux, install relevant ARM and GCC packages, use `prosv5 conductor new-project ./test` code fails to compile.

#### System information:
Platform:  V5
PROS Kernel Version: 
```
PROS Project for v5 at: /home/milo/help (help)
Name      Version    Origin
--------  ---------  -------------
kernel    3.1.6      pros-mainline
okapilib  3.3.13     pros-mainline
```

#### Additional Information

This was happening on my system and after a while of messing around I was not able to fix it, I reinstall a fresh copy Arch Linux and still get this problem. I am able to compile other pieces of code for ARM on this machine as well.
My mate was able to compile just fine on a Ubuntu system, perhaps it may be that Arch has newer version of ARM GCC compilers which cause this error?

#### Screenshots/Output Dumps/Stack Traces
Level 0 Verbosity from `prosv5 make`
```
Adding timestamp [OK]
Linking project with libpros,okapilib [ERRORS]
/usr/lib/gcc/arm-none-eabi/9.1.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/9.1.0/../../../../arm-none-eabi/lib/libstdc++.a(locale.o): in function `std::locale::_Impl::_M_install_cache(std::locale::facet const*, unsigned int)':
locale.cc:(.text._ZNSt6locale5_Impl16_M_install_cacheEPKNS_5facetEj+0x18): undefined reference to `__sync_synchronize'
/usr/lib/gcc/arm-none-eabi/9.1.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/9.1.0/../../../../arm-none-eabi/lib/libstdc++.a(locale_init.o): in function `(anonymous namespace)::get_locale_mutex()':
locale_init.cc:(.text._ZN12_GLOBAL__N_116get_locale_mutexEv+0xc): undefined reference to `__sync_synchronize'
[...]

Issue #134, opened 2019-03-17
GCC 8 error with __sync_synchronize


lights0123:
#### Expected Behavior:
Successful compilation

#### Actual Behavior:
An error message referencing `__sync_synchronize` appears (see below).

#### Steps to reproduce:
I had created a new project with `prosv5 conductor new-project test`, after installing `arm-none-eabi-gcc 8.3.0`.

#### System information:
Platform: V5
PROS Kernel Version:

|  Name  | Version |       Origin       |
|-----------|-----------|-------------------|
|  kernel  |   3.1.6   | pros-mainline |
| okapilib |  3.3.12  | pros-mainline |

#### Additional Information
I am running Arch Linux, using `arm-none-eabi-gcc (Arch Repository) 8.3.0`. Is this a GCC 8 issue? If so, that should be fixed or added to the documentation.

#### Screenshots/Output Dumps/Stack Traces
```c++
❱ make
Adding timestamp [OK]
Linking project with libpros,okapilib [ERRORS]
/usr/lib/gcc/arm-none-eabi/8.3.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/8.3.0/../../../../arm-none-eabi/lib/libstdc++.a(cxx11-ios_failure.o): in function `(anonymous namespace)::__io_category_instance()':
cxx11-ios_failure.cc:(.text._ZN12_GLOBAL__N_122__io_category_instanceEv+0xc): undefined reference to `__sync_synchronize'
/usr/lib/gcc/arm-none-eabi/8.3.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/8.3.0/../../../../arm-none-eabi/lib/libstdc++.a(locale_init.o): in function `(anonymous namespace)::get_locale_mutex()':
locale_init.cc:(.text._ZN12_GLOBAL__N_116get_locale_mutexEv+0xc): undefined reference to `__sync_synchronize'
/usr/lib/gcc/arm-none-eabi/8.3.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/8.3.0/../../../../arm-none-eabi/lib/libstdc++.a(locale.o): in function `std::locale::_Impl::_M_install_cache(std::locale::facet const*, unsigned int)':
locale.cc:(.text._ZNSt6locale5_Impl16_M_install_cacheEPKNS_5facetEj+0x18): undefined reference to `__sync_synchronize'
[...]

PR #155, opened 2019-08-12
🚨 Make things work with GCC 8.3

### LINKED ISSUE #153: Linking fails with Arm Toolchain 8-2019-q3-update Linux 64-bit
#### Expected Behavior:
Ran prosv5 make all and expected to compile into hot and cold image.

#### Actual Behavior:
Compiling of source files works until attempting to link with libpros and libokapi. The same set of files compile with the previous 2018 December Arm Toolchain.

#### Steps to reproduce:
1. Download the new Linux 64-bit toolchain from [Arm Developers](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-rm/downloads)
2. Extract and move the folder to /opt/arm-gcc-none-eabi/ replacing old toolchain
3. Export to PATH using ~/.profile and sourcing the file.
4. Make a new PROS project with conductor.
5. Run prosv5 make in project directory (both with and without hot and cold linking yields the same error).

#### System information:
Manjaro 18.0.4 (4.19.60-1-MANJARO)

PROS Version: pros, version 3.1.4

#### Additional Information
Ran on a clean install with no other conflicting packages and whereis points to the extracted toolchain. The linking failure occurs with the default PROS project source files too.
Python 3.7.3 used and PROS installed using pip3 without --user flag.

#### Screenshots/Output Dumps/Stack Traces
image

HotelCalifornia:
#### Summary:

arm-none-eabi-8-2019-q3-update was released, which finally bumped arm-none-eabi-gcc to 8.3 officially, so we started running into the missing `__sync_synchronize` symbol as well. This PR fixes that, as well as a new multiple definitions issue (`__malloc_lock`/`unlock`).

The solution to the `__sync_synchronize` issue is described in https://github.com/purduesigbots/pros/issues/153#issuecomment-519335375.

`__malloc_lock` and `__malloc_unlock` were moved from src/system/newlib_stubs.c to their own file src/system/mlock.c, which [apparently is how newlib determines overrides](http://sourceware-org.1504.n7.nabble.com/malloc-lock-tp179955p179956.html).

#### Motivation:

Since the latest official version of the toolchain now includes 8.3, we can't really just tell people that they should use the official one instead of the preview build arch package managers had access to. We should also be endeavoring to build the kernel against the latest version of the toolchain anyway.

##### References (optional):

fixes #153, ref #146, ref #134, forum posts: https://www.vexforum.com/t/pros-programming-help/64328 https://www.vexforum.com/t/problem-with-building-files-in-pros-on-a-mac/64326, at least one email

#### Test Plan:

- [x] compiles with no errors or warnings against arm-none-eabi-8-2019-q3-update
- [x] compiles with no errors or warnings against an older version of the toolchain

### DIFF
--- a/src/system/cpp_support.cpp
+++ b/src/system/cpp_support.cpp
@@ -25,7 +25,7 @@ extern "C" void task_fn_wrapper(task_fn_t fn, void* args) {
 #endif
 		fn(args);
 #ifdef __cpp_exceptions
-	} catch (std::runtime_error re) {
+	} catch (std::runtime_error& re) {
 		vexDisplayString(7, "caught runtime error: %s", re.what());
 	} catch (...) {
 		vexDisplayString(7, "caught an unknown error");
--- a/src/system/mlock.c
+++ b/src/system/mlock.c
@@ -1,0 +1,24 @@
+/**
+ * \file system/mlock.c
+ *
+ * memory lock newlib stubs
+ *
+ * Contains implementations of memory-locking functions for newlib.
+ *
+ * Copyright (c) 2017-2019, Purdue University ACM SIGBots.
+ * All rights reserved.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include "rtos/task.h"
+
+ void __malloc_lock(void) {
+ 	rtos_suspend_all();
+ }
+
+ void __malloc_unlock(void) {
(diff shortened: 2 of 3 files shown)


MayorMonty:
For those looking for temporary solutions until this gets merged into master, you can define your own stub to rid the `__sync_synchronize` build errors. Create `src/stub.c`, and add the following code:
```
void __sync_synchronize(void)
{
    __sync_synchronize();
}
```

Delete your bin and run `pros make` again, and the issue should be resolved.

Octogonapus:
Can we merge this? Looks like the testing you wanted to do is done, and other people have been using this fix with success.

PR #156, opened 2019-08-23
Add default `.gitignore` to template

theol0403:
#### Summary:
Add a default `.gitignore` file to the project template.

#### Motivation:
Some beginner programmers don't know how to use git, and fail to add a `.gitignore` file to their project before creating a repository. Creating a repository is relatively simple, so it is likely that someone who does not know about `.gitignore` will still be able to create and commit to a repository.
PROS and cquery generate a lot of junk files, which get tracked by git if not explicitly ignored. This leads to programmers having ~500 files in their staging area each time they commit.

#### Known Problem:
For some reason, the file does not like being marked as a user file.
I have `--user ".gitignore"` in the `CREATE_TEMPLATE_ARGS` flag, and in the generated `project.pros` the file is there:
image
However, when I apply the template to a project, the `.gitignore` file gets overwritten, but `Makefile` for example does not get touched (as user files shouldn't).
Is it possible the weird naming of `.gitignore` is messing with the file detection?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ EXCLUDE_SRC_FROM_LIB+=$(foreach file, $(SRCDIR)/opcontrol $(SRCDIR)/initialize $
 TEMPLATE_FILES=$(ROOT)/common.mk $(FWDIR)/v5.ld $(FWDIR)/v5-common.ld $(FWDIR)/v5-hot.ld
 TEMPLATE_FILES+= $(INCDIR)/api.h $(INCDIR)/main.h $(INCDIR)/pros/*.* $(INCDIR)/display
 TEMPLATE_FILES+= $(SRCDIR)/opcontrol.cpp $(SRCDIR)/initialize.cpp $(SRCDIR)/autonomous.cpp
+TEMPLATE_FILES+= $(ROOT)/template-gitignore
 
 PATCHED_SDK=$(FWDIR)/libv5rts/sdk/vexv5/libv5rts.patched.a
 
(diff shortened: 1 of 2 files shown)


Octogonapus:
I'm not sure if conductor supports what we want here. The `.gitignore` file should be added once as part of a new project and then never touched again. @HotelCalifornia 	Do you know if we can get this behavior?

HotelCalifornia:
> I'm not sure if conductor supports what we want here. The `.gitignore` file should be added once as part of a new project and then never touched again. @HotelCalifornia Do you know if we can get this behavior?

yes, that's what it should be doing as a user file. user files are not touched on upgrade unless told to (the `--force-user` flag or whatever it is).

HotelCalifornia:
@theol0403 can you elaborate on the issues you were running into?

theol0403:
Humm that's weird. Here is what was happening for me:
1. Build template
2. `prosv5 c purge-template kernel` to remove any other conflicting templates
2. Fetch new template to conductor
3. Create new project from template
4. Modify .gitignore
5. Modify Makefile
6. `prosv5 c u kernel --force-apply` to reapply the template
7. .gitignore file gets reverted to default
8. as a contrast changes to Makefile remain

HotelCalifornia:
Well yeah, you were telling it to force application. The Makefile should also have been overwritten-- how were you verifying that it wasn't?

theol0403:
I thought `--force-apply` simply removes the check that the template is already installed, and re-applies the template regardless of previous version. At least, this is the behavior I've noticed (such as with the Makefile and other user files) until this thing with the .gitignore.
To reapply the user files, I thought you had to pass `--force-user`.
image

HotelCalifornia:
yeah, that's what i meant. I'll have to (or you can) look at the CLI code governing this a bit more closely then I guess.

HotelCalifornia:
for what it's worth, not using the `--force-apply` flag seemed to work, as I mentioned above

theol0403:
I'm pretty unfamiliar with the CLI, but I can take a quick look.
It will probably have to be one of you guys to figure this out.

theol0403:
This problem is minor enough, its probably okay to merge the PR.
It's just people might find it annoying that their `.gitignore` gets replaced when they call `--force-apply`, but those people who already use a gitignore probably know how to use git to discard the change anyways =).

PR #157, opened 2019-08-23
Untrack 'version'

theol0403:
#### Summary:
This file is generated each time you compile. It bothers me because I always have to reset or stash the working directory when switching branches. This PR simply deletes and untracks the file, causing the existing rule in the `.gitignore` to kick in and make the file disappear forever.
I hope I'm not being silly and this file is actually important. Also, I'm new to PRs so let me know if this is way too insignificant to waste your time for.


HotelCalifornia:
This file is actually important. It is used for template version generation (and for filling in the version string info in api.h).

PR #158, opened 2019-08-29
🚸 Add alternate constructor for pros::Task

SpencerJ21:
#### Summary:
Adds an alternate constructor to pros::Task that omits task priority and stack depth, while keeping the task function, parameter, and name.

#### Motivation:
Modifying priority and stack depth is not always used, however users such as myself like to always have task names set. This alternate constructor will assume defaults without needing them specified.

#### Test Plan:
There is no reason why this should not work, as it just passing to task_create, but any tests that apply to standard tasks could be easily adapted to ensure this works.


HotelCalifornia:
apparently I can't read. merging now.

PR #163, opened 2019-08-31
💚 Update toolchain installation for azure

HotelCalifornia:
the PPA was hanging, and it installs an old version anyway

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -16,7 +16,7 @@ jobs:
   - bash: |
       curl -LSso toolchain.tar.bz2 $(toolchain)
       tar -xjvf toolchain.tar.bz2
-      echo "##vso[task.prependpath]$(pwd)/toolchain/bin"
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -10,9 +14,9 @@ jobs:
       sshPublicKey: $(gh.publickey)
   - checkout: self
   - bash: |
-      sudo add-apt-repository ppa:team-gcc-arm-embedded/ppa
-      sudo apt-get update
-      sudo apt-get install gcc-arm-embedded
+      curl -LSso toolchain.tar.bz2 $(toolchain)
+      tar -xjvf toolchain.tar.bz2
+      echo "##vso[task.prependpath]$(pwd)/toolchain/bin"
```
edjubuh:
```suggestion
      echo "##vso[task.prependpath]$(pwd)/gcc-arm-none-eabi-$(toolchain_update)-update/bin"
```

HotelCalifornia:
going to investigate using a docker image, as it seems azure can cache build environments that way

PR #165, opened 2019-09-03
🐛 Add default LLEMU init code back to main.cpp

### LINKED ISSUE #164: 🐛 LLEMU initialization code removed in #152
#### Expected Behavior:
LLEMU initialization code should be present in `initialize()`

#### Actual Behavior:

It was accidentally removed in #152

HotelCalifornia:
#### Summary:

Fixes a regression introduced in #152 whereby the default LLEMU initialization code was removed.

#### Motivation:

Code in `opcontrol()` that displays information to LLEMU was not removed, so users will (reasonably) expect the code to Just Work™ out of the box.

##### References (optional):

closes #164 

#### Test Plan:

- [x] Code compiles
- [x] Default project displays LLEMU as expected

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -1,5 +1,15 @@
 #include "main.h"
 
+void on_center_button() {
+	static bool pressed = false;
+	pressed = !pressed;
+	if (pressed) {
+		pros::lcd::set_text(2, "I was pressed!");
+	} else {
+		pros::lcd::clear_line(2);
+	}
+}
+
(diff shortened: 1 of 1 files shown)


src/main.cpp
```diff
@@ -1,21 +1,34 @@
 #include "main.h"
 
+void on_center_button() {
```
edjubuh: Should add a docstring explaining what this function is

PR #167, opened 2019-09-05
Recompile on header change

nathan-moore:
#### Summary:
When you change a header file, every src file depending on that header will now recompile.

This follows [this](http://make.mad-scientist.net/papers/advanced-auto-dependency-generation/) for having gcc auto generate makefiles for each src file.

#### Motivation:
This makes development easier for header only changes.

#### Test Plan:

- [ ] Compiles pros
- [ ] Compiles forkner
- [ ] Changing a header causes recompilation.


theol0403:
This should be a useful feature.
I recommend you also update the [template-gitignore](https://github.com/purduesigbots/pros/blob/develop/template-gitignore), as that is the gitignore that is shipped with projects.

PR #169, opened 2019-09-13
✨ Provide error if updates are sent too quickly to the controller or if it is not connected

TixoRebel:
#### Summary:
Through some testing I determined that `vexControllerTextSet` returns 0 if it fails to send data to the controller (such as when the user sends updates quicker then every 50ms or if the controller is unplugged). Using this, `controller_set_text`, `controller_print` and `controller_clear_line` have been updated to return `PROS_ERR` on failure setting `errno` to `EAGAIN`.

I also updated `controller_clear` to block for the required 110ms for it to work. I would probably recommend either deprecating this function or implementing a more complex controller system that queues updates and pushes to each line in a cyclic pattern every 50ms, but that's another matter.

#### Motivation:
Would probably be nice to give the user feedback on whether their calls succeed instead of silently failing.

#### Test Plan:
If someone could test this kernel that would be great. I won't have access to a V5 for the next little while.

- [ ] Ensure that the functions return `PROS_ERR` and set `errno` if the updates are sent quicker then 50ms
- [ ] Ensure that the functions do not return `PROS_ERR` and do not set `errno` if the updates are sent slower then 50ms
- [ ] Test `controller_clear`

### DIFF
--- a/include/pros/misc.h
+++ b/include/pros/misc.h
@@ -337,7 +337,8 @@ int32_t controller_clear_line(controller_id_e_t id, uint8_t line);
  * Clears all of the lines on the controller screen.
  *
  * \note Controller text setting is currently in beta, so continuous, fast
- * updates will not work well.
+ * updates will not work well. On vexOS version 1.0.0 this function will block
+ * for 110ms.
  *
  * This function uses the following values of errno when an error state is
  * reached:
--- a/include/pros/misc.h
+++ b/include/pros/misc.h
@@ -337,8 +337,7 @@ int32_t controller_clear_line(controller_id_e_t id, uint8_t line);
  * Clears all of the lines on the controller screen.
  *
  * \note Controller text setting is currently in beta, so continuous, fast
- * updates will not work well. To be able to clear all 3 lines, the function
- * blocks for a total of 110ms.
+ * updates will not work well.
  *
  * This function uses the following values of errno when an error state is
  * reached:
--- a/include/pros/misc.h
+++ b/include/pros/misc.h
@@ -337,7 +337,8 @@ int32_t controller_clear_line(controller_id_e_t id, uint8_t line);
  * Clears all of the lines on the controller screen.
  *
  * \note Controller text setting is currently in beta, so continuous, fast
- * updates will not work well.
+ * updates will not work well. To be able to clear all 3 lines, the function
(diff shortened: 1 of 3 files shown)


theol0403:
What I believe VEXCode does is that they embed delays directly in the functions.
The implementation is already there in PROS, its not that much extra work to simply go all-in and make the delay automatic. I have had many frustrated people come to me asking why setting text does not work.

However, I think it is still important to have the functionality of this PR, to allow users to delay themselves but know if something went wrong. Maybe it would be useful having a defaulted to true optional parameter at the end of each command, that disables automatic delay.

If that suggestion gets implemented, then the `clear` command should delay 3 times so that a command after the clear goes through. 

I didn't realize that was how `clear` was implemented in the past, that explains why there have been so many people frustrated at it not working 🙂 . That qualifies this as a bugfix.

HotelCalifornia:
@TixoRebel I've been informed that, as of vexOS 1.0.1, sending a zero-length string to line zero of the controller LCD will clear it in one operation.

TixoRebel:
> @TixoRebel I've been informed that, as of vexOS 1.0.1, sending a zero-length string to line zero of the controller LCD will clear it in one operation.

That's good to know, I'll change the clear function to use that.

HotelCalifornia:
If you want to be really fancy, you can check ```c
vexSystemVersion() > 0x01000000
```
before using that method, otherwise fall back to the old, slow method

TixoRebel:
> If you want to be really fancy, you can check
> 
> ```c
> vexSystemVersion() > 0x01000000
> ```
> 
> before using that method, otherwise fall back to the old, slow method

Fancy it is!

TixoRebel:
Uhhhhhhhh, I don't think this actually works yet. I only had a brief moment to test it and it didn't seem to be working properly.

nathan-moore:
That's interesting. I tried both clears and they both worked?

nathan-moore:
Can you retest when you have a chance? Here's a build with the change in:
[kernel@3.2.1-dirty.47.7fa5c61.zip](https://github.com/purduesigbots/pros/files/3876939/kernel%403.2.1-dirty.47.7fa5c61.zip)

TixoRebel:
Will do, sorry about the confusion.

PR #170, opened 2019-09-15
✨Add support for extra Make scripts

edjubuh:
#### Summary:
Automatically include any firmware/*.mk scripts in Make system to allow
templates to modify build process (e.g. add extra flags or
dependencies).

#### Motivation:
A library wishes to dynamically generate a C file at Make time based on some text file and include in project's binary. There is currently no support for this. With this modification, library author creates a firmware/mylibrary.mk file which adds the corresponding variable modifications and rules to ensure the C file is generated and compiled as part of the user's project.

#### Test Plan:

- [X] Scripts are included

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -162,6 +162,8 @@ ifeq ($(USE_PACKAGE),1)
 DEFAULT_BIN=$(HOT_BIN)
 endif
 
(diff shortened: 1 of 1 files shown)


Octogonapus:
This looks good, what else is missing before a merge?

PR #177, opened 2019-10-17
🚸 Allow tasks to be constructed from any void() callable

nickmertin:
#### Summary:
Add two constructors to `Task`, analogous to the existing ones, to enable construction from any `void()` callable (including lambdas with captures).

#### Motivation:
This should improve ease of use and intuitive interface for creating Tasks. It is also a C++-idiomatic approach.

#### Test Plan:
Simple tests can be constructed using C++ lambdas with captures. Example code:

```
std::unique_ptr<int> data{new int(7)};
pros::Task task{[=] {
  pros::delay(1000);
  std::cout << *data << std::endl;
}};
```

- [ ] test item

### DIFF
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -26,6 +26,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <functional>
+#include <type_traits>
 
 namespace pros {
 class Task {
@@ -102,11 +103,14 @@ class Task {
 	template <class F>
 	Task(F&& function, std::uint32_t prio = TASK_PRIORITY_DEFAULT,
 	     std::uint16_t stack_depth = TASK_STACK_DEPTH_DEFAULT, const char* name = "")
-	     : Task(&[] (void* parameters) {
-		auto ptr = std::static_cast<std::function<void()>*>(parameters);
+	     : Task([] (void* parameters) {
+		auto ptr = static_cast<std::function<void()>*>(parameters);
 		(*ptr)();
 		delete ptr;
-	     }, new std::function<void()>(std::forward<F>(function))) {}
+	     }, new std::function<void()>(std::forward<F>(function)), prio, stack_depth, name) {
+		static_assert(std::is_invocable_v<F>);
+	     }
+	}
 
 	/**
 	 * Creates a new task and add it to the list of tasks that are ready to run.
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -26,6 +26,7 @@
 #include <cstdint>
 #include <cstdlib>
 #include <functional>
+#include <memory>
 #include <type_traits>
 
 namespace pros {
@@ -104,9 +105,8 @@ class Task {
 	Task(F&& function, std::uint32_t prio = TASK_PRIORITY_DEFAULT,
 	     std::uint16_t stack_depth = TASK_STACK_DEPTH_DEFAULT, const char* name = "")
 	     : Task([] (void* parameters) {
-		auto ptr = static_cast<std::function<void()>*>(parameters);
+		std::unique_ptr<std::function<void()>> ptr{static_cast<std::function<void()>*>(parameters)};
 		(*ptr)();
-		delete ptr;
 	     }, new std::function<void()>(std::forward<F>(function)), prio, stack_depth, name) {
 		static_assert(std::is_invocable_r_v<void, F>);
 	}
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -108,7 +108,7 @@ class Task {
 		(*ptr)();
 		delete ptr;
 	     }, new std::function<void()>(std::forward<F>(function)), prio, stack_depth, name) {
-		static_assert(std::is_invocable_v<F>);
+		static_assert(std::is_invocable_r_v<void, F>);
 	     }
 	}
 
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -128,8 +128,8 @@ class Task {
 	 *
 	 */
 	template <class F>
-	Task(F&& function, const char* name) : Task(std::forward<F>(function),
-	     TASK_PRIORITY_DEFAULT, TASK_STACK_DEPTH_DEFAULT, name) {}
+	Task(F&& function, const char* name)
+	    : Task(std::forward<F>(function), TASK_PRIORITY_DEFAULT, TASK_STACK_DEPTH_DEFAULT, name) {}
 
 	/**
 	 * Create a C++ task object from a task handle
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -25,6 +25,7 @@
 #undef delay
 #include <cstdint>
 #include <cstdlib>
+#include <functional>
 
 namespace pros {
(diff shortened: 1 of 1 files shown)


include/pros/rtos.hpp
```diff
@@ -78,6 +79,53 @@ class Task {
...
+	 *
+	 */
+	template <class F>
+	Task(F&& function, std::uint32_t prio = TASK_PRIORITY_DEFAULT,
+	     std::uint16_t stack_depth = TASK_STACK_DEPTH_DEFAULT, const char* name = "")
+	     : Task(&[] (void* parameters) {
+		auto ptr = std::static_cast<std::function<void()>*>(parameters);
+		(*ptr)();
+		delete ptr;
+	     }, new std::function<void()>(std::forward<F>(function))) {}
```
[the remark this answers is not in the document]
HotelCalifornia: you can probably specify a parameter pack above in the constructor and pass them on down here
e.g.
```cpp
template<class F, class ... Args>
Task(F&& function, Args... args) : Task(<lambda>, <forward_func>, &args...) {};
```
so that we can avoid introducingg [more] ambiguous constructors and also only have to write one constructor for this.

include/pros/rtos.hpp
```diff
@@ -78,6 +80,56 @@ class Task {
...
+	 */
+	template <class F>
+	Task(F&& function, std::uint32_t prio = TASK_PRIORITY_DEFAULT,
+	     std::uint16_t stack_depth = TASK_STACK_DEPTH_DEFAULT, const char* name = "")
+	     : Task([] (void* parameters) {
+		auto ptr = static_cast<std::function<void()>*>(parameters);
+		(*ptr)();
+		delete ptr;
+	     }, new std::function<void()>(std::forward<F>(function)), prio, stack_depth, name) {
+		static_assert(std::is_invocable_v<F>);
```
HotelCalifornia: should probably be explicit here that F can also only return void:
```suggestion
		static_assert(std::is_invocable_r_v<void, F>);
```

HotelCalifornia:
looks good to me, just gotta compile and test

HotelCalifornia:
> 1 failed
ah well, thought I'd give it a shot anyway.

Octogonapus:
It failed because it failed installing the cli:
`ERROR: No matching distribution found for pros-cli-v5`

HotelCalifornia:
Yeah, it failing was expected. See purduesigbots/pros-docs#156

baylessj:
Working on getting my environment set up again so I can actually get some use out of the v5 sitting on my desk. I'll report back in a bit on how this works.

include/pros/rtos.hpp
```diff
@@ -78,6 +80,56 @@ class Task {
...
+	Task(F&& function, std::uint32_t prio = TASK_PRIORITY_DEFAULT,
+	     std::uint16_t stack_depth = TASK_STACK_DEPTH_DEFAULT, const char* name = "")
+	     : Task([] (void* parameters) {
+		auto ptr = static_cast<std::function<void()>*>(parameters);
+		(*ptr)();
+		delete ptr;
+	     }, new std::function<void()>(std::forward<F>(function)), prio, stack_depth, name) {
+		static_assert(std::is_invocable_r_v<void, F>);
+	     }
+	}
```
baylessj: Got a compilation error here because this "}" character is extraneous and matches with the opening brace of the class definition. Needs removed to compile properly

baylessj:
Additional note - got compilation errors with the given example for a couple of reasons.

I ended up with the following working code:
```
std::shared_ptr<int> data{new int(7)};
pros::Task task([=] {
  pros::delay(1000);
  std::cout << *data << std::endl;
}, "Test");
```
Got the following error about the `unique_ptr`:
```
src/main.cpp: In function 'void initialize()':
src/main.cpp:33:17: error: use of deleted function 'std::unique_ptr<_Tp, _Dp>::unique_ptr(const std::unique_ptr<_Tp, _Dp>&) [with _Tp = int; _Dp = std::default_delete<int>]'
 pros::Task task([=] {
                 ^
In file included from /home/bayle/gcc-arm-none-eabi-8-2019-q3-update/arm-none-eabi/include/c++/8.3.1/memory:80,
                 from src/main.cpp:2:
/home/bayle/gcc-arm-none-eabi-8-2019-q3-update/arm-none-eabi/include/c++/8.3.1/bits/unique_ptr.h:394:7: note: declared here
       unique_ptr(const unique_ptr&) = delete;
       ^~~~~~~~~~
make: *** [common.mk:255: bin/main.cpp.o] Error 1
```

an error about the lack of a task name, and then the following error about using a brace initializer:
```
src/main.cpp:36:2: error: no matching function for call to 'pros::Task::Task(<brace-enclosed initializer list>)'
 }};
  ^
```
The above issues are a bit beyond my knowledge of modern C++, so I can't offer a whole lot in the way of a solution besides providing the errors and the working test code that I ran.

Octogonapus:
Not sure if we can provide a default configuration that will work for all these cases; `unique_ptr` complicates things a bit.

HotelCalifornia:
The example was ill-formed as far as I know.

```cpp
std::unique_ptr<int> data{new int(7)};
pros::Task task([data{move(data)}] (void*) {
  pros::delay(1000);
  std::cout << *data << std::endl;
});
```
should do the trick. Main problem was that the `=` capture specifier doesn't move things by default, so it was complaining that it couldn't copy the unique pointer. Anyway changing to a shared pointer is also a valid solution as you found.

Not sure what's going on with the initializer list confusion except that maybe leaving off the argument list for the lambda caused it to get confused? Since tasks should specifically have the signature `void (*)(void*)` idk why we're allowing `std::function<void()>` here tbh. Or we could extend the is_invocable check to make sure we're only getting either `void()` or `void(void*)`.

Octogonapus:
Can you explain what `data{move(data)}` does?

HotelCalifornia:
[generalized lambda capture](https://isocpp.org/wiki/faq/cpp14-language#lambda-captures), added in c++14

nickmertin:
It would probably be most sensible to change the test case to use `shared_ptr`, and keep the default copy-assign capture.

nickmertin:
Verified that the latest version builds with the following main.cpp:

```
#include "main.h"
#include <iostream>
#include <memory>

/**
 * Runs the operator control code. This function will be started in its own task
 * with the default priority and stack size whenever the robot is enabled via
 * the Field Management System or the VEX Competition Switch in the operator
 * control mode.
 *
 * If no competition control is connected, this function will run immediately
 * following initialize().
 *
 * If the robot is disabled or communications is lost, the
 * operator control task will be stopped. Re-enabling the robot will restart the
 * task, not resume it from where it left off.
 */
void opcontrol() {
        std::shared_ptr<int> data{new int(7)};
        pros::Task task{[=] {
                pros::delay(1000);
                std::cout << *data << std::endl;
        }};
}
```

I do not currently have access to a V5 to test.

nickmertin:
Point of explanation on removing the default value of `name` on the second added constructor: it is never necessary, because a single-argument call will always default to the first one.

nickmertin:
Verified sample code working on a V5

HotelCalifornia:
@nickmertin would you mind running this through clang-format? There's something hinky going on with the braces and indentation

Issue #90, opened 2019-01-31
Make LVGL thread-safe


HotelCalifornia:
#### Expected Behavior:

LVGL functions should be thread-safe.

#### Actual Behavior:

Per littlevgl/lvgl#544, we need to mutex guard the call to `lv_task_handler`, and then make the same mutex available to guard the calls to the `lv_*` API.

#### Steps to reproduce:

This hasn't yet been the obvious cause of any observed issues.

#### System information:
Platform: V5 
PROS Kernel Version: any 

#### Additional Information

As I'm thinking about it, it seems the best solution will be to add accessor functions:

```c
lv_mutex_take(void);
lv_mutex_give(void);
```

My first thought was to just have a single function that would just return the mutex itself, but I have since realized that would potentially allow users to delete it, which would not be desirable. As such, I think the above solution is the way to go.

#### Screenshots/Output Dumps/Stack Traces
N/A

HotelCalifornia:
@edjubuh said:
> A recursive mutex would be fine here.
>
> A PR to LVGL might not be a terrible idea, allowing for
>
>
```c
#define LV_LOCK_TAKE (whatever)
#define LV_LOCK_GIVE (whatever)
```
> and take/give at every entry/exit point of every function.

lights0123:
Theoretically, shouldn't this only occur if someone yields (such as using `delay`) to FreeRTOS within an LVGL call? As far as I know, the FreeRTOS configuration used in PROS only context switches with explicit yields, not based on a timer. So the only times thread safety would be an issue would be:

- Someone [extends] an LVGL object and causes a FreeRTOS yield such as using `delay` or `taskYIELD` directly in between LVGL modifications
- Someone (and this is more likely) `delay`s within an LVGL callback

[extends]: https://blog.littlevgl.com/2018-12-13/extend-lvgl-objects

HotelCalifornia:
> only context switches with explicit yields, not based on a timer.

The configuration we use includes priority-based preemptive scheduling, in which at any given time the currently running task is the highest priority task that isn't blocked. When there are multiple tasks running at the same priority, they are round-robined such that each gets an equal amount of processing time.

As I mentioned in the original comment, this has yet to be the obvious cause of any issues, which means in the end you may be right. My original intent though was just to be safer than sorry (especially as the porting guide now recommends this course of action).

lights0123:
[...]

PR #180, opened 2019-10-23
🐛 Fix race and an abandoned mutex

nathan-moore:
#### Summary:
Within lv_task, we were using a non atomic swap as a mutex. This changes these to be volatile/atomic where necessary. I also noticed a path where said "mutex" could be abandoned, so that's fixed as well.

#### Motivation:
Non deterministic races aren't fun.

##### References (optional):
Pointed out in https://github.com/purduesigbots/pros/issues/90#issuecomment-545628430

#### Test Plan:
- [x] lvgl still works.

### DIFF
--- a/src/display/lv_misc/lv_task.c
+++ b/src/display/lv_misc/lv_task.c
@@ -33,7 +33,7 @@ static bool lv_task_exec(lv_task_t * lv_task_p);
 /**********************
  *  STATIC VARIABLES
  **********************/
-static bool lv_task_run = false;
+static volatile bool lv_task_run = false;
 static uint8_t idle_last = 0;
 static bool task_deleted;
 static bool task_created;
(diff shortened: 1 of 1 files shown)


src/display/lv_misc/lv_task.c
```diff
@@ -66,14 +66,21 @@ LV_ATTRIBUTE_TASK_HANDLER void lv_task_handler(void)
...
+    if (taken) return;
 
     static uint32_t idle_period_start = 0;
     static uint32_t handler_start = 0;
     static uint32_t busy_time = 0;
 
-    if(lv_task_run == false) return;
+    if(lv_task_run == false)
+    {
+        goto cleanup;
```
Octogonapus: Is there a way to structure this code without `goto`?

HotelCalifornia:
looks good pending testing. the PR build actually finished fine, but it was unable to make the template due to some weird versioning thing

nathan-moore:
@HotelCalifornia Seems to work fine.

PR #182, opened 2019-11-06
Remove ambiguous case for constructing pros::Task

SpencerJ21:
#### Summary:
There is an ambiguous case for construction of pros::Task when only 1 or 2 parameters are specified. This is fixed by removing the defaults on the 2nd and 3rd params.

#### Motivation:
As is, task construction when only the function and/or task parameter is specified is not allowed.

##### References (optional):
I accidentally introduced this in #158

#### Test Plan:
Make sure a Task object can be created while only specifying the function. Example:

```cpp
void func(void* param){
    std::cout << "This task is running\n";
}

pros::Task task(func);
```

- [ ] test item


HotelCalifornia:
I wonder how this interacts with #177

SpencerJ21:
Looks like #177 doesn't interact with the shortened constructor at all.

PR #187, opened 2019-12-18
✨ Add support for VEX Inertial Sensor

HotelCalifornia:
#### Summary:

Does what it says on the tin-- adds support for the new VEX Inertial Sensor (IMU).

#### Motivation:

Got at least one person a day asking me where this feature was since VEX announced (and started shipping) the new sensor last week.

##### References (optional):

Surprisingly none.

#### Test Plan:

- TBD

### DIFF
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -369,6 +369,7 @@ void queue_reset(queue_t queue);
 typedef enum v5_device_e {
 	E_DEVICE_NONE = 0,
 	E_DEVICE_MOTOR = 2,
+  E_DEVICE_IMU = 6,
 	E_DEVICE_RADIO = 8,
 	E_DEVICE_VISION = 11,
 	E_DEVICE_ADI = 12,
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -369,7 +369,7 @@ void queue_reset(queue_t queue);
 typedef enum v5_device_e {
 	E_DEVICE_NONE = 0,
 	E_DEVICE_MOTOR = 2,
-  E_DEVICE_IMU = 6,
+	E_DEVICE_IMU = 6,
 	E_DEVICE_RADIO = 8,
 	E_DEVICE_VISION = 11,
 	E_DEVICE_ADI = 12,
--- a/include/pros/imu.h
+++ b/include/pros/imu.h
@@ -41,7 +41,7 @@ typedef struct __attribute__((__packed__)) quaternion_s {
 	double w;
 } quaternion_s_t;
 
-struct __attribute__((__packed__)) imu_raw_s {
+struct imu_raw_s {
 	double x;
 	double y;
 	double z;
@@ -95,25 +95,6 @@ int32_t imu_reset(uint8_t port);
 double imu_get_rotation(uint8_t port);
 
 /**
- * Get the total number of degrees the Inertial Sensor has spun about the z-axis
- *
- * This function is an alias for imu_get_rotation. It is preserved for use by
- * those familiar with the vexOS API.
- *
- * This function uses the following values of errno when an error state is
- * reached:
- * ENXIO - The given value is not within the range of V5 ports (1-21).
- * ENODEV - The port cannot be configured as an Inertial Sensor
- * EAGAIN - The sensor is still calibrating
- *
- * \param  port
- * 				 The V5 Inertial Sensor port number from 1-21
- * \return The degree value or PROS_ERR_F if the operation failed, setting
- * errno.
- */
-double imu_get_vex_heading(uint8_t port);
-
-/**
  * Get the Inertial Sensor's heading relative to the initial direction of its
  * x-axis
  *
@@ -133,26 +114,6 @@ double imu_get_vex_heading(uint8_t port);
  * errno.
  */
 double imu_get_heading(uint8_t port);
-
-/**
- * Get the Inertial Sensor's heading relative to the initial direction of its
- * x-axis
- *
- * This function is an alias for imu_get_heading. It is preserved for use by
- * those familiar with the vexOS API.
- *
- * This function uses the following values of errno when an error state is
- * reached:
- * ENXIO - The given value is not within the range of V5 ports (1-21).
- * ENODEV - The port cannot be configured as an Inertial Sensor
- * EAGAIN - The sensor is still calibrating
- *
- * \param  port
- * 				 The V5 Inertial Sensor port number from 1-21
- * \return The degree value or PROS_ERR_F if the operation failed, setting
- * errno.
- */
-double imu_get_vex_degrees(uint8_t port);
 
 /**
  * Get a quaternion representing the Inertial Sensor's orientation
--- a/include/pros/imu.h
+++ b/include/pros/imu.h
@@ -41,7 +41,7 @@ typedef struct __attribute__((__packed__)) quaternion_s {
 	double w;
 } quaternion_s_t;
 
-struct imu_raw_s {
+struct __attribute__((__packed__)) imu_raw_s {
 	double x;
 	double y;
 	double z;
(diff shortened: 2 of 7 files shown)


HotelCalifornia:
Some notes:
- If the IMU is still calibrating when a function is called, the function will return an error value and `errno` will be set to `EAGAIN` 
  - this functionality could be put up for debate, just seemed like a good idea when I was thinking about it
- it takes about 2,000ms to calibrate, as was mentioned on the forums (I'm consistently getting 2,020ms when testing)

Here's the template built by azure in case anyone wants to test it:
[kernel@3.2.1-pr187.857.zip](https://github.com/purduesigbots/pros/files/3976462/kernel%403.2.1-pr187.857.zip)

HotelCalifornia:
my test code looks like this
```cpp
void opcontrol() {
	pros::Imu imu(2);
	imu.reset();
	std::size_t iter = 0;
	auto time = pros::millis();
	while (imu.is_calibrating()) {
		pros::lcd::print(0, "IMU is calibrating (%u)", iter);
		pros::lcd::print(1, "status: %d", imu.get_status());
		pros::delay(10);
		iter += 10;
	}
	pros::lcd::print(0, "IMU done calibrating (took %u milliseconds)", pros::millis() - time);

	while (true) {
		pros::lcd::print(1, "heading: %05.2f", imu.get_heading());
		pros::lcd::print(2, "degrees: %05.2f deg", imu.get_degrees());
		auto quaternion = imu.get_quaternion();
		pros::lcd::print(3, "%03.2f+(%03.2f)i+(%03.2f)j+(%03.2f)k", quaternion.a, quaternion.b, quaternion.c, quaternion.d);
		auto attitude = imu.get_attitude();
		pros::lcd::print(4, "att:{p:%03.2f,r:%03.2f,y:%03.2f}", attitude.pitch, attitude.roll, attitude.yaw);
		auto gyro = imu.get_raw_gyro();
		pros::lcd::print(5, "gyro:{%03.2f,%03.2f,%03.2f,%03.2f}", gyro.x, gyro.y, gyro.z, gyro.w);
		auto accel = imu.get_raw_gyro();
		pros::lcd::print(6, "accel:{%03.2f,%03.2f,%03.2f,%03.2f}", accel.x, accel.y, accel.z, accel.w);
		pros::delay(20);
	}
}
```

include/pros/imu.h
```diff
@@ -0,0 +1,86 @@
...
+	E_IMU_STATUS_ERROR = 0xFF,  // NOTE: used for returning an error from the get_status function, not that the IMU is
+	                            // necessarily in an error state
+} imu_status_e_t;
+
+typedef struct __attribute__((__packed__)) quaternion_s {
+	double a;
+	double b;
+	double c;
+	double d;
+} quaternion_s_t;
```
Octogonapus: Quaternions are comprised of a vector component and a scalar component, usually denoted by `x, y, z` and `w`. Should this be renamed, or does vexOS do something weird with the data?

HotelCalifornia: wikipedia told me
> _a_ + _b_**i** + _c_**j** + _d_**k**

so i just went with that, but i'm willing to change it (back) to w/x/y/z if that's more correct

nathan-moore: Is there a difference between this and imu_s_t besides the layout? We could combine them on the pros API side.

HotelCalifornia: the difference is semantic

HotelCalifornia: .... which I suppose could just be handled with a typedef

BillWyeman:
As per the conversation on discord, the name `Imu`could be confusing, and in my opinion doesn't look good. Possible alternatives include `Inertial` or `IMU`, but this is all up for debate.

include/pros/imu.h
baylessj: I'm more familiar with w,x,y,z for quaternions, and I would advocate for that naming only because the (x,y,z) component is a bit more obvious in that form than the a,b,c,d format. I don't think it's a big deal which version we go with, but that's my $0.02

HotelCalifornia:
This latest commit does a couple things:
- updates the quaternion type to be `{x,y,z,w}` (this could still be changed to `{w,x,y,z}` if that's what people are more used to. I guess since it's a struct it doesn't really matter)
- provides two typedefs (gyroscope and accelerometer) for the "raw" IMU type, which has been updated to no longer carry the `w` component that vexOS gives it
- changes `imu_get_attitude` to `imu_get_euler` so that it more closely matches `imu_get_quaternion`
- renames `imu_get_raw_{gyro|accel}` to `imu_get_gyro_rate` and `imu_get_accel`

thanks for the suggestions from @Octogonapus (here) and @tehkhop (on discord)

HotelCalifornia:
build as of ebc04a9:
[kernel@3.2.1-pr187.870.zip](https://github.com/purduesigbots/pros/files/3980692/kernel%403.2.1-pr187.870.zip)

HotelCalifornia:
still looking for feedback:
- `imu_get_degrees` and `imu_get_heading` are ambiguously named:
  - the former returns a value bounded by [-360, 360] that represents degrees about the z-axis (where zero is along the x-axis)
  - the latter returns a value that is not bounded, aside from being a relative number based on the same zero as above (i.e. it accumulates degrees as you turn)
- is it worth adding getters like `imu_get_{pitch|roll|yaw}`?
- anything else I'm missing?

kunwarsahni01:
Add `getRemapped` which would get the current sensor value remapped into the target range ([1800, -1800] by default). Exactly like the ADI Gyro.

HotelCalifornia:
> Add `getRemapped` which would get the current sensor value remapped into the target range ([1800, -1800] by default). Exactly like the ADI Gyro.

so which sensor value would this be? the gyro's raw yaw rate?

HotelCalifornia:
jk, I just realized that's an okapilib convenience function so i'll leave that up to @Octogonapus if they want to implement that on top of our API

HotelCalifornia:
d42738d adds convenience getters for pitch, yaw, and roll as some have suggested might be of use.

in 78e5edd..fa28304:
- rename `imu_get_degrees` -> `imu_get_vex_degrees`
- rename `imu_get_heading` -> `imu_get_vex_heading`
- add `imu_get_heading` which calls `imu_get_vex_degrees`
- add `imu_get_rotation` which calls `imu_get_vex_heading`
- document the C API

I tried to pick names I think will make the most people happy. `imu_get_heading` (now) really does return the (-360,360) degrees-bounded heading relative to the initial direction of the sensor's x-axis. As far as I could tell, aeronautical navigation seems to only ever use the angle between nose/bow direction and true north, but if we just concede that "true north" is wherever the sensor was pointing at calibration time then there's no problem :sweat_smile: 

`imu_get_rotation` seemed like a nice, concise alternative to `imu_get_absolute_degrees` as someone suggested to me in slack that still carries the right semantic meaning (I think... I hope).

As always, I'd love to hear community feedback and suggestions before we get this merged

HotelCalifornia:
here's the latest template 
[kernel@3.2.1-commit.57.fa28304.zip](https://github.com/purduesigbots/pros/files/4004578/kernel%403.2.1-commit.57.fa28304.zip)

include/pros/imu.h
```diff
@@ -0,0 +1,295 @@
...
+} imu_status_e_t;
+
+typedef struct __attribute__((__packed__)) quaternion_s {
+	double x;
+	double y;
+	double z;
+	double w;
+} quaternion_s_t;
+
+struct __attribute__((__packed__)) imu_raw_s {
```
nathan-moore: The layout would probably be best left to the compiler, since it's not used for interop with vexos.

include/pros/imu.h
```diff
@@ -0,0 +1,295 @@
...
+ * ENXIO - The given value is not within the range of V5 ports (1-21).
+ * ENODEV - The port cannot be configured as an Inertial Sensor
+ * EAGAIN - The sensor is still calibrating
+ *
+ * \param  port
+ * 				 The V5 Inertial Sensor port number from 1-21
+ * \return The degree value or PROS_ERR_F if the operation failed, setting
+ * errno.
+ */
+double imu_get_rotation(uint8_t port);
```
nathan-moore: imu_get_rotations sounds better to me, since it's unbounded.

include/pros/imu.h
```diff
@@ -0,0 +1,295 @@
...
+ * ENXIO - The given value is not within the range of V5 ports (1-21).
+ * ENODEV - The port cannot be configured as an Inertial Sensor
+ * EAGAIN - The sensor is still calibrating
+ *
+ * \param  port
+ * 				 The V5 Inertial Sensor port number from 1-21
+ * \return The degree value or PROS_ERR_F if the operation failed, setting
+ * errno.
+ */
+double imu_get_heading(uint8_t port);
```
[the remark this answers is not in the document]
HotelCalifornia: We're going to have to accept some level of confusion, whether that's not maintaining any compatibility with the API RMS and VexCode expose or whether that's doing what I've done here and providing aliased functions that match the other APIs.

That said I'm not sure that autocompletion will be a problem? The `*_vex` functions should (I think) always show up after the others, and like you said people who wonder about the difference should read the docs

include/pros/imu.h
HotelCalifornia: That's fair. I tossed that around in my head a bit, and eventually figured I'd just pack it (e.g. in case someone wants to send it directly over a wire, or something). Not terribly committed either way, so I'll get rid of that

include/pros/imu.h
```diff
@@ -0,0 +1,295 @@
...
+ * ENXIO - The given value is not within the range of V5 ports (1-21).
+ * ENODEV - The port cannot be configured as an Inertial Sensor
+ * EAGAIN - The sensor is already calibrating
+ *
+ * \param port
+ *        The V5 Inertial Sensor port number from 1-21
+ * \return 1 if the operation was successful or PROS_ERR if the operation
+ * failed, setting errno.
+ */
+int32_t imu_reset(uint8_t port);
```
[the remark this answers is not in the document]
tehkhop: Does it reset all the integrated angles to 0 as well? If so, calibrate/recalibrate might be a misleading name.

include/pros/imu.h
tehkhop: "Rotations" implies a quantity of full revolutions to me. I'm still not a fan of "rotation" either as it doesn't specify an axis, but I don't have anything better to offer.

include/pros/imu.h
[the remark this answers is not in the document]
tehkhop: Maybe the _vex functions could live in the extended API header? I'd presume it to be incredibly rare for someone to want to use them, but they can still be there for completeness.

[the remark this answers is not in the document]
HotelCalifornia: I was thinking of that too, but I'm not sure if that jives with my assumed goal of easing the transition from vexcode/RMS to PROS...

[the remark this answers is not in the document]
HotelCalifornia: I say "assumed" because that was my thought process, rather than leaving them in for advanced users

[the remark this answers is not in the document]
tehkhop: Right. As far as I'm aware though, vexcode only exposes the imu through it's C++ API, which looks rather different to the sdk api. Also, I believe the method they use for yaw in the vexcode example code is actually called rotation and behaves the same as imu_get_rotation. Not checked RMS though.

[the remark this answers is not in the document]
HotelCalifornia: Okay, I'll have to verify that then

BillWyeman:
So it seems like VexCode flip heading and rotation, so they match `imu_get_heading` and `imu_get_rotation`, so I don't see the case for `imu_get_vex_heading` and `imu_get_vex_degrees`, as they will just lead to added confusion.

include/pros/imu.h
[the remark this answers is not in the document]
HotelCalifornia: Right so I've just checked the vexcode c++ API for the IMU and it looks like they use the same heading/rotation idiom that we've got here. This means that we can safely remove the `_vex` versions and not have to confuse people who switch.

include/pros/imu.h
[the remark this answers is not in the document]
HotelCalifornia: The vexcode API seems to call this `startCalibration`.... It also accepts an `int,32_t` whose purpose seems to be undocumented

HotelCalifornia:
from b0c262e: the C++ API now matches the C one.

[kernel@3.2.1-pr187.883.zip](https://github.com/purduesigbots/pros/files/4016883/kernel%403.2.1-pr187.883.zip)

still to do:
- remove the `_vex` versions of the heading/rotation functions

include/pros/imu.h
HotelCalifornia: gonna stick with imu_get_rotation

include/pros/apix.h
```diff
@@ -369,6 +369,7 @@ void queue_reset(queue_t queue);
 typedef enum v5_device_e {
 	E_DEVICE_NONE = 0,
 	E_DEVICE_MOTOR = 2,
+  E_DEVICE_IMU = 6,
```
nathan-moore: nit: looks like you need a run of clang format

include/pros/imu.h
```diff
@@ -0,0 +1,256 @@
...
+ *
+ * Contains prototypes for functions related to the VEX Inertial sensor.
+ *
+ * Visit https://pros.cs.purdue.edu/v5/tutorials/topical/imu.html to learn
+ * more.
+ *
+ * This file should not be modified by users, since it gets replaced whenever
+ * a kernel upgrade occurs.
+ *
+ * Copyright (c) 2017-2019, Purdue University ACM SIGBots.
```
nathan-moore:
```suggestion
 * Copyright (c) 2017-2020, Purdue University ACM SIGBots.
```

include/pros/imu.hpp
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * Contains prototypes for functions related to the VEX Inertial sensor.
+ *
+ * Visit https://pros.cs.purdue.edu/v5/tutorials/topical/imu.html to learn
+ * more.
+ *
+ * This file should not be modified by users, since it gets replaced whenever
+ * a kernel upgrade occurs.
+ *
+ * Copyright (c) 2017-2019, Purdue University ACM SIGBots.
```
nathan-moore:
```suggestion
 * Copyright (c) 2017-2020, Purdue University ACM SIGBots.
```

src/devices/vdml_imu.c
```diff
@@ -0,0 +1,144 @@
...
+ */
+
+#include <errno.h>
+#include "pros/imu.h"
+#include "v5_api.h"
+#include "vdml/registry.h"
+#include "vdml/vdml.h"
+
+#define ERROR_IMU_STILL_CALIBRATING(port, device, err_return)                  \
+	if (vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING) { \
```
nathan-moore: Would be nice to add an assert here that we're under the mutex, but I remember looking earlier and not seeing a way to do it.

include/pros/apix.h
HotelCalifornia: ah, it didn't catch this one because it wasn't tracked

nathan-moore: A couple of imu.hpp's comments are also messed up.

PR #188, opened 2019-12-20
✨ Add a new class method: get_port

Jerrylum:
#### Summary:

Someone(yes, he is me) thinks it is better to have a getter method. So I added `get_port` method in Motor, Vision and Serial classes. It is useful for VEXU team which has more than 16 motors in a robot and messes it up.

#### Motivation:
Love, yes, it is.

##### References (optional):
Null

#### Test Plan:
```cpp
void opcontrol() {
    pros::Motor the_motor(2);
    pros::Serial the_serial(3);
    pros::Vision the_vision(4);
    while (true) {
        pros::lcd::print(2, "The port number is: %d", the_motor.get_port());
        pros::lcd::print(3, "The port number is: %d", the_serial.get_port());
        pros::lcd::print(4, "The port number is: %d", the_vision.get_port());
        pros::delay(20);
    }
}
```

- [x] run the test code

### DIFF
--- a/include/pros/motors.hpp
+++ b/include/pros/motors.hpp
@@ -814,6 +814,13 @@ class Motor {
 	 */
 	virtual std::int32_t get_voltage_limit(void) const;
 
+	/**
+	 * Gets the port number of the motor.
+	 *
+	 * \return The motor's port number.
+	 */
+	virtual std::uint8_t get_port(void) const;
+
 	private:
 	const std::uint8_t _port;
 };
--- a/include/pros/serial.hpp
+++ b/include/pros/serial.hpp
@@ -120,6 +120,13 @@ class Serial {
 	virtual std::int32_t get_write_free() const;
 
 	/**
+	 * Gets the port number of the serial port.
+	 *
+	 * \return The serial port's port number.
+	 */
+	std::uint8_t get_port() const;
+
+	/**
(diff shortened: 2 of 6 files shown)


HotelCalifornia:
Can you elaborate a little on why you think this is necessary? I'm not sure I see immediately how it's useful even when you have a large number of motors...

Jerrylum:
It is very useful when you store motors in the array and need to loop through all of them. 

**Case**
1. Log all motors’ temperature in the motor group.
```cpp
// Declare motors
motor_group my_group (mtr_1, mtr_3, mtr_5, mtr7);

for (int i = 0; i < my_group.length(); i++) {
    pros::Motor mtr = my_group[i];
    cout <<  mtr.get_port() << ": " << mtr.get_temperature() << endl;
}
```

2. Motor connection check
```cpp
void disconnect_check(motor_group group) {
    cout << group.name() << ": ";
    for (int i = 0; i < my_group.length(); i++) {
        if (my_group[i].move_voltage(0) == PROS_ERR) {
            cout <<  mtr.get_port() << " ";
        }
    }
}

// Declare motor group
disconnect_check(group_left);
disconnect_check(group_right);
```

Like vexCode API has `.index()` method too. I am thinking about is there any issue to access the port variable? In my opinion, Although It is not necessary, it is useful and convenient since you can recognize the motor without variable name or index in the array. There is nothing to lose.

Issue #174, opened 2019-09-22
SD Card Detection Command


1069B:
#### Expected Behavior:
There is a boolean function that returns true if the system detects an SD Card and false if there is not SD Card detected.

#### Actual Behavior:
There is no current function that detects the SD Card.

#### System information:
Platform: V5
PROS Kernel Version: 3.2.0

#### Additional Information
Hotel stated that the function exists but isn't released to the public. [Vex Forum where this was disscused](https://www.vexforum.com/t/sd-card-detection-command-in-pros/66632/2)

Jerrylum:
We can now close this issue. Thank you.

PR #189, opened 2019-12-20
✨ Add SD card detection command

Jerrylum:
#### Summary:
Well, I am still thinking about is it a suitable name for namespace and method.

**Namespace**: usd? SDcard? sdcard? or without namespace? (like: `pros::is_usd_installed()`)
**Method**: is_installed? is_inserted? is_mounted? is_connected? is_valid? etc...

#### Motivation:
Add it

##### References (optional):
#174 @nathan-moore 

#### Test Plan:
```cpp
void opcontrol() {
    while (true) {
        pros::lcd::print(2, "Installed? %d", pros::usd::is_installed());
        pros::lcd::print(3, "Installed? %d", pros::c::usd_is_installed());
        pros::delay(20);
    }
}
```

- [x] run the test code

### DIFF
--- a/include/pros/misc.h
+++ b/include/pros/misc.h
@@ -424,6 +424,13 @@ double battery_get_temperature(void);
  */
 double battery_get_capacity(void);
 
+/**
+ * Checks if the SD card is installed.
+ *
+ * \return 1 if the SD card is installed, 0 otherwise
+ */
+int32_t usd_is_installed(void);
+
 #ifdef __cplusplus
 }
 }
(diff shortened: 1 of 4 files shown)


HotelCalifornia:
As far as a the naming, I think we've pretty consistently been using the prefix of a C function as a namespace or a class. In other words, I think you should make a tiny `usd` namespace:
```
namespace pros {
namespace usd {
bool is_installed() {
  return pros::c::usd_is_installed();
}
}
}
```

Alternatively you could make a class whose default constructor is deleted and only has the static member function `is_installed`

Jerrylum:
Ok, we are on the right track.

Octogonapus:
@HotelCalifornia It looks like @Jerrylum made that change, how's the PR look now?

PR #190, opened 2019-12-29
🔇 Stop warnings for GCC 7.1 ABI change

Octogonapus:
#### Summary:
This PR adds `-Wno-psabi` to `WARNFLAGS`.

#### Motivation:
Many users encounter a warning similar to this, especially when using OkapiLib:
```
 parameter passing for argument of type 'std::initializer_list<okapi::Point>' changed in GCC 7.1
```

It causes a lot of unnecessary confusion, so we should ignore it by default.

##### References (optional):
None.

#### Test Plan:
There is no test plan.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -5,7 +5,7 @@ MFLAGS=-mcpu=cortex-a9 -mfpu=neon-fp16 -mfloat-abi=softfp -Os -g
 CPPFLAGS=-D_POSIX_THREADS -D_UNIX98_THREAD_MUTEX_ATTRIBUTES
 GCCFLAGS=-ffunction-sections -fdata-sections -fdiagnostics-color -funwind-tables
 
-WARNFLAGS+=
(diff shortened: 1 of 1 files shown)


edjubuh:
Why is the warning flag emitted? Why is it safe to ignore?

"Confusion" is okay justification for suppressing the warning, but not also a good justification for ignoring.

Octogonapus:
The warning is emitted whenever you use an initializer list because of an ABI change. It's safe to ignore because PROS, OkapiLib, and anything other users create is all compiled with at least gcc 7.1 (making the ABI change irrelevant).

HotelCalifornia:
The warning flag is emitted due to a subtle ABI change that happened quite a while ago. Since our ecosystem requires gcc ≥ 7.2 (and everyone for the most part uses the same toolchain) this is never a problem for us

Octogonapus:
Bump? This change is pretty simple

PR #198, opened 2020-01-21
✨ Enable LVGL keyboard

HotelCalifornia:
#### Summary:

does what it says on the tin

#### Motivation:

someone asked for it

##### References (optional):

closes #194 

#### Test Plan:

- [x] program with lv_kb compiles and works

PR #185, opened 2019-11-22
🐛 Fix clearing controller LCD

nathan-moore:
#### Summary:
Fixes clearing both the full screen and a single line.

#### Test Plan:
- [x] VEX Atomic clear works
- [x] non atomic clear works
- [x] clearing a line works

With this working:

`master.print(0, 0, "abcdefghijklmnopq");`
`pros::delay(50);`
`master.print(1, 0, "abcdefghijklmno");`
`pros::delay(50);`
`master.print(2,0,strerror(errno));`
`pros::delay(50);`
`master.clear();`
`pros::delay(3000);`
`master.print(0, 0, "abcdefghijklmnopq");`
`pros::delay(50);`
`master.clear_line(0);`

PR #179, opened 2019-10-20
🚸 Implement exception logging to terminal

theol0403:
#### Summary:
When exceptions are thrown, PROS catches them and prints them to the screen.
However, long messages are often cut short and unreadable.
With this PR, exceptions caught by PROS will be printed to the terminal as well as to the V5 Screen.

I have also separated the screen printing commands to try and fit the most of the error message as possible.

#### Known Issues:
After testing this PR, it does not seem to be printing to the terminal.
This might be due to the order of initialization or something, idk.

PR #162, opened 2019-08-30
🐛 Fix motor PID functions bug

Jerrylum:
#### Summary:
Someone forgot to check the port number. It should 0-based numbering.

#### Motivation:
Love? Maybe...

##### References (optional):
closes #161;  #160 

#### Test Plan:
- [x] Run test code from https://github.com/purduesigbots/pros/pull/162#issuecomment-531485060

PR #200, opened 2020-01-21
🔖 Release 3.2.1

HotelCalifornia:
#### Summary:

- Makefile dependency tracking (#167, #191)
- handle errors properly in VDML functions that return floating point values (#160)
- fix bug with port indexing in motor positional PID functions (#162)
- log exceptions to terminal (#179)
- add support for extra Make scripts (#170)
- clear up pros::Task constructor ambiguities (#182)
- allow tasks to be constructed from any void callable (#177)
- provide error when controller commands are being sent too quickly (#169)
- fix LVGL race and abandoned mutex (#180)
- fix clearing controller LCD (#185)
- update copyright dates (#193)
- add missing include directives (#195)
- suppress warnings for GCC 7.1 ABI change (#190)
- add support for vex inertial sensor (#187)
- enable LVGL on-screen keyboard (#198)
- add function to detect SD card presence (#189)
- add class function to get port number on motor, serial, vision sensor classes (#188)
- fix ultrasonic documentation (#199)
- add stubs so that gcc 9.x users can use hot/cold linking (#201)

#### Motivation:

it's been long enough i think

##### References (optional):

see above

#### Test Plan:

tested in individual PRs


HotelCalifornia:
there are a lot of changes listed here that are already on master, not sure what that's about

nathan-moore:
The extra commits are likely because the last merge into master from develop was squashed, not merged. We should probably be merging these.

Issue #168, opened 2019-09-12
PROS 4: Decouple LVGL and Make LVGL Template


HotelCalifornia:
#### Motivation
Currently, LVGL is quite tightly coupled with the kernel, which makes it somewhat of a nightmare to upgrade. Further, when compiling the kernel, we have to also compile all of LVGL which takes a not insignificant amount of time.

As such, I propose (as some others have proposed before me, namely @3038922 and @theol0403) that we separate LVGL out into its own template, and include it in the `default_templates` list akin to how OkapiLib is.

#### Benefits
As mentioned in the **Motivation** section, the main benefit will be to make LVGL easier to upgrade without hassle. A side benefit of this is that we will also be able to offer multiple, potentially incompatible versions of LVGL. For example, we will be able to provide both templates for both versions 5.3 and 6.0, and users will be able to choose which they prefer to use.
Also mentioned in the **Motivation** section, kernel compile time will be reduced, and we won't have to deal with seeing LVGL warnings show up in kernel builds (e.g. on Azure).

#### Technical issues
There are a number of technical concerns surrounding this proposal. I'll go over the ones I've thought about here.

- **The touch driver and flush function use functions from libv5rts.** This ends up not being much of a problem, as we can simply forward declare the functions and types we use.
- **`disp_daemon` is a static task.** Same deal here. We can either forward declare the relevant static task function(s) or just decide to expose them to users in `pros/apix.h`.
- **`display_initialize` should be called during low-level initialization.** The only requirement for users is that all LVGL setup is completed before `initialize` or `competition_initialize` are called. As such, all we'd need to do is mark the function with `__attribute__((constructor))`.
- **What about internal kernel functions that use LVGL?** To my knowledge the only place the kernel refers to LVGL (aside from LLEMU, which will be addressed next) is for the VDML port unplugging/mismatch warning. We might want to consider removing this warning entirely, as it duplicates vexOS functionality (that was not present when the system was initially implemented).
- **What about LLEMU?** LLEMU may also be refactored out into this separate template. If we make it a default template, then new projects will still work. Advanced users who modify their default templates should know that removing this will break the default PROS kernel template code.

3038922:
I'd love to help.I will not take part in the world championships this year.I'm very free now. 
But I didn't have 
`
[submodule "firmware/libv5rts"]
	path = firmware/libv5rts
	url = git@github.com:purduesigbots/libv5rts.git
	ignore = dirty
`
[...]

PR #204, opened 2020-01-25
✨ Expose the static task function

Jerrylum:
#### Summary:
I was looking at issue #168. Expose the static task function can help us get a foot in the door.

#### Motivation:
Expose it and 90% of users don't even know what is it.

##### References (optional):
#168 

#### Test Plan:
Tested 957603f
- [x] run test code mutexes.c
- [x] run test code semaphores.c
- [x] run test code static_tast_states.c

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,6 @@ EXCLUDE_SRC_FROM_LIB+=$(foreach file, $(SRCDIR)/main,$(foreach cext,$(CEXTS),$(f
 TEMPLATE_FILES=$(ROOT)/common.mk $(FWDIR)/v5.ld $(FWDIR)/v5-common.ld $(FWDIR)/v5-hot.ld
 TEMPLATE_FILES+=$(FWDIR)/libc.a $(FWDIR)/libm.a
 TEMPLATE_FILES+= $(INCDIR)/api.h $(INCDIR)/main.h $(INCDIR)/pros/*.* $(INCDIR)/display
-TEMPLATE_FILES+= $(INCDIR)/rtos/FreeRTOS.h $(INCDIR)/rtos/FreeRTOSConfig.h $(INCDIR)/rtos/portable.h $(INCDIR)/rtos/portmacro.h $(INCDIR)/rtos/projdefs.h
 TEMPLATE_FILES+= $(SRCDIR)/main.cpp
 TEMPLATE_FILES+= $(ROOT)/template-gitignore
 
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ EXCLUDE_SRC_FROM_LIB+=$(foreach file, $(SRCDIR)/main,$(foreach cext,$(CEXTS),$(f
 TEMPLATE_FILES=$(ROOT)/common.mk $(FWDIR)/v5.ld $(FWDIR)/v5-common.ld $(FWDIR)/v5-hot.ld
 TEMPLATE_FILES+=$(FWDIR)/libc.a $(FWDIR)/libm.a
 TEMPLATE_FILES+= $(INCDIR)/api.h $(INCDIR)/main.h $(INCDIR)/pros/*.* $(INCDIR)/display
+TEMPLATE_FILES+= $(INCDIR)/rtos/FreeRTOS.h $(INCDIR)/rtos/FreeRTOSConfig.h $(INCDIR)/rtos/portable.h $(INCDIR)/rtos/portmacro.h $(INCDIR)/rtos/projdefs.h
 TEMPLATE_FILES+= $(SRCDIR)/main.cpp
 TEMPLATE_FILES+= $(ROOT)/template-gitignore
 
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -35,6 +35,9 @@
 /**                            RTOS CONFIGURATION                            **/
 /******************************************************************************/
 
+// if not included rtos/FreeRTOS.h
+#ifndef INC_FREERTOS_H
+
 struct xSTATIC_LIST_ITEM {
 	uint32_t xDummy1;
 	void *pvDummy2[ 4 ];
@@ -88,6 +91,7 @@ typedef struct xSTATIC_QUEUE {
 } static_queue_s_t;
 typedef static_queue_s_t static_sem_s_t;
 
+#endif // !INC_FREERTOS_H
 
 #ifdef __cplusplus
 #include "pros/serial.hpp"
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -29,7 +29,122 @@
 #include "display/lvgl.h"
 #pragma GCC diagnostic pop
 #include "pros/serial.h"
-#include "rtos/FreeRTOS.h"
+
+
+/******************************************************************************/
+/**                            RTOS CONFIGURATION                            **/
+/******************************************************************************/
+
+struct xSTATIC_LIST_ITEM {
+	uint32_t xDummy1;
+	void *pvDummy2[ 4 ];
+};
+typedef struct xSTATIC_LIST_ITEM StaticListItem_t;
+
+struct xSTATIC_MINI_LIST_ITEM {
+	uint32_t xDummy1;
+	void *pvDummy2[ 2 ];
+};
+typedef struct xSTATIC_MINI_LIST_ITEM StaticMiniListItem_t;
+
+typedef struct xSTATIC_LIST {
+	uint32_t uxDummy1;
+	void *pvDummy2;
+	StaticMiniListItem_t xDummy3;
+} StaticList_t;
+
+#define portUSING_MPU_WRAPPERS                  0
+#define configMAX_TASK_NAME_LEN                 ( 32 )
+#define portSTACK_GROWTH                        ( -1 )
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -38,6 +38,7 @@ EXCLUDE_SRC_FROM_LIB+=$(foreach file, $(SRCDIR)/main,$(foreach cext,$(CEXTS),$(f
 TEMPLATE_FILES=$(ROOT)/common.mk $(FWDIR)/v5.ld $(FWDIR)/v5-common.ld $(FWDIR)/v5-hot.ld
 TEMPLATE_FILES+=$(FWDIR)/libc.a $(FWDIR)/libm.a
 TEMPLATE_FILES+= $(INCDIR)/api.h $(INCDIR)/main.h $(INCDIR)/pros/*.* $(INCDIR)/display
+TEMPLATE_FILES+= $(INCDIR)/rtos/FreeRTOS.h $(INCDIR)/rtos/FreeRTOSConfig.h $(INCDIR)/rtos/portable.h $(INCDIR)/rtos/portmacro.h $(INCDIR)/rtos/projdefs.h
```
HotelCalifornia: oh, i see what you meant now. not sure I like this

include/pros/apix.h
```diff
@@ -555,4 +556,107 @@ int32_t fdctl(int file, const uint32_t action, void* const extra_arg);
 }
 #endif
 
+#ifdef __cplusplus
+namespace pros {
+extern "C" {
```
HotelCalifornia: is there a reason these can't be the same block? these functions should be in the `pros::c` namespace anyway

include/pros/apix.h
```diff
@@ -555,4 +556,107 @@ int32_t fdctl(int file, const uint32_t action, void* const extra_arg);
...
+ * \return A handle to a newly created queue, or NULL if the queue cannot be
+ * created.
+ */
+queue_t queue_create_static(uint32_t length, uint32_t item_size, uint8_t* storage_buffer,
+                            static_queue_s_t* queue_buffer);
+
+#ifdef __cplusplus
+#undef task_t
+#undef task_fn_t
+#undef mutex_t
```
HotelCalifornia: you should `#undef` the other things you `#define`d in lines 562-566 as well

Makefile
HotelCalifornia: will have to think on the solution, might be just pulling the relevant structures and types out into this file as well...

include/pros/apix.h
Jerrylum: Make sense

include/pros/apix.h
Jerrylum: Oh... I copied from kapi.h
I have the same question too.

HotelCalifornia: hm, probably an oversight in kapi.h then. it's not a huge concern for that file as (afaik) all the things that include it are C files anyway. either way if we're going to have it here it should be addressed

include/pros/apix.h
Jerrylum: Updated, and now we have a messy namespace. Look at the test code.

Makefile
Jerrylum: Trying to. I think we will make it to 1800 lines.

Jerrylum: Updated

include/pros/apix.h
```diff
@@ -30,6 +30,122 @@
...
+#define configUSE_MUTEXES                       1
+#define configUSE_APPLICATION_TASK_TAG          0
+#define configNUM_THREAD_LOCAL_STORAGE_POINTERS 2
+#define configGENERATE_RUN_TIME_STATS           1
+#define configUSE_NEWLIB_REENTRANT              1
+#define configUSE_TASK_NOTIFICATIONS            1
+#define configSUPPORT_STATIC_ALLOCATION         1
+#define configSUPPORT_DYNAMIC_ALLOCATION        1
+#define INCLUDE_xTaskAbortDelay                 1
+#define configUSE_QUEUE_SETS                    0
```
nathan-moore: I don't think these should be exposed to the end user. I think manually evaluating the macros would be best. Also if we change these in the future, it would be a binary breaking change. I don't think it's a big issue, but it's worth mentioning.

include/pros/apix.h
```diff
@@ -550,7 +666,110 @@ int32_t fdctl(int file, const uint32_t action, void* const extra_arg);
...
+#define mutex_t pros::mutex_t
+#define sem_t pros::c::sem_t
+#define queue_t pros::c::queue_t
+#endif
+
+typedef uint32_t task_stack_t;
+
+/**
+ * Creates a task using statically allocated buffers. All tasks used by the PROS
+ * system must use statically allocated buffers.
```
nathan-moore: Instead of saying PROS systems must use buffers, I'd recommend saying the advantages of static tasks instead.

include/pros/apix.h
```diff
@@ -550,7 +666,110 @@ int32_t fdctl(int file, const uint32_t action, void* const extra_arg);
...
+ * \param prio
+ *        The priority at which the task should run.
+ *        TASK_PRIO_DEFAULT plus/minus 1 or 2 is typically used.
+ * \param stack_depth
+ *        The number of words (i.e. 4 * stack_depth) available on the task's
+ *        stack. TASK_STACK_DEPTH_DEFAULT is typically sufficienct.
+ * \param name
+ *        A descriptive name for the task.  This is mainly used to facilitate
+ *        debugging. The name may be up to 32 characters long.
+ *
```
nathan-moore: You're missing docs for stack_buffer and task_buffer here.

include/pros/apix.h
nathan-moore: This applies to all of these types.

include/pros/apix.h
```diff
@@ -550,7 +666,110 @@ int32_t fdctl(int file, const uint32_t action, void* const extra_arg);
...
+#endif
+
+typedef uint32_t task_stack_t;
+
+/**
+ * Creates a task using statically allocated buffers. All tasks used by the PROS
+ * system must use statically allocated buffers.
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * ENOMEM - The stack cannot be used as the TCB was not created.
```
nathan-moore: The only errno value I see is EINVAL

include/pros/apix.h
```diff
@@ -550,7 +666,110 @@ int32_t fdctl(int file, const uint32_t action, void* const extra_arg);
...
+task_t task_create_static(task_fn_t task_code, void* const param, uint32_t priority, const size_t stack_size,
+                          const char* const name, task_stack_t* const stack_buffer, static_task_s_t* const task_buffer);
+
+/**
+ * Creates a statically allocated mutex.
+ *
+ * All FreeRTOS primitives must be created statically if they are required for
+ * operation of the kernel.
+ *
+ * \param[out] mutex_buffer
```
nathan-moore: These buffers aren't out. They're the memory used run the type, so they're in. This applies to the buffers later as well.

include/pros/apix.h
Jerrylum: In order to create static_task_s_t, I evaluated the final value for the macros here. Is that what `manually evaluating` mean or do you what to remove all config macro to the end-user?

include/pros/apix.h
Jerrylum: I see. I wonder if you could help me rewrite the comment since you are familiar with this topic.

Jerrylum:
> A couple notes on wrong headers and stuff I don't think we should expose. Your test also only covers tasks, but it really should have 100% coverage over what you're adding. You can also consider adding it/augmenting some of the tests in the test folder as there are some related.
> 
> On another note related to decoupling lvgl, this isn't strictly necessary for that. You likely could just use tasks on the heap and it would work fine. I'm not opposed to taking this route though, as there are likely uses for this that I don't know about.

Yes, I will change my test code since it is not suitable after commit 5d10690.

include/pros/apix.h
nathan-moore: Maybe something like: 
Creates a task using pre-allocated buffers. Given valid inputs, the task will be created.
That's kind of what the FreeRTOS headers say.

include/pros/apix.h
Jerrylum: Fixed at cb96b0c

include/pros/apix.h
Jerrylum: Fixed at cb96b0c

include/pros/apix.h
Jerrylum: Fixed at cb96b0c

include/pros/apix.h
Jerrylum: Okay, I removed all the config macros.

HotelCalifornia:
The more I think about this, the more conflicted I get. I know I had mentioned wanting to expose static tasks so that we could more easily decouple LVGL from the kernel, but now I don't think we really even need them---a normal task would probably work just as well in this case.

Sorry to bring this up after you've done so much work on it already...

nathan-moore:
I agree with Alex. That, and with how tightly coupled static tasks are with FreeRTOS, I don't really like exposing them. Thanks for looking into it though. Do you have an opinion otherwise Jerrylum, or should this be closed?

Jerrylum:
### 👌

PR #209, opened 2020-02-05
✨ Multiple phases of initialization

Jerrylum:
#### Summary:
I was looking at issue #168. The second step is to rework kernel initialization.

#### Motivation:
Because step 1 is skipped :see_no_evil:

##### References (optional):
#168 

#### Test Plan:
- [ ] Run default user code

#### Help Wanted:
Please help me to compile the kernel for test

### DIFF
--- a/src/display/display.c
+++ b/src/display/display.c
@@ -51,9 +51,7 @@ static bool vex_read_touch(lv_indev_data_t* data) {
 	return false;
 }
 
-// HACK: this function should be called after pros_pre_init and before
-// pros_init function.
-__attribute__((constructor(115))) static void display_init(void) {
+__attribute__((constructor(PROS_KERNEL_INIT-1))) static void display_init(void) {
 	lv_init();
 
 	lv_disp_drv_t disp_drv;
--- a/src/display/display.c
+++ b/src/display/display.c
@@ -51,7 +51,9 @@ static bool vex_read_touch(lv_indev_data_t* data) {
 	return false;
 }
 
-void display_initialize(void) {
+// HACK: this function should be called after pros_pre_init and before
+// pros_init function.
+__attribute__((constructor(115))) static void display_init(void) {
 	lv_init();
 
 	lv_disp_drv_t disp_drv;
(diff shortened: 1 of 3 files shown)


src/system/startup.c
```diff
@@ -21,31 +21,36 @@
...
+// XXX: pros_pre_init and pros_init happen inside __libc_init_array, and
+// before any globalC++ constructors are invoked. This is accomplished by
+// instructing GCC to include this function in the __init_array. These
+// arguments give the compiler instructions on the priority of the constructor,
 // from 0-~65k. The first 0-100 priorities are reserved for language
 // implementation.
-__attribute__((constructor(101))) static void pros_init(void) {
+
+#define PROS_KERNEL_PRE_INIT 110
+#define PROS_KERNEL_INIT     120
```
edjubuh: Move these to common kernel header or apix.h

src/display/display.c
```diff
@@ -51,7 +51,9 @@ static bool vex_read_touch(lv_indev_data_t* data) {
 	return false;
 }
 
-void display_initialize(void) {
+// HACK: this function should be called after pros_pre_init and before
+// pros_init function.
+__attribute__((constructor(115))) static void display_init(void) {
```
edjubuh:
```suggestion
__attribute__((constructor(PROS_KERNEL_PRE_INIT+1))) static void display_init(void) {
```
or
```suggestion
__attribute__((constructor(PROS_KERNEL_INIT-1))) static void display_init(void) {
```

src/system/startup.c
```diff
@@ -21,31 +21,33 @@
...
 	vdml_initialize();
+}
 
-	display_initialize();
+// HACK: External templates can hook into initialization process by setting
+// their priority between PROS_KERNEL_PRE_INIT and PROS_KERNEL_INIT. For
+// exmaple the priority of display initialization is 119. All other initialize
+// functions should be called before pros_init function. For an exmaple of
+// what could happen if this is not the case, see
+// https://github.com/purduesigbots/pros/pull/144/#issuecomment-496901942
```
nathan-moore: I don't think this comment it right, or at least the reasoning behind this specific case. There was no user code being run between the different parts of initialization (as the scheduler wasn't started yet), so the only way the ordering would matter is if it's relying on the order of tasks being started.

WillXuCodes:
PR Reverted Due to this change making the kernel project unusable when uploaded to a brain.

WillXuCodes:
Note: This PR had to be reverted because it would cause a stop the kernel project from working on v5 brains when compiled like a normal pros project. Needs fixing.

PR #211, opened 2020-02-16
Pros lvgl sim

3038922:
#### Summary:

Modified the macro definition of pros / adi.hpp
#### Motivation:

The macro definition name was modified because it conflicts with MSVC macro definition name.
I use CCLS language server and ignore caching.
##### References (optional):

none
#### Test Plan:

none
- [ ] test item
pros_lvgl_sim

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ compile_commands.json
 temp.log
 temp.errors
 .d/
+ccls-cache/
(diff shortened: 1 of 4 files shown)


HotelCalifornia:
Um, since we don't target MSVC ever this is unnecessary.

Personally, I'm against changing this part of the API.

nathan-moore:
This is a breaking change.

I don't see a reason why we should take this break.

PR #212, opened 2020-02-17
Fix Imu.hpp imu.cpp

3038922:
#### Summary:

1.Fixed a comment error
2.add get_port();
#### Motivation:

none
##### References (optional):

none
#### Test Plan:

- [ ] test item
ps:
1. I want to add `get_port()` to `adi.hpp`.encoder i will use `const std::part<int ,int>port`
2. I want to add `std:: string_name` ,`std::string get_name() `,` Device::Device( int port,std::string name ="")` to all devices.
Can you accept such a change?

### DIFF
--- a/include/pros/imu.hpp
+++ b/include/pros/imu.hpp
@@ -19,6 +19,7 @@
 #define _PROS_IMU_HPP_
 
 #include <cstdint>
+
 #include "pros/imu.h"
 
 namespace pros {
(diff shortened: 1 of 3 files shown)


include/pros/imu.hpp
```diff
@@ -19,15 +19,32 @@
...
+	 * This function uses the following values of errno when an error state is
+	 * reached:
+	 * ENXIO - The given value is not within the range of V5 ports (1-21).
+	 * ENODEV - The port cannot be configured as an Inertial Sensor
+	 * EAGAIN - The sensor is already calibrating
+	 *
+	 * \param port
+	 *        The V5 Inertial Sensor port number from 1-21
+	 * \return 1 if the operation was successful or PROS_ERR if the operation
+	 * failed, setting errno.
```
nathan-moore: We shouldn't have return here. In other c++ constructors, we just omit it.

include/pros/imu.hpp
```diff
@@ -19,15 +19,32 @@
...
+	/**
+	 * Calibrate IMU
+	 *
+	 * This takes approximately 2 seconds, and is a non-blocking operation.
+	 *
+	 * This function uses the following values of errno when an error state is
+	 * reached:
+	 * ENXIO - The given value is not within the range of V5 ports (1-21).
+	 * ENODEV - The port cannot be configured as an Inertial Sensor
+	 * EAGAIN - The sensor is already calibrating
```
nathan-moore: The constructor doesn't set any of these, nor does it start a new calibration.

nathan-moore:
About the api additions you asked about, It's probably best to open up an issue about it instead of doing it in a pull request.

PR #216, opened 2020-03-16
Added an std::string overload of Controller::set_text()

djava:
#### Summary:
Basically what the title says.... it's pretty simple.  I also included std::string into the file because I wasn't really sure if it was included elsewhere but I figured I might as well be safe.

#### Motivation:
It was annoying me that I had to add .c_str() every time I called set_text();
Also hotel said it was ok :)

#### Test Plan:
uh you can call it and see if it works I guess.  I don't foresee there being any major issues with this change.

- [ ] test item

### DIFF
--- a/src/devices/controller.cpp
+++ b/src/devices/controller.cpp
@@ -13,6 +13,7 @@
  */
 
 #include "kapi.h"
+#include <string>
 
 namespace pros {
 using namespace pros::c;
@@ -45,6 +46,10 @@ std::int32_t Controller::get_digital_new_press(pros::controller_digital_e_t butt
 
 std::int32_t Controller::set_text(std::uint8_t line, std::uint8_t col, const char* str) {
 	return controller_set_text(_id, line, col, str);
+}
+
+std::int32_t Controller::set_text(std::uint8_t line, std::uint8_t col, std::string str) {
+	return controller_set_text(_id, line, col, str.c_str());
 }
 
 std::int32_t Controller::clear_line(std::uint8_t line) {
--- a/src/devices/controller.cpp
+++ b/src/devices/controller.cpp
@@ -47,7 +47,7 @@ std::int32_t Controller::set_text(std::uint8_t line, std::uint8_t col, const cha
 	return controller_set_text(_id, line, col, str);
 }
 
-std::int32_t Controller::set_text(std::uint8_t line, std::uint8_t col, std::string str) {
+std::int32_t Controller::set_text(std::uint8_t line, std::uint8_t col, const std::string& str) {
 	return controller_set_text(_id, line, col, str.c_str());
 }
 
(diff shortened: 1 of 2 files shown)


nathan-moore:
You need to declare the function in a header.

djava:
oopsie, my bad

Fixed

src/devices/controller.cpp
```diff
@@ -47,6 +47,10 @@ std::int32_t Controller::set_text(std::uint8_t line, std::uint8_t col, const cha
 	return controller_set_text(_id, line, col, str);
 }
 
+std::int32_t Controller::set_text(std::uint8_t line, std::uint8_t col, std::string str) {
```
nathan-moore: should this be a const std::string& ? It'll save a copy.

djava:
As far as I know (not an expert) , it shouldn't make any difference if we're not actually doing anything with the string at this level of the call stack.

Making it string& would make it a reference, which seems misleading to me since nothing modifies str, and that's usually what I would view the arg being a reference as implying.

Const makes sense I guess because it's replacing a const char *.

Based off of a stack overflow answer I just read on this topic, it seems like since (presumably) somewhere along the call stack of this function, the string gets saved into another variable, it shouldn't make any performance different because of c++11 std::move semantics that I don't really understand. 

I feel like either way is probably fine, it won't make any appreciable performance difference for this application anyway.

HotelCalifornia:
> Making it string& would make it a reference, which seems misleading to me since nothing modifies str, and that's usually what I would view the arg being a reference as implying.

Passing by reference avoids copying the data, like Nathan said. Since it's not going to be modified, it can be a const reference.

You're right as far as the performance hit being inappreciable, but doing it with a const reference is idiomatic C++ code.

nickmertin:
Nathan is correct here. Move semantics don't apply because you don't move the string object at any point. The way you have it now, when calling `set_text` it will make a deep copy of the string (i.e. if it's longer than a few characters this will require a heap allocation and memcpy), then get the address of the internal C-string, then call `controller_set_text`, which itself makes *another* copy of the string from that pointer. The second copy (the one implemented internally in `controller_set_text`) is likely necessary due to some VEXos API convention, but the first one serves no purpose and would be eliminated by changing to `const std::string &`.

HotelCalifornia:
> The second copy (the one implemented internally in controller_set_text) is likely necessary due to some VEXos API convention

(iirc this was just to cap the maximum length of the string to the width of the screen and ensure a null terminator is present to mitigate any potential buffer overruns)

nathan-moore:
@kunwarsahni01 If you have a chance, can you test this one too?

djava:
Sorry I forgot to ever respond on this, I tested it compiling (which worked), but don't have a real controller to test it on

nickmertin:
I can test if you send me sample code to use

kunwarsahni01:
Was able to get text to print. image

nathan-moore:
Thanks!

PR #221, opened 2020-07-04
⬆ Upgrade vexOS to 1.0.11

HotelCalifornia:
#### Summary:

Does what it says on the tin. Also adds the IMU data refresh rate setter function.

#### Motivation:

This will allow us to support the new sensors when they come out. Also people were asking for the IMU part.

##### References (optional):

N/A

#### Test Plan:

- [x] compiles
- [x] `imu_set_data_rate` works

### DIFF
--- a/include/pros/imu.h
+++ b/include/pros/imu.h
@@ -56,6 +56,8 @@ typedef struct __attribute__((__packed__)) euler_s {
 	double yaw;
 } euler_s_t;
 
+#define IMU_MINIMUM_DATA_RATE 5
+
 /**
  * Calibrate IMU
  *
--- a/include/pros/imu.h
+++ b/include/pros/imu.h
@@ -79,7 +79,13 @@ int32_t imu_reset(uint8_t port);
  * Set the Inertial Sensor's refresh interval in milliseconds.
  *
  * The rate may be specified in increments of 5ms, and will be rounded down to
- * the nearest increment. The minimum allowable refresh rate is 5ms.
+ * the nearest increment. The minimum allowable refresh rate is 5ms. The default
+ * rate is 10ms.
+ *
+ * As values are copied into the shared memory buffer only at 10ms intervals,
+ * setting this value to less than 10ms does not mean that you can poll the
+ * sensor's values any faster. However, it will guarantee that the data is as
+ * recent as possible.
  *
  * This function uses the following values of errno when an error state is
  * reached:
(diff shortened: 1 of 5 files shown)


HotelCalifornia:
[kernel@3.2.1-pr221.1077.zip](https://github.com/purduesigbots/pros/files/4872095/kernel%403.2.1-pr221.1077.zip)

src/devices/vdml_imu.c
```diff
@@ -29,6 +29,17 @@ int32_t imu_reset(uint8_t port) {
 	return_port(port - 1, 1);
 }
 
+int32_t imu_set_data_rate(uint8_t port, uint32_t rate) {
+	claim_port_i(port - 1, E_DEVICE_IMU);
+	ERROR_IMU_STILL_CALIBRATING(port, device, PROS_ERR);
+
+	// rate is not less than 5ms, and rounded down to nearest increment of 5
+	rate -= rate == 0 ? 5 : (rate - 5) % 5;
```
nathan-moore, line 37: If rate is eg 3, then this won't work.

HotelCalifornia: C modulo takes the sign of the dividend, so you'd have 3 - ((3 - 5) % 5) -> 3 - (-2 % 5) -> 3 - (-2) -> 5

nathan-moore: I see now, yea. 🧠

HotelCalifornia: this is what i get for being clever i suppose

include/pros/imu.h
```diff
@@ -74,6 +74,27 @@ typedef struct __attribute__((__packed__)) euler_s {
  */
 int32_t imu_reset(uint8_t port);
 
+
+/**
+ * Set the Inertial Sensor's refresh interval in milliseconds.
+ *
+ * The rate may be specified in increments of 5ms, and will be rounded down to
+ * the nearest increment. The minimum allowable refresh rate is 5ms.
```
theol0403, line 82:
```suggestion
 * the nearest increment. The minimum allowable refresh rate is 5ms. The default 
 * rate is 10ms. Setting to less than 10ms will not increase the refresh rate on the 
 * user side, which is 10ms. However, it will guarantee that the data is more recent.
```
I guessed the wrap.

src/devices/vdml_imu.c
theol0403, line 37:
```suggestion
	rate = rate <= 5 ? 5 : rate - rate % 5
```

kunwarsahni01:
Do we need physical testing?

Issue #217, opened 2020-04-09
Controller screen character limit is lower than necessary


djava:
#### Expected Behavior:
You should be able to print enough to fill the screen, or whatever the VexOS limit is, if there is one

#### Actual Behavior:
There is a fairly arbitrary 15 character limit, which basically mean that unless you're printing "WWWWWWWWWWWWWWW", there is more space left over in the screen.

#### System information:
Platform: V5
PROS Kernel Version: All

#### Additional Information
I don't know if there is a limit, or what the maximum number of characters you can print on the screen is, but whatever that number is is probably what PROS should let you print.  I know that VEXCode allows up to 19 characters, I don't know if that's the hard limit.
Magik from QUEEN said that he was going to test what the limit is, you should ask him if I haven't already edited this issue to include it.

The actual change should be as simple as changing the `CONTROLLER_MAX_COLS` value in `src/devices/controller.c`

Really the only possible side effect I can think of is that it could maybe cause a bandwidth issue in the communication between the controller and the brain but I really doubt that it's significant enough to actually cause one.

nickmertin:
Tested with RMS python on latest firmware, it appears to support 19 characters, suggesting that this is the underlying VEXos limit.

djava:
It's probably worth testing it with the VexOS system calls to test that it's actually a system limitation and not just the limit that RMS and Vexcode both use

WillXuCodes:
So apparently, the line clearing function we have for the controller text never cleared consistently, as it would only clear consistently if the old text was less pixels wide than the 19 or 15 spaces used to clear the line.

This is what I mean by this, where the word being cleared here is wider than the spaces.
![image](https://user-images.githubusercontent.com/54247087/87382391-a59d7400-c54b-11ea-97d2-1abb206b8002.png)

What needs to be up for discussion then, is that:
Would we rather have allow users to use all 19 chars? Or attempt have more consistent clearing but with 15 (or less) chars?

djava:
I'd rather just open it up, possibly with a warning in the docs

nathan-moore:
I don't know any details here Will, but I'm guessing both robot mesh and vexcode both clear the line when they do what you're doing. So either the pr is doing something wrong with regards to that, or there's a better way of clearing a line that we're missing.

WillXuCodes:
Yeah... I've heard that this also happened when we were down to 15 lines (on the current build of pros) too. I'll try to find a different method.

djava:
[...]

PR #222, opened 2020-07-07
✨Increasing the Max String Size on the Controller Display

WillXuCodes:
#### Summary:
 VCS and RMS have had their controller max string sizes at 19 for a while, and PROS also having this wouldn't be too hard of a change. This was done by changing `CONTROLLER_MAX_COLS` under `src/devices/controller.c` to 19.

#### Motivation:
There was an issue opened by @djava requesting that we do this, but the discussion surround this died pretty suddenly so I decided to open a PR regarding this to restart the conversation

##### References (optional):
fixes https://github.com/purduesigbots/pros/issues/217

#### Test Plan:

- [x] Build and Upload to a robot with a 19 character display, see if it shows up.

### DIFF
--- a/src/devices/controller.c
+++ b/src/devices/controller.c
@@ -17,7 +17,7 @@
 #include "v5_api.h"
 #include "vdml/vdml.h"
 
-#define CONTROLLER_MAX_COLS 15
+#define CONTROLLER_MAX_COLS 19
 
 // From enum in misc.h
 #define NUM_BUTTONS 12
--- a/src/devices/controller.c
+++ b/src/devices/controller.c
@@ -259,7 +259,7 @@ int32_t controller_print(controller_id_e_t id, uint8_t line, uint8_t col, const 
 }
 
 int32_t controller_clear_line(controller_id_e_t id, uint8_t line) {
-	static const char* clear = "               ";
+	static const char* clear = "                   ";
 	return controller_print(id, line, 0, clear);
 }
 
--- a/src/devices/controller.c
+++ b/src/devices/controller.c
@@ -12,6 +12,8 @@
  */
 
 #include <stdio.h>
+#include <string.h>
+#include <assert.h>
 
 #include "kapi.h"
 #include "v5_api.h"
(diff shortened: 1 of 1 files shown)


kunwarsahni01:
LGTM image

nathan-moore:
You'll probably have to increase this line https://github.com/purduesigbots/pros/blob/337b4c5593e50a0a68368cf7567a101fbe0b48aa/src/devices/controller.c#L261

src/devices/controller.c
```diff
@@ -259,7 +259,7 @@ int32_t controller_print(controller_id_e_t id, uint8_t line, uint8_t col, const
 }
 
 int32_t controller_clear_line(controller_id_e_t id, uint8_t line) {
-	static const char* clear = "               ";
+	static const char* clear = "                   ";
```
nathan-moore, line 262: If you want to be fancy, you could assert that the lengths match up.

djava:
Seems worth a try at least putting it to like 50 and and just filling it with I's.
It's probably a VexOS limit at 19 if that's what RMS does but not necessarily

src/devices/controller.c
WillXuCodes: like, making it so that it prints " " * Max string length?

WillXuCodes:
Somebody with a controller could probably try that, but for pros itself the 19 limit seems reasonable given the small amount of space left with the 19 character word in the test case.

kunwarsahni01:
I tried increasing it to 20, the max number of characters that the controller will display is 19.

src/devices/controller.c
djava: Does `char*` clear = ' ' * CONTROLLER_MAX_COLS` not work in C?

WillXuCodes: I don't think so... you'd probably have to use a for loop for it.

nathan-moore: strings aren't first class citizens in C. While what you have will do something, it won't do what you expect unless you want to know what 32 * 19 is.

nathan-moore:
```c
kassert(strlen(clear) == CONTROLLER_MAX_COLS)
```
Is what I had in mind

Issue #224, opened 2020-07-13
improvements for pros4


nathan-moore:
There's been some discussion around changes we'd like for pros4. In particular, 

- LVGL sidelining  https://github.com/purduesigbots/pros/issues/168
- code reorganization https://github.com/purduesigbots/pros/issues/223
- remove direct dependence on libv5rts.a (possibly) 

are the major ones that we were discussing. Some of these changes breaking, so that's why we're thinking of a major version increment. It also coincides nicely with some other changes.

There was discussion elsewhere about what other changes people like to see, so part of the reason why this exists is to help document that.

HotelCalifornia:
possibly the migration to a more modern build system, like CMake

TixoRebel:
Also potentially add a HAL that thinly wraps the SDK to decouple the kernel from the SDK.

theol0403:
Switch the compiler to llvm/clang, which has better tooling and more optimizations than gcc, along with the move to cmake. 

Automated documentation, ie doxygen, whether it be a full migration to a new doxygen-based setup or take the autogenerated documentation and feed into read the docs.

Motor set PID command (and related helpers) being conditionally hidden using a macro, so that:
- it does not clutter up documentation (though doxygen commands could also be used to hide method from documentation)
- since it is a dangerous command, it is not trivially accessible or found by a new user yet available when needed.

Octogonapus:
> Switch the compiler to llvm/clang

I don't think clang has support for linking against arm yet, does it? Is it still "experimental" or something?

> better tooling

Does this really matter for us? We would just invoke `clang ....` instead of `gcc ....`. We aren't creating our own language here.

> more optimizations than gcc

Traditionally, gcc has been the better optimizing compiler between the two of them. They have pretty much caught up, though.

nickmertin:
> I don't think clang has support for linking against arm yet, does it? Is it still "experimental" or something?

ARM support is listed on the [LLD webpage](https://lld.llvm.org/).

> Traditionally, gcc has been the better optimizing compiler between the two of them. They have pretty much caught up, though.

Counterpoint: using Clang and LLD would enable us to support [Link-Time Optimization](https://llvm.org/docs/LinkTimeOptimization.html), which can greatly improve cross-unit optimizations, both for performance and code size. This could be especially useful for reducing monolithic executable size. There is (to my understanding) no equivalent system on the GNU stack.

Octogonapus:
> ARM support is listed on the LLD webpage.

[...]

Issue #223, opened 2020-07-10
PROS 4: Better Namespace Layout


HotelCalifornia:
#### what do you mean?

I was looking over the API today and I realized that we have a lot of classes sitting in the top level `pros` namespace that can probably be grouped for clarity of purpose. The most egregious example is all the warty `ADI*` classes.

#### what do you propose?

```cpp
namespace adi {
class AnalogIn;
class AnalogOut;
class DigitalIn;
class DigitalOut;
class Encoder;
class Gyro;
class Motor;
class Port;
class Ultrasonic;
}

/** these ones are less of a concern, but could still be done */

namespace device { // also I'm not set on this name
class Controller;
class Imu;
class Motor;
class Vision;
// plus the other new devices
}

namespace rtos {
class Mutex;
class Task;
}

// pros::Serial can either stay part of the top-level namespace or get moved to a new one like pros::misc or something
```

#### what about backwards compatibility?

```cpp
#define LEGACY_TYPEDEF(old_name, new_name) \
    using old_name [[deprecated("use " #new_name " instead")]] = new_name

LEGACY_TYPEDEF(ADIAnalogIn, pros::adi::AnalogIn);

// etc
```

#### ...but why?

really I'm just not sure why we ended up going with warty names for the ADI classes only, and I think it would just be a better move to group them using the c++ method for logical groupings 😄 

#### RFC

this is a request for comments, and not a concrete plan of action by any means. I'd love for people to let me know what they think 👍

WillXuCodes:
For device we could use "hardware" to be more clear about what's in that inner layer, I feel like "devices" would refer more to the brain or controller rather than pieces of hardware.

djava:
I think that v5 might be a better name for devices, since it's all the hardware specific to the v5 system.

nickmertin:
Another option for backwards-compatibility would be to make the nested namespaces `inline`. That would everything to be accessible through the `pros` namespace, not just types.

HotelCalifornia:
> Another option for backwards-compatibility would be to make the nested namespaces `inline`.

If I understand how inline namespaces work correctly, this on its own will not be sufficient in the ADI types case at least (where the actual type names would be changing)

nickmertin:
> where the actual type names would be changing

Good point, I forgot about that. Using `inline` namespaces would require keeping the unqualified names the same.

WillXuCodes:
> I think that v5 might be a better name for devices, since it's all the hardware specific to the v5 system.
[...]

PR #225, opened 2020-07-15
✨Better Code Organization

WillXuCodes:
#### Summary:
- ADI classes were given their own `adi` namespace under pros, and backwards compatibility was added with the old API with a deprecated warning.
- Controller and new devices were added to a `v5` namespace (under pros namespace).
- `mutex` and `Task` were given their own `rtos` namespace (also under pros).
- `rtos` and `v5` namespaces are `inline` so they already have backwards compatibility.

#### Motivation:
For better organization if we're to move to doxygen for our docs.

##### References (optional):
#223  #224 

#### Test Plan:
(This code builds properly from my deviceless testing)

- [ ] Try initiating different pros v5 devices, motors, and adi sensors/motors to see if they work and scope properly with the new code organization.


WillXuCodes:
@kunwarsahni01 you mind testing this when u got the time? Thanks.

kunwarsahni01:
Everything seems to be working here, tested with Controller and IMU. 
image

Octogonapus:
This is supposed to be fully backward compatible, right? If so, I would appreciate the chance to test it on OkapiLib.

WillXuCodes:
> This is supposed to be fully backward compatible, right? If so, I would appreciate the chance to test it on OkapiLib.

Yes, everything should be backwards compatible. The v5 devices use an inline namespace to do so, and the ADI stuff uses a macro that also prints a deprecated warning if the old api is used.

edjubuh:
> prints a deprecated warning if the old api is used.
When will the old API be removed? Why is the old API being removed? These explanations would be good to give to users when breaking their code

WillXuCodes:
> > prints a deprecated warning if the old api is used.
> 
> When will the old API be removed? Why is the old API being removed? These explanations would be good to give to users when breaking their code

As for the why, it's to do with us attempting to switch to doxygen for our documentation. This change isn't going to break their code due to the backwards compatibility measures in place, unless you're referring to when we actually remove it down the line.

edjubuh:
> This change isn't going to break their code due to the backwards compatibility measures in place, unless you're referring to when we actually remove it down the line.

If you're putting a warning on their code, it's breaking their code 😃

HotelCalifornia:
> > This change isn't going to break their code due to the backwards compatibility measures in place, unless you're referring to when we actually remove it down the line.
> 
> If you're putting a warning on their code, it's breaking their code 😃

so this is where the discussion of bumping major versions comes into play. the goal i think is that we'll try to maintain backwards compat (i.e. the code users have works albeit with warnings) for the short term but since it's a major version upgrade there are no guarantees

edjubuh:
> As for the why, it's to do with us attempting to switch to doxygen for our documentation.

I didn't really get the sense that the maintaining backwards compatibility with the current API has a significant impact on improving documentation. I think I'm missing something here.

It's worth mentioning that PROS 2 projects still today should compile in PROS 3. We have a header which #defines all the PROS 2 APIs into PROS 3 functions.

nickmertin:
> so this is where the discussion of bumping major versions comes into play. the goal i think is that we'll try to maintain backwards compat (i.e. the code users have works albeit with warnings) for the short term but since it's a major version upgrade there are no guarantees

> It's worth mentioning that PROS 2 projects still today should compile in PROS 3. We have a header which #defines all the PROS 2 APIs into PROS 3 functions.

So perhaps, keep the backwards-compatibility for a PROS 3 release, then break compatibility with PROS 4? Otherwise, we run into the issue of having some things break on 4.0, while others break on 4.1, etc. IMO breaking everything that we plan to break in one release is better because it limits the number of releases in which people have to worry about breaking changes, and in the long run doesn't really affect the compatibility. Thoughts?

HotelCalifornia:
IMO this discussion has gone a little bit off the rails. the specific purpose of this PR is to better organize PROS classes into logical units. deprecating the old ADI class names is, however disruptive to users' workflow, not a _breaking_ change in my mind. code will still compile, and those who choose to ignore the warnings may do so at their own risk. beyond that, there will be no tangible change to the end-user API, except that some classes _may_, if the user so chooses, be referred to by more fully-qualified names.

if and when version 4 lands, we can certainly create another "legacy header" for PROS 3 in the same way that one was created for PROS 2. however, that change specifically is out of the scope of this PR.

WillXuCodes:
For PROS4, this will probably be one of the later steps in the process of kernel changes (especially if we're writing a HAL). Needs to be revisited later

HotelCalifornia:
that's why I converted it into a draft PR... no need to close it

WillXuCodes:
Fixed up this PR after merges with the ext-adi changes, not much other than renaming and fixing up a couple issues with merge conflicts. Needs re-testing before PROS-4 work and all that.

WillXuCodes:
Closing since all work for PROS 4 will be done on the develop-pros-4 branch

PR #226, opened 2020-07-20
🐛 Fix controller_clear_line

WillXuCodes:
#### Summary:
-Used a macro to cleanup repeating code.
-Added a test implementation for clearing controller text by passing a 0
length/size string (as jpearman said on VEXForum is supposed to do:
https://www.vexforum.com/t/pros-controller-screen-problem/83186/5?u=ranchsauce).

(I'm aware the way that I set up the controller_clear_line method to test this _may_ be incorrect but I'll fix it later if it is).

**Code compiles, just needs somebody with device to test methods.**

#### Test Plan:
- [x] Initiate a controller and try testing a few general methods to make sure the macro works.
- [x] Try using a 19 char word such as "Adrenocorticotropic" and test if the cleartext function works.


WillXuCodes:
(Wait I forgot to push most recent code, still broken currently)

WillXuCodes:
Ok it's good now.

WillXuCodes:
@kunwarsahni01 when you got time, you mind testing this? Jpearman just confirmed that sending an empty string to a line should work in clearing a line: https://www.vexforum.com/t/pros-controller-screen-problem/83186/24

HotelCalifornia:
> If you are working with the single line text display underneath the match clock etc. then sending 0 length string does not work, we may fix this in a future release.

> Jpearman just confirmed that sending an empty string to a line should work in clearing a line: https://www.vexforum.com/t/pros-controller-screen-problem/83186/24

> If you are working with the single line text display underneath the match clock etc. then sending 0 length string does not work, we may fix this in a future release.

?

WillXuCodes:
> > Jpearman just confirmed that sending an empty string to a line should work in clearing a line: https://www.vexforum.com/t/pros-controller-screen-problem/83186/24
> 
> > If you are working with the single line text display underneath the match clock etc. then sending 0 length string does not work, we may fix this in a future release.
> 
> ?

"If you are using more than one line of text (ie. the screen is dedicated to user code text) then sending a 0 length string does work, it will correctly clear the whole line."
Wait is this a different thing?

HotelCalifornia:
yes, it will work in the case where there are multiple lines being shown on the controller LCD. it will, however, not work when there is only a single line. if i recall correctly, the tests being done in #222 all involved the single line case, did they not? i only bring this up so that we can make clear that this is not a fix-_all_ until there is a new vexOS release that addresses that one case.

WillXuCodes:
New VEXos update has been released regarding this, needs testing once we have time to do so.

WillXuCodes:
This is confirmed to work.imageimage

src/devices/controller.c
```diff
@@ -259,8 +148,16 @@ int32_t controller_print(controller_id_e_t id, uint8_t line, uint8_t col, const
 }
 
 int32_t controller_clear_line(controller_id_e_t id, uint8_t line) {
-	static const char* clear = "               ";
-	return controller_print(id, line, 0, clear);
+	uint8_t port;
+	CONTROLLER_PORT_MUTEX_TAKE(id,port);
+	static char* clear = "";
+	uint32_t rtn_val = vexControllerTextSet(id, ++line, 1, clear);
+	internal_port_mutex_give(port);
```
nathan-moore, line 155: Ideally we would switch on the vexSystemVersion like we do for controller_clear.

WillXuCodes:
Confirmed to still work with these last couple changes.

src/devices/controller.c
```diff
@@ -260,9 +148,16 @@ int32_t controller_print(controller_id_e_t id, uint8_t line, uint8_t col, const
...
+	if (vexSystemVersion() > 0x01000000) {
+		uint8_t port;
+		CONTROLLER_PORT_MUTEX_TAKE(id, port);
+		static char* clear = "";
+		vexControllerTextSet(id, ++line, 1, clear);
+		internal_port_mutex_give(port);
+		return 1;
+	} else {
+		return PROS_ERR;
+	}
```
nathan-moore, line 160: The vexSystemVersion is off. It should be whatever it was just updated to that fixed the clearing bug. For the case where we're on the older version of vex os and it's a case that's potentially buggy, I think we should call into the old version. It may not work the best, but it would prevent us from regressing currently working code if they don't update.

If you think it's too much effort for what it's worth, we can axe this idea, go back to what you had, and tell them to update vexos.

WillXuCodes: Oooh ok I get it now. How do I indicate below version 1.12 though?

nathan-moore: See what vexSystemVersion returns on a 1.12 vexOS. My guess is something like 0x01C00000. If vexSystemVersion is below that, then that's what you're asking.

version
```diff
@@ -1 +1 @@
-3.2.0
+3.2.1
```
HotelCalifornia, line 1: this should not be done here, please revert

HotelCalifornia (changes requested):

version
```diff
@@ -1 +1 @@
-3.2.0
+3.2.0
```
HotelCalifornia, line 1:
```suggestion
3.2.0

```

src/devices/controller.c
```diff
@@ -260,9 +148,20 @@ int32_t controller_print(controller_id_e_t id, uint8_t line, uint8_t col, const
 }
 
 int32_t controller_clear_line(controller_id_e_t id, uint8_t line) {
-	static const char* clear = "                   ";
-	kassert(strlen(clear) == CONTROLLER_MAX_COLS);
-	return controller_print(id, line, 0, clear);
+	uint8_t port;
+	CONTROLLER_PORT_MUTEX_TAKE(id, port);
+	static char* clear;
```
HotelCalifornia, line 153:
```suggestion
	static char* clear;
```
this is unused and can be removed

HotelCalifornia (approved):
lgtm

Issue #229, opened 2020-08-04
RFC: RAII interface for PROS Mutexes


nickmertin:
## Summary
Essentially, implement a proper RAII interface for PROS mutexes, to make locking and unlocking fully safe. My preferred method of doing so would be to make `pros::Mutex` conform to [*BasicLockable*](https://en.cppreference.com/w/cpp/named_req/BasicLockable), such that `std::lock_guard` can be used with it. This should be relatively simple. Conforming to [*TimedLockable*](https://en.cppreference.com/w/cpp/named_req/TimedLockable) would extend this to full support for `std::unique_lock`, and similarly conforming to [*Lockable*](https://en.cppreference.com/w/cpp/named_req/Lockable) would extend to full support for `std::scoped_lock`. None of these require introducing new behaviour or features to `pros::Mutex`; rather, they can be implemented simply by extending the existing interface to meet the named requirements.

PR #231, opened 2020-08-07
Added mutex_delete functionality

Buzzec:
#### Summary:
Added a function to allow for deleting a mutex within a c++ context as well as a specific c function in order to fix the possible memory leak.

Currently this code will cause the v5 to freeze to an unrecoverable state after <15 seconds:
`while(true){pros::Mutex{});}`

Some possible issues:
-The function to free memory is called pros::Mutex::free, delete would have been preferable to match the c api but is a reserved word.
-This function cannot be implemented within a destructor because for a mutex to fulfill it's purpose multiple objects with the same memory pointer must exist. This can be done utilizing reference counting but that is a larger change outside the scope of this.

Future improvements
-Adding reference counting to make destruction automatic
-Making free also set the pointer to nullptr, however this will not effect other instances unless pros::Mutex is changed to hold a mutex_t**
-Adding scoped locking functionality akin to #229

#### Motivation:
pros::Mutex did not have a way for the user to free its memory and the c api did not have a clear way to do the same.

#### Test Plan:

Run the following code:
`while(true){pros::Mutex{}.free();}`
The v5 should not crash due to memory leakage, test for ~1 min.

I have tested the concept by running the following code with header changes for public access to pros::Mutex::mutex, however I cannot build the pros kernel myself.
`while(true){pros::c::sem_delete(pros::Mutex{}.mutex);}`
This code was tested up to 3 min.

### DIFF
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -356,6 +356,11 @@ class Mutex {
 	 */
 	bool give(void);
 
+	/**
+	 * Should be called only on the last instance to free the memory the mutex takes up.
+	 */
+	void free(void);
+
 	private:
 	mutex_t mutex;
 };
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -356,13 +356,8 @@ class Mutex {
 	 */
 	bool give(void);
 
-	/**
-	 * Should be called only on the last instance to free the memory the mutex takes up.
-	 */
-	void free(void);
-
 	private:
-	mutex_t mutex;
+	std::shared_ptr<std::remove_pointer_t<mutex_t>> mutex;
 };
 
 /**
--- a/src/rtos/rtos.cpp
+++ b/src/rtos/rtos.cpp
@@ -96,7 +96,7 @@ using namespace pros::c;
     return task_get_count();
   }
 
-  Mutex::Mutex(void) : mutex(std::shared_ptr<std::remove_pointer_t<mutex_t>>{mutex_create(), mutex_delete}) { }
+  Mutex::Mutex(void) : mutex(mutex_create(), mutex_delete) { }
 
   bool Mutex::take(std::uint32_t timeout) {
     return mutex_take(mutex.get(), timeout);
--- a/src/rtos/rtos.cpp
+++ b/src/rtos/rtos.cpp
@@ -105,4 +105,8 @@ using namespace pros::c;
   bool Mutex::give(void) {
     return mutex_give(mutex);
   }
+
+  void Mutex::free(void) {
+	  mutex_delete(mutex);
+  }
 }
--- a/src/rtos/rtos.cpp
+++ b/src/rtos/rtos.cpp
@@ -96,10 +96,6 @@ using namespace pros::c;
     return task_get_count();
   }
 
-  class MutexDeleter{
-
-  };
-
   Mutex::Mutex(void) : mutex(std::shared_ptr<std::remove_pointer_t<mutex_t>>{mutex_create(), mutex_delete}) { }
 
   bool Mutex::take(std::uint32_t timeout) {
--- a/src/rtos/rtos.cpp
+++ b/src/rtos/rtos.cpp
@@ -96,17 +96,17 @@ using namespace pros::c;
     return task_get_count();
   }
 
-  Mutex::Mutex(void) : mutex(mutex_create()) { }
+  class MutexDeleter{
+
+  };
+
+  Mutex::Mutex(void) : mutex(std::shared_ptr<std::remove_pointer_t<mutex_t>>{mutex_create(), mutex_delete}) { }
 
   bool Mutex::take(std::uint32_t timeout) {
-    return mutex_take(mutex, timeout);
+    return mutex_take(mutex.get(), timeout);
   }
 
   bool Mutex::give(void) {
-    return mutex_give(mutex);
-  }
-
-  void Mutex::free(void) {
-	  mutex_delete(mutex);
+    return mutex_give(mutex.get());
   }
 }
--- a/src/rtos/semphr.c
+++ b/src/rtos/semphr.c
@@ -203,7 +203,7 @@ uint8_t mutex_take(mutex_t mutex, uint32_t timeout) {
 }
 
 void mutex_delete(mutex_t mutex){
-  queue_delete((queue_t)(mutex));
+  sem_delete((sem_t)(mutex));
 }
 
 /**
--- a/src/rtos/semphr.c
+++ b/src/rtos/semphr.c
@@ -202,6 +202,10 @@ uint8_t mutex_take(mutex_t mutex, uint32_t timeout) {
 	return xQueueSemaphoreTake( ( mutex ), ( timeout ) );
 }
 
+void mutex_delete(mutex_t mutex){
+  queue_delete((queue_t)(mutex));
+}
+
 /**
  * semphr. h
  * <pre>sem_t sem_binary_create( void )</pre>
--- a/include/pros/rtos.h
+++ b/include/pros/rtos.h
@@ -403,6 +403,14 @@ bool mutex_take(mutex_t mutex, uint32_t timeout);
  */
 bool mutex_give(mutex_t mutex);
 
+/**
+ * Deletes a mutex
+ *
+ * \param mutex
+ *        Mutex to unlock.
+ */
+void mutex_delete(mutex_t mutex);
+
 #ifdef __cplusplus
 }  // namespace c
 }  // namespace pros


src/rtos/semphr.c
```diff
@@ -202,6 +202,10 @@ uint8_t mutex_take(mutex_t mutex, uint32_t timeout) {
 	return xQueueSemaphoreTake( ( mutex ), ( timeout ) );
 }
 
+void mutex_delete(mutex_t mutex){
+  queue_delete((queue_t)(mutex));
```
nathan-moore, line 206: sem_delete is what freeRTOS has doc'd on their website. It's the same code, but can you use that instead?

nathan-moore:
> This function cannot be implemented within a destructor because for a mutex to fulfill it's purpose multiple objects with the same memory pointer must exist. This can be done utilizing reference counting but that is a larger change outside the scope of this.

I think it's worth doing the more idiomatic way the first time, so we don't have a somewhat redundant free function hanging around. I think it's straightforward with a shared pointer with a custom destructor. 

The other option to consider to consider is to take a breaking change and remove the copy constructor. I don't think I've ever seen it used, so it might be acceptable.

nickmertin:
> The other option to consider to consider is to take a breaking change and remove the (implicit) copy constructor. I don't think I've ever seen it used, so it might be acceptable. We should probably go with the reference counting though.

I agree with this, but save the breaking change for PROS 4 maybe? In the meantime, the idiomatic (i.e. RAII) solution would probably be best.

HotelCalifornia:
Using a shared pointer with a custom deleter to wrap the internal mutex handle would I think nicely solve a lot of the issues mentioned here

src/rtos/semphr.c
Buzzec: This is the same way sem_delete is implemented on line 522 of this file

nathan-moore: I know. Using sem_delete doesn't break abstraction barriers though and is what free rtos recommends.

Buzzec:
I've updated it to delete with a shared pointer and custom deleter

Buzzec:
> 
> 
> > The other option to consider to consider is to take a breaking change and remove the (implicit) copy constructor. I don't think I've ever seen it used, so it might be acceptable. We should probably go with the reference counting though.
> 
> I agree with this, but save the breaking change for PROS 4 maybe? In the meantime, the idiomatic (i.e. RAII) solution would probably be best.

This defeats the purpose of a mutex because that would eliminate the ability to have two copies of the same mutex, meaning two threads could not hold the same mutex

src/rtos/semphr.c
Buzzec: Are you saying to use sem_delete from this file or freeRTOS's vSemaphoreDelete? I think you might have been saying the former while I assumed you meant the later.

nickmertin:
> This defeats the purpose of a mutex because that would eliminate the ability to have two copies of the same mutex, meaning two threads could not hold the same mutex.

I agree that we shouldn't remove the copy constructor, but I stand by the suggestion of using a shared pointer with a custom deleter. In this way, the `pros::Mutex` object is truly treated as a handle to the underlying mutex (which is what it is anyway), and the underlying mutex is destroyed once the last copy is removed. Of course, it would be pertinent to also add a `release`-esque function to get a C `mutex_t` out of it, in the same way as `release` lets you get a C pointer out of a shared pointer object.

HotelCalifornia:
> This defeats the purpose of a mutex because that would eliminate the ability to have two copies of the same mutex, meaning two threads could not hold the same mutex.

```cpp
pros::Mutex mtx;

void task_a(void*) {
    mtx.take();
    // ...
    mtx.give();
}

void task_b(void*) {
    mtx.take();
    // ...
    mtx.give();
}
```

nickmertin:
@HotelCalifornia as Brett pointed out, that assumes static lifetime, which obviously we are not assuming if we are creating a `free()` or equivalent.

HotelCalifornia:
> @HotelCalifornia as Brett pointed out, that assumes static lifetime, which obviously we are not assuming if we are creating a `free()` or equivalent.

yeah i know, just pointing out that having no copy constructor doesn't actually prevent someone from sharing an object between tasks

src/rtos/semphr.c
nathan-moore: This one: https://github.com/purduesigbots/pros/blob/7862fc52e2f351cb7977305c4595bfb50c3a87fc/src/rtos/semphr.c#L518

Buzzec: It references the other function now

src/rtos/rtos.cpp
```diff
@@ -96,13 +96,13 @@ using namespace pros::c;
     return task_get_count();
   }
 
-  Mutex::Mutex(void) : mutex(mutex_create()) { }
+  Mutex::Mutex(void) : mutex(std::shared_ptr<std::remove_pointer_t<mutex_t>>{mutex_create(), mutex_delete}) { }
```
nickmertin, line 99: I'm pretty sure this can be simplified to:
```suggestion
  Mutex::Mutex(void) : mutex{mutex_create(), mutex_delete} { }
```

Buzzec: You are very correct

nathan-moore:
> Edit: Thinking further, if you assume (effectively) static lifetimes and pass around pointers/references then there is no need for there to be a destructor and you run into issues of enforcing the effectively static lifetimes leading to ambiguous problems. You could fix this by saying the user should wrap this in a shared pointer or similar but I think it's cleaner for the mutex to handle it itself.

IMHO having the mutex object itself copyable is extremely odd. In every stdlib I know, it's either a pointer copy, forbidden (in the stl), or ub. It's the best option here, but it's weird. (Does copying a mutex make a new mutex or a handle to the one copied?) In case it wasn't clear, I meant removing the copy constructor and adding this destructor.

@kunwarsahni01 @Buzzec Either of you have a chance to give the final version a test run? 
[kernel@3.2.1-dirty.72.7862fc5.zip](https://github.com/purduesigbots/pros/files/5048185/kernel%403.2.1-dirty.72.7862fc5.zip)

include/pros/rtos.hpp
```diff
@@ -357,7 +357,7 @@ class Mutex {
 	bool give(void);
 
 	private:
-	mutex_t mutex;
+	std::shared_ptr<std::remove_pointer_t<mutex_t>> mutex;
```
nathan-moore, line 360: I don't think you need this remove pointer as well.

nickmertin: What he's doing is avoiding an additional layer of heap allocation and indirection, by using the fact that `mutex_t` is a pointer. I support the way he's currently implemented it.

nickmertin:
> IMHO having the mutex object itself copyable is extremely odd. In every stdlib I know, it's either a pointer copy, forbidden (in the stl), or ub. It's the best option here, but it's weird. (Does copying a mutex make a new mutex or a handle to the one copied?) In case it wasn't clear, I meant removing the copy constructor and adding this destructor.

I disagree, I think having it copyable is quite useful (and also clear what that means). In any event, if people wanted to make it copyable and it wasn't, they'd just put it in a `shared_ptr`, which adds another layer of indirection for no reason. As long as the docs make it clear what copying means (i.e. that the object is a handle to a mutex, not the mutex itself, and the handle is copyable), then I think the current approach is the best.

Buzzec:
> 
> 
> > Edit: Thinking further, if you assume (effectively) static lifetimes and pass around pointers/references then there is no need for there to be a destructor and you run into issues of enforcing the effectively static lifetimes leading to ambiguous problems. You could fix this by saying the user should wrap this in a shared pointer or similar but I think it's cleaner for the mutex to handle it itself.
> 
> IMHO having the mutex object itself copyable is extremely odd. In every stdlib I know, it's either a pointer copy, forbidden (in the stl), or ub. It's the best option here, but it's weird. (Does copying a mutex make a new mutex or a handle to the one copied?) In case it wasn't clear, I meant removing the copy constructor and adding this destructor.

I would agree with you but for that pros::Mutex itself is a pointer to the heap while most stdlib mutexes keep the data with it (on the stack or the heap).

> @kunwarsahni01 @Buzzec Either of you have a chance to give the final version a test run?
> [kernel@3.2.1-dirty.72.7862fc5.zip](https://github.com/purduesigbots/pros/files/5048185/kernel%403.2.1-dirty.72.7862fc5.zip)

I would if I knew how, I can't build the kernel myself and the linked one doesn't seem to have the changes (based on the headers).

nickmertin:
> I would agree with you but for that pros::Mutex itself is a pointer to the heap while most stdlib mutexes keep the data with it (on the stack or the heap).

Another point here is that the C API's `mutex_t` is freely copyable (because, as you point out, it's a pointer), and there's no reason to artificially restrict the C++ API beyond it.

nathan-moore:
@Buzzec Sorry about that. I updated the link I originally posted with the right stuff. LMK (here or on discord) if you need help figuring out conductor

Buzzec:
Build works as expected, tested with initially proposed test code.

PR #233, opened 2020-08-08
🥅 Remove redundant library links

nathan-moore:
#### Summary:
Remove linking in the compiler provided libc and libm and libsupc++ as its a subset of libc++

- [x] forkner, and hasselhoff still compile (hot/cold both on and off)
- [x] Okapi compiles, and some other c++ heavy code bases (theo tested this)
- [x] Some smoke test to make sure things actually work (https://github.com/nathan-moore/pros/tree/remove-linkings) It hasn't been run though.

Will happily add more.

Note that this fixed some compiler errors when using a very recent version of gcc.


HotelCalifornia (approved):
confirmed that this fixes linking issues related to multiple definitions of `__retarget_lock_init`, `__retarget_lock_init_recursive`, `__retarget_lock_close`, `__retarget_lock_close_recursive`, `__retarget_lock_acquire`, `__retarget_lock_acquire_recursive`, `__retarget_lock_try_acquire`, `__retarget_lock_try_acquire_recursive`, `__retarget_lock_release`, and `__retarget_lock_release_recursive` when compiling with the arm-none-eabi-gcc 10 preview

PR #234, opened 2020-08-09
pros::Task::notify_take static and pros::Task::operator= stl compliant

Buzzec:
#### Summary:
Makes pros::Task::notify_take static and pros::Task::operator= stl compliant.

#### Motivation:
pros::Task::notify_take used to require a task handle to be used but that task handle was never used in the function. With it being static it can now be called without a handle without breaking current user code.

pros::Task::operator= used to return void and this makes it return pros::Task&, a good explanation for why this is better can be found here: [explanation](https://stackoverflow.com/questions/42335200/assignment-operator-overloading-returning-void-versus-returning-reference-param)

### DIFF
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -152,7 +152,7 @@ class Task {
 	 *        A task handle from task_create() for which to create a pros::Task
 	 *        object.
 	 */
-	void operator=(const task_t in);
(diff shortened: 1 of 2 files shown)


HotelCalifornia:
Sure wish we had a good way to refer to the current task, but.... I suppose this is as good as it gets

PR #235, opened 2020-08-09
LL Function or Data Detection Flag

WillXuCodes:
Added a flag that detects if a linked list node is a function or a data node. When removing data or function nodes in `linked_list_remove_data` or `linked_list_remove_func` it will skip over any nodes that are not data or func respectively.

Can confirm it compiles.


edjubuh:
When does this happen?

WillXuCodes:
> When does this happen?

In the code, there were these two comments with a TODO in it:
// NOTE: this will likely break if the list has intermixed data and function ptrs
// TODO: add tag field to ll_node_s_t to easily check what is present

While I have not personally seen this breaking with an intermixed data and function list, I added this as I assumed the comparison between a mismatched data and function pointer could _potentially_ lead to something breaking.

src/common/linkedlist.c
```diff
@@ -75,23 +75,22 @@ void linked_list_append_func(linked_list_s_t* list, generic_fn_t func) {
...
 	ll_node_s_t* p = NULL;
 	while (it != NULL) {
-		if (it->payload.func == func) {
-			if (p == NULL)
-				list->head = it->next;
-			else
-				p->next = it->next;
-			kfree(it);
+		if(it->isFunc) { //Outer if loop for protective wrapping
+			if (it->payload.func == func) {
```
nathan-moore, line 85: Shouldn't the pointer be enough to see what path we should take? Eg the caller knows based on context.

nathan-moore: It's just a pointer compare, so I don't see how it would matter anyway.

WillXuCodes: Makes sense, I guess we could leave this here in the case that we may need to create a mixed function/data LL somewhere down the line. It'll just be a future proofing change that may be merged if necessary.

nathan-moore: Just to be clear, the the potential error here is that if there is a non-pointer added to the union, then it could potentially be re-interpreted casted to a valid pointer and be removed accidentally.

Perhaps just updating the comments would be good enough, and putting one on the union in ll_node_s on what's going on here.

WillXuCodes:
NOTE: This PR is ready for re-review. After a year of thinking on this "heavily" I've come to the conclusion that updating the comments is good enough and those who are using our built in LL in the first place should have enough foresight to not mix the two pointer types, and that updating the comments is good enough.

(not that many people use it in the first place... given that it's mostly for system level calls related to tasks).

PR #236, opened 2020-08-09
Removed non-sensical imu calibrating check in imu_get_status

WillXuCodes:
Removed a ERROR_IMU_STILL_CALIBRATING in vdml_imu.c, which doesn't make sense to check if the imu was still calibrating and not return this status.


Octogonapus:
Bump

nathan-moore:
@kunwarsahni01 Can you also test this while you're in the lab?

kunwarsahni01 (approved):
Tested with hardware and should be good to go.

PR #237, opened 2020-08-09
Added the ability to start a task with delay

Buzzec:
#### Summary:
Makes pros::Task have the ability to be started with a delay. This change does not affect existing user code as it adds all parameters at the end of constructors with a default parameter.

#### Motivation:
Wanted to start a discussion about adding the ability to delay a task with a concrete example as a backbone.

#### Test Plan:
- prove new changed don't affect  existing user code
- prove new changes launch functions as expected from a delayed state

### DIFF
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -30,6 +30,18 @@
 #include <type_traits>
 
 namespace pros {
+//Internal task start data storage class
+struct TaskStartData{
+	std::function<void(void*)> function;
+  void* parameters;
+	std::uint32_t prio;
+	std::uint16_t stack_depth;
+	std::string name;
+
+	TaskStartData();
+  TaskStartData(task_fn_t function, void* parameters, uint32_t prio, uint16_t stack_depth, const char* name);
+};
+
 class Task {
 	public:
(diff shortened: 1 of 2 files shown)


nickmertin:
I like the idea here, I'm not sure if `delay` is a good name for it though, as users may infer that it means a specific amount of time (due to familiarity with functions such as `pros::delay`). Maybe `defer`, or something like that?

nathan-moore:
> This change does not affect existing user code as it adds all parameters at the end of constructors with a default parameter.

While not source breaking, this will break all currently compiled code (eg okapi), fyi.

HotelCalifornia:
It would be preferable if this functionality was available in the C API as well

nathan-moore:
Is there a reason why this should be added to the pros kernel? You can do it in user land easily and it adds some weird api complexity.

Buzzec:
I think it's a nice thing to have because it's something I want when I create a new project and want something in a short time frame. It's something that adds more functionality, but that can be totally ignored if you don't want to use it.

HotelCalifornia:
@Buzzec will you add this to the C API as well?

PR #238, opened 2020-08-10
Fixed queue_recv doc comment

Buzzec:
#### Summary:
Fixed a documentation comment for queue_recv

### DIFF
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -307,9 +307,10 @@ bool queue_peek(queue_t queue, void* const buffer, uint32_t timeout);
  * \param buffer
  *        Pointer to a buffer to which the received item will be copied
  * \param timeout
- *        Time to wait for space to become available. A timeout of 0 can be used
- *        to attempt to post without blocking. TIMEOUT_MAX can be used to block
- *        indefinitely.
+ *        The maximum amount of time the task should block
+ *        waiting for an item to receive should the queue be empty at the time
+ *        of the call. queue_recv() will return immediately if timeout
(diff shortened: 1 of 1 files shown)


nathan-moore:
Can you fix here as well? 
https://github.com/purduesigbots/pros/blob/7862fc52e2f351cb7977305c4595bfb50c3a87fc/include/pros/apix.h#L291

include/pros/apix.h
```diff
@@ -307,9 +307,10 @@ bool queue_peek(queue_t queue, void* const buffer, uint32_t timeout);
  * \param buffer
  *        Pointer to a buffer to which the received item will be copied
  * \param timeout
- *        Time to wait for space to become available. A timeout of 0 can be used
- *        to attempt to post without blocking. TIMEOUT_MAX can be used to block
- *        indefinitely.
+ *        The maximum amount of time the task should block
+ *        waiting for an item to receive should the queue be empty at the time
+ *        of the call. queue_recv() will return immediately if timeout
+ *        is zero and the queue is empty.
```
nathan-moore, line 313:
```suggestion
 *        of the call. queue_recv() will return immediately if the timeout
 *        is zero.
```

I think it would also be nice to keep the mention to TIMEOUT_MAX, but ¯\\\_(ツ)_/¯

Buzzec: This is copied from the freeRTOS doc comment, there is a chance this blocks if the queue is not empty, it will yield to allow all the waiting publishers to publish before even checking the timeout.

Issue #197, opened 2020-01-13
Calling adi_ultrasonic_get hangs the program


Octogonapus:
#### Expected Behavior:
Calling `adi_ultrasonic_get` should read the ultrasonic and return the value like normal.

#### Actual Behavior:
Calling `adi_ultrasonic_get` hangs the program.

#### Steps to reproduce:
Run this code:
```cpp
void opcontrol() {
    pros::c::adi_ultrasonic sonic = pros::c::adi_ultrasonic_init(7, 8);
    printf("Errno %d:%s\n", errno, strerror(errno));
  while(true) {
    std::cout << pros::c::adi_ultrasonic_get(sonic) << '\n';
        ;
    pros::delay(20);
  }
}
```

#### System information:
Platform: V5
PROS Kernel Version: v3.2.0

#### Additional Information
> Ok so with yellow input on 7 and orange output on 8 both the cpp and c api do not work. The cpp api gives me a 0 and the c api doesn't 'run'. With yellow input on 8 and orange output on 7, the c api does not 'run', but the cpp api does work. With all 4 of these tests the ping is set to 7 and the echo is set to 8.

nathan-moore:
From a preliminary look, it likely hangs in vexDeviceAdiPortConfigGet, as that's the only difference I see between the c and c++ apis.

PR #239, opened 2020-08-13
Add Support for the ADI Expander

djava:
#### Summary:
Changed the C and C++ API for ADI devices to support changing the selected smart port, which lets you use the ADI Expander.

I used the format `function(smart_port, adi_port)`, i.e.  `int32_t adi_port_get_value(uint8_t smart_port, uint8_t adi_port);`, as was suggested by @kunwarsahni01 
The intention is for people to use the `INTERNAL_ADI_PORT` macro for those ports.
I changed the C types for ADI Devices (i.e. `adi_encoder_t`) from`typedef int32_t adi_encoder_t` to 
`
typedef struct {
	int32_t smart_port;
	int32_t adi_port;
} adi_encoder_t;
` to include the smart port and the adi port, since now we have to worry about storing both.

#### Motivation:
PROS should support the 3-wire expander because it's a VRC-legal electronic part and PROS supports all the VRC-legal electronics parts.

#### Test Plan:
Try all the ADI devices through a 3-wire expander and through the internal brain ports.

PR #245, opened 2020-08-30
✨ Add Support for the ADI Expander

kunwarsahni01:
#### Summary:
Created a new API for the Expander

#### Motivation:
PROS should support the 3-wire expander because it's a VRC-legal electronic part and PROS supports all the VRC-legal electronics parts.

#### Refs:
Supercedes #239

#### Test Plan:
- [x] Encoder
- [x] Button
- [x] Ultrasonic
- [x] Motor
- [x] Gyro


HotelCalifornia:
note to maintainers: add @djava to co-authored-by line in the squash merge commit

WillXuCodes:
Note: Latest commit above could be potential fix to #197

WillXuCodes:
This is confirmed to fix #197

src/devices/vdml_ext_adi.c
```diff
@@ -0,0 +1,416 @@
...
+		errno = EADDRINUSE;                                                                                   \
+		return PROS_ERR;                                                                                      \
+	}
+
+#define validate_motor(device, port)                                                                      \
+	adi_port_config_e_t config = (adi_port_config_e_t)vexDeviceAdiPortConfigGet(device->device_info, port); \
+	if (config != E_ADI_LEGACY_PWM && config != E_ADI_LEGACY_SERVO) {                                       \
+		errno = EADDRINUSE;                                                                                   \
+		return PROS_ERR;                                                                                      \
+	}
```
nathan-moore, line 75: Alex and I talked about this a bit on slack/discord. I'd prefer to move away from these macros as they can cause a lot of error (I'll comment on atleast some of them I saw). You can commit this and I can fix it up later as I don't want to hold up the pr on this, or you can merge in some of the changes from the branch I shared earlier.

At the very least, they shouldn't be duplicated

src/devices/vdml_ext_adi.c
```diff
@@ -0,0 +1,416 @@
...
+	transform_adi_port(adi_port);
+	claim_port_i(smart_port - 1, E_DEVICE_ADI);
+	vexDeviceAdiValueSet(device->device_info, adi_port, value);
+	return_port(smart_port - 1, 1);
+}
+
+int32_t ext_adi_analog_calibrate(uint8_t smart_port, uint8_t adi_port) {
+	transform_adi_port(adi_port);
+	claim_port_i(smart_port - 1, E_DEVICE_ADI);
+	validate_type(device, adi_port, E_ADI_ANALOG_IN);
```
nathan-moore, line 137: This can abandon a mutex.

src/devices/vdml_ext_adi.c
```diff
@@ -0,0 +1,416 @@
...
+	}
+	adi_data_s_t* const adi_data = &((adi_data_s_t*)(device->pad))[adi_port];
+	adi_data->analog_data.calib = (int32_t)((total + 16) >> 5);
+	return_port(smart_port - 1, (int32_t)((total + 256) >> 9));
+}
+
+int32_t ext_adi_analog_read(uint8_t smart_port, uint8_t adi_port) {
+	transform_adi_port(adi_port);
+	claim_port_i(smart_port - 1, E_DEVICE_ADI);
+	validate_type(device, adi_port, E_ADI_ANALOG_IN);
```
nathan-moore, line 151: This can abandon a mutex, ect...

src/devices/vdml_ext_adi.c
```diff
@@ -0,0 +1,416 @@
...
+		errno = EINVAL;                             \
+		return PROS_ERR;                            \
+	}
+
+adi_port_config_e_t ext_adi_port_get_config(uint8_t smart_port, uint8_t adi_port) {
+	transform_adi_port(adi_port);
+	claim_port_i(smart_port - 1, E_DEVICE_ADI);
+	adi_port_config_e_t rtn = (adi_port_config_e_t)vexDeviceAdiPortConfigGet(device->device_info, adi_port);
+	return_port(smart_port - 1, rtn);
+}
```
nathan-moore, line 111: If you have to duplicate the api, how hard would it to re-wire the existing functions to call into this one? Having two functions with the same code copy and pasted isn't that great.

src/devices/vdml_ext_adi.c
```diff
@@ -0,0 +1,416 @@
...
+	int32_t rtn = vexDeviceAdiValueGet(device->device_info, adi_port);
+	return_port(smart_port - 1, rtn);
+}
+
+int32_t ext_adi_analog_read_calibrated(uint8_t smart_port, uint8_t adi_port) {
+	transform_adi_port(adi_port);
+	claim_port_i(smart_port - 1, E_DEVICE_ADI);
+	validate_type(device, adi_port, E_ADI_ANALOG_IN);
+	adi_data_s_t* const adi_data = &((adi_data_s_t*)(device->pad))[adi_port];
+	int32_t rtn = (vexDeviceAdiValueGet(device->device_info, adi_port) - (adi_data->analog_data.calib >> 4));
```
nathan-moore, line 161: nit: extra parens

src/devices/vdml_ext_adi.c
```diff
@@ -0,0 +1,416 @@
...
+	adi_data_s_t* const adi_data = &((adi_data_s_t*)(device->pad))[adi_port];
+
+	int32_t pressed = vexDeviceAdiValueGet(device->device_info, adi_port);
+
+	if (!pressed)
+		adi_data->digital_data.was_pressed = false;
+	else if (!adi_data->digital_data.was_pressed) {
+		// Button is currently pressed and was not detected as being pressed during last check
+		adi_data->digital_data.was_pressed = true;
+		return_port(smart_port - 1, true);
```
nathan-moore, line 196: Pre-existing, but this is some fairly poor debouncing. We might want to improve it, but that's for another pr.

version
```diff
@@ -1 +1 @@
-3.2.0
+3.2.1-dirty.98.c265df5
```
nathan-moore, line 1: nit: revert

include/pros/ext_adi.hpp
```diff
@@ -0,0 +1,528 @@
...
+	 * failed, setting errno.
+	 */
+	std::int32_t set_value(std::int32_t value) const;
+
+	protected:
+	EXT_ADIPort(void);
+	std::uint8_t _smart_port;
+	std::uint8_t _adi_port;
+	std::uint32_t _port;
+
```
nathan-moore, line 90: nit: extra space

nathan-moore (commented):
I haven't reviewed in too much depth, but my main concern is the large amount of duplicate code this introduces.

The other thing that strikes me is that changing between an internal ADI port and external code wise is a lot of work. You need to update all types and functions that reference it. Preferably, you'd only have to update one place.

include/pros/ext_adi.h
```diff
@@ -0,0 +1,720 @@
...
+ *        UP
+ * \param adi_port_bottom
+ *        The "bottom" wire from the encoder sensor
+ * \param reverse
+ *        If "true", the sensor will count in the opposite direction
+ *
+ * \return An adi_encoder_t object to be stored and used for later calls to
+ * encoder functions
+ */
+ext_adi_encoder_t ext_adi_encoder_init(uint8_t smart_port, uint8_t adi_port_top, uint8_t adi_port_bottom, const bool reverse);
```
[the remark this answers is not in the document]
HotelCalifornia: can you elaborate on what you mean here

include/pros/ext_adi.hpp
```diff
@@ -0,0 +1,528 @@
...
+#ifndef _PROS_EXT_ADI_HPP_
+#define _PROS_EXT_ADI_HPP_
+
+#include "pros/ext_adi.h"
+
+#include <cstdint>
+
+namespace pros {
+class EXT_ADIPort {
```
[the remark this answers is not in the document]
HotelCalifornia: I really don't like the name of these classes. underscores in class names are verboten

include/pros/ext_adi.h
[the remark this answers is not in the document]
nathan-moore: I was thinking of something like
```C
adi_encoder_t encoder = ext_adi_encoder_init(1, 2, 3, false);
adi_encoder_reset(encoder); //pre-existing function
```

include/pros/ext_adi.h
```diff
@@ -0,0 +1,720 @@
...
+ * \return The difference of the sensor value from its calibrated default from
+ * -16384 to 16384
+ */
+int32_t ext_adi_analog_read_calibrated_HR(uint8_t smart_port, uint8_t adi_port);
+
+/**
+ * Gets the digital value (1 or 0) of a port configured as a digital input.
+ *
+ * If the port is configured as some other mode, the digital value which
+ * reflects the current state of the poprt is returned, which may or may not
```
nathan-moore, line 301:
```suggestion
 * reflects the current state of the port is returned, which may or may not
```

include/pros/ext_adi.h
```diff
@@ -0,0 +1,720 @@
...
+ * \return True if the pin is HIGH, or false if it is LOW
+ */
+int32_t ext_adi_digital_read(uint8_t smart_port, uint8_t adi_port);
+
+/**
+ * Gets a rising-edge case for a digital button press.
+ *
+ * This function is not thread-safe.
+ * Multiple tasks polling a single button may return different results under the
+ * same circumstances, so only one task should call this function for any given
```
nathan-moore, line 324: You may want to specify that this isn't thread safe with adi_digital_get_new_press as well.

include/pros/ext_adi.h
```diff
@@ -0,0 +1,720 @@
...
+ * Sets the encoder value to zero.
+ *
+ * It is safe to use this method while an encoder is enabled. It is not
+ * necessary to call this method before stopping or starting an encoder.
+ *
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * ENXIO - The given value is not within the range of ADI Ports
+ * EADDRINUSE - The port is not configured as an encoder
+
```
nathan-moore, line 504: nit: extra space here. Not sure if doxygen will like it.

src/devices/vdml_adi.c
```diff
@@ -10,371 +10,110 @@
...
-		return PROS_ERR;                            \
-	}
-
-adi_port_config_e_t adi_port_get_config(uint8_t port) {
-	transform_adi_port(port);
-	claim_port_i(INTERNAL_ADI_PORT, E_DEVICE_ADI);
-	adi_port_config_e_t rtn = (adi_port_config_e_t)vexDeviceAdiPortConfigGet(device->device_info, port);
-	return_port(INTERNAL_ADI_PORT, rtn);
+adi_port_config_e_t adi_port_get_config(int32_t port) {
+	return ext_adi_port_get_config(INTERNAL_ADI_PORT, port);
```
nathan-moore, line 18: You expanded port to 32 bits, but you are using INTERNAL_ADI_PORT here instead of seeing if the upper bits contain that data. What are you planning on doing with this api?

Note that the upper bits being 0 will have to map to INTERNAL_ADI_PORT, instead of 22.

src/devices/vdml_ext_adi.c
```diff
@@ -0,0 +1,405 @@
...
+	int32_t rtn = (vexDeviceAdiValueGet(device->device_info, adi_port) - (adi_data->analog_data.calib >> 4));
+	return_port(smart_port - 1, rtn);
+}
+
+int32_t ext_adi_analog_read_calibrated_HR(uint8_t smart_port, uint8_t adi_port) {
+	transform_adi_port(adi_port);
+	claim_port_i(smart_port - 1, E_DEVICE_ADI);
+	validate_type(device, adi_port, E_ADI_ANALOG_IN);
+	adi_data_s_t* const adi_data = &((adi_data_s_t*)(device->pad))[adi_port];
+	int32_t rtn = ((vexDeviceAdiValueGet(device->device_info, adi_port) << 4) - adi_data->analog_data.calib);
```
nathan-moore, line 168: nit: extra parens

nathan-moore (changes requested):
Looks better, though I think we're both confused on what the api is supposed to be.

src/devices/vdml_adi.c
mooreBrendan: This is intended to be for users of only the internal adi api, so that users will not see any change in the original api functionality.  It only takes the adi port (a-h) so there is no need to map 0 to 22.

nathan-moore: Then why the change moving port from 8 bits to 32?

mooreBrendan: I'm currently working on changing smart_port and adi_port to be stored as uint8_t, so that shouldn't be an issue

include/pros/ext_adi.h
mooreBrendan: I don't think this is necessary.  The descriptions of both api’s already say that the function is not thread-safe with two threads referring to the same button at the same time.  I think that the current descriptions should be clear enough that the two functions do not work together.  Additionally, the two c api's are not intended to ever refer to the same device.  The external adi should only refer to ports 1-21 and the internal adi should only refer to port 22.  The external adi api can refer to the internal adi, but that is not its intended purpose for end-users.  Anyone who can understand that they can use the external adi api to refer to the internal should also be able to understand from the current warning that these two functions would conflict.

src/devices/vdml_ext_adi.c
HotelCalifornia: if vex doesn't debounce on their side then we posibly end up with a random value anyway, but there's not anything we can do about that really. that said, i think this is mostly useful for getting a latching button

HotelCalifornia (commented):
i mentioned this briefly in one of my comments but we have the opportunity to dedupe a lot of our C++ ctors here by using default parameters. the one (potentially big) problem is that we'd have to rearrange arguments such that those with default values go at the end, which would mean swapping smart_port and adi_port in most cases, and i'm not really sure how i feel about that from a consistency standpoint.

that said, even if we don't go with that full consolidation, there's some amount of duplicated code we can get rid of by using constructor delegation

include/pros/adi.hpp
```diff
@@ -32,13 +32,29 @@ class ADIPort {
...
+	 * ENXIO - The given value is not within the range of ADI Ports
+	 *
+	 * \param smart_port
+	 *        The smart port number (from 1-22) to configure
+	 * \param adi_port
+	 *        The ADI port number (from 1-8, 'a'-'h', 'A'-'H') to configure
+	 * \param type
+	 *        The configuration type for the port
+	 */
+	ADIPort(std::uint8_t smart_port, std::uint8_t adi_port, adi_port_config_e_t type= E_ADI_TYPE_UNDEFINED);
```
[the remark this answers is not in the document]
mooreBrendan: if we make this change, should the ext adi, be changed to match (adi, smart) as well?

include/pros/ext_adi.h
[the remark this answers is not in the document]
nathan-moore: Yea, that's fair. It's just what I posted works fine if you know that you can cast between the types fine, so it's just _odd_

include/pros/adi.hpp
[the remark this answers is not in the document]
WillXuCodes: Yes, we should to maintain api consistency and reduce user confusion.

include/pros/adi.hpp
```diff
@@ -446,7 +455,7 @@ class ADIUltrasonic : private ADIPort {
 	 * \return The distance to the nearest object in m^-4 (10000 indicates 1
 	 * meter), measured from the sensor's mounting points.
 	 */
-	using ADIPort::get_value;
+	std::int32_t get_value(void) const;
```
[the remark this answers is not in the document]
mooreBrendan: it uses the ultrasonic's get_value function, instead of the default adi port get_value function which didn't read correctly for the ultrasonic

include/pros/adi.hpp
```diff
@@ -466,13 +581,37 @@ class ADIGyro : private ADIPort {
...
 	 *        The ADI port to initialize as a gyro (from 1-8, 'a'-'h', 'A'-'H')
 	 * \param multiplier
 	 *        A scalar value that will be multiplied by the gyro heading value
 	 *        supplied by the ADI
 	 */
-	ADIGyro(std::uint8_t port, double multiplier = 1);
+	ADIGyro(std::uint8_t adi_port, double multiplier = 1);
+	
+	/**
+	 * Initializes a gyroscope on the given port pf an adi expander. If the given
```
HotelCalifornia, line 593: "port pf"?

include/pros/ext_adi.h
```diff
@@ -0,0 +1,641 @@
...
+/**                                                                          **/
+/**       These functions allow for interaction with any ADI port type       **/
+/******************************************************************************/
+
+/**
+ * Gets the configuration for the given ADI port.
+ *
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * ENXIO - The given value is not within the range of ADI Ports.
```
HotelCalifornia, line 49: ENXIO can be set in these functions either
- when `adi_port` is invalid, or
- when `smart_port` is invalid

please change this text (for all of these functions) to clarify that

include/pros/ext_adi.h
```diff
@@ -0,0 +1,641 @@
...
+ *
+ * \param adi_port
+ *	      The ADI port number (from 1-8, 'a'-'h', 'A'-'H') for which to return
```
HotelCalifornia, line 52:
```suggestion
 *        The ADI port number (from 1-8, 'a'-'h', 'A'-'H') for which to return
```

This affects all (or at least many) of the comments in this file

include/pros/ext_adi.h
```diff
@@ -0,0 +1,641 @@
...
+   *
+   */
+#define get_smart_port(ports, smart_port) {\
+	uint32_t uport = (uint32_t)ports; \
+	smart_port = uport & SMART_PORT_MASK; \
+	}
+
+static inline uint32_t merge_adi_ports(uint8_t adi_port, uint8_t smart_port) {
+	return (adi_port << SMART_PORT_BITS) | smart_port;
+}
```
HotelCalifornia, line 633: do these really need to be in the public header? end users should not be using them

src/devices/vdml_adi.cpp
```diff
@@ -16,89 +16,134 @@
...
+	ext_adi_port_set_config(_adi_port, _smart_port, type);
+}
+
+ADIPort::ADIPort(ext_adi_port_pair_t port, adi_port_config_e_t type) : _adi_port(port.first), _smart_port(port.second) {
+	ext_adi_port_set_config(_adi_port, _smart_port, type);
 }
 
 ADIPort::ADIPort(void) {
 	// for use by derived classes like ADIEncoder
 }
```
HotelCalifornia, line 29: hm this should probably be defaulted in the header

src/devices/vdml_adi.cpp
```diff
@@ -16,89 +16,134 @@
...
 
 std::int32_t ADIEncoder::get_value(void) const {
-	return adi_encoder_get(_port);
+	return adi_encoder_get(merge_adi_ports(_adi_port,_smart_port));
+}
+
+ADIUltrasonic::ADIUltrasonic(std::uint8_t adi_port_ping, std::uint8_t adi_port_echo) {
+	std::int32_t _port = ext_adi_ultrasonic_init(adi_port_ping, adi_port_echo, INTERNAL_ADI_PORT);
+	get_ports(_port,_smart_port,_adi_port);
```
HotelCalifornia, line 119:
```suggestion
	get_ports(_port, _smart_port, _adi_port);
```

src/devices/vdml_adi.cpp
```diff
@@ -16,89 +16,134 @@
...
+}
+
+ADIUltrasonic::ADIUltrasonic(ext_adi_port_tuple_t port) {
+	std::int32_t _port = ext_adi_ultrasonic_init(std::get<0>(port), std::get<1>(port), std::get<2>(port));
+	get_ports(_port,_smart_port,_adi_port);
```
HotelCalifornia, line 124:
```suggestion
	get_ports(_port, _smart_port, _adi_port);
```

src/devices/vdml_adi.cpp
```diff
@@ -16,89 +16,134 @@
...
 }
 
 std::int32_t ADIEncoder::reset(void) const {
-	return adi_encoder_reset(_port);
+	return adi_encoder_reset(merge_adi_ports(_adi_port,_smart_port));
 }
 
 std::int32_t ADIEncoder::get_value(void) const {
-	return adi_encoder_get(_port);
+	return adi_encoder_get(merge_adi_ports(_adi_port,_smart_port));
```
HotelCalifornia, line 114:
```suggestion
	return adi_encoder_get(merge_adi_ports(_adi_port, _smart_port));
```

src/devices/vdml_adi.cpp
```diff
@@ -16,89 +16,134 @@
...
+
+ADIUltrasonic::ADIUltrasonic(ext_adi_port_tuple_t port) {
+	std::int32_t _port = ext_adi_ultrasonic_init(std::get<0>(port), std::get<1>(port), std::get<2>(port));
+	get_ports(_port,_smart_port,_adi_port);
 }
 
-ADIUltrasonic::ADIUltrasonic(std::uint8_t port_ping, std::uint8_t port_echo) {
-	_port = adi_ultrasonic_init(port_ping, port_echo);
+std::int32_t ADIUltrasonic::get_value(void) const {
+	return adi_ultrasonic_get(merge_adi_ports(_adi_port,_smart_port));
```
HotelCalifornia, line 128:
```suggestion
	return adi_ultrasonic_get(merge_adi_ports(_adi_port, _smart_port));
```

src/devices/vdml_adi.cpp
```diff
@@ -16,89 +16,134 @@
...
+}
+
+
+ADIGyro::ADIGyro(std::uint8_t adi_port, double multiplier) {
+	std::int32_t _port = ext_adi_gyro_init(adi_port, INTERNAL_ADI_PORT, multiplier);
+	get_ports(_port,_smart_port,_adi_port);
```
HotelCalifornia, line 134:
```suggestion
	get_ports(_port, _smart_port, _adi_port);
```

src/devices/vdml_adi.cpp
```diff
@@ -16,89 +16,134 @@
...
 }
 
-ADIGyro::ADIGyro(std::uint8_t port, double multiplier) {
-	_port = adi_gyro_init(port, multiplier);
+ADIGyro::ADIGyro(ext_adi_port_pair_t port, double multiplier) {
+	std::int32_t _port = ext_adi_gyro_init(port.first, port.second, multiplier);
+	get_ports(_port,_smart_port,_adi_port);
```
HotelCalifornia, line 139:
```suggestion
	get_ports(_port, _smart_port, _adi_port);
```

src/devices/vdml_adi.cpp
```diff
@@ -16,89 +16,134 @@
...
 }
 
 double ADIGyro::get_value(void) const {
-	return adi_gyro_get(_port);
+	return adi_gyro_get(merge_adi_ports(_adi_port,_smart_port));
```
HotelCalifornia, line 143:
```suggestion
	return adi_gyro_get(merge_adi_ports(_adi_port, _smart_port));
```

src/devices/vdml_adi.cpp
```diff
@@ -16,89 +16,134 @@
...
 }
 
 std::int32_t ADIGyro::reset(void) const {
-	return adi_gyro_reset(_port);
+	return adi_gyro_reset(merge_adi_ports(_adi_port,_smart_port));
```
HotelCalifornia, line 147:
```suggestion
	return adi_gyro_reset(merge_adi_ports(_adi_port, _smart_port));
```

HotelCalifornia:
@nickmertin @TixoRebel @Octogonapus @theol0403 maybe some of you want to review this, especially from the perspective of API consumers

nickmertin (commented):
Overall LGTM

include/pros/ext_adi.h
```diff
@@ -0,0 +1,641 @@
...
+ *
+ * \param adi_port
+ *	      The ADI port number (from 1-8, 'a'-'h', 'A'-'H') for which to return
+ *        the configuration
+ * \param smart_port
+ *        The smart port number that the ADI Expander is in
+ *
+ * \return The ADI configuration for the given port
+ */
+adi_port_config_e_t ext_adi_port_get_config(uint8_t adi_port, uint8_t smart_port);
```
TixoRebel, line 59: Should `adi_port` and `smart_port` be switched for all the external adi functions to stay consistent with all the other smart port devices?

include/pros/ext_adi.h
```diff
@@ -0,0 +1,641 @@
...
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * ENXIO - The given value is not within the range of ADI Ports
+ *
+ * \param adi_port
+ *	      The ADI port number (from 1-8, 'a'-'h', 'A'-'H') to configure
+ * \param smart_port
+ *        The smart port number that the ADI Expander is in
+ * \param mode
+ *        One of INPUT, INPUT_ANALOG, INPUT_FLOATING, OUTPUT, or OUTPUT_OD
```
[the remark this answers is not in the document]
HotelCalifornia: that would probably be the smart move but I think we can save that for a future diff

include/pros/ext_adi.h
HotelCalifornia: since we're no longer using the default parameters method to distinguish between internal and external ADI objects in the C++ API I think it would be smart to go back to `smart_port, adi_port` as the idiom (of course this would need to change in the port pair and port tuple "types" as well)

baylessj (dismissed):
On the whole this looks good to me. I might recommend calling the ext_adi_port_pair_t parameter to the constructors "port_pair" rather than just "port" to be clearer. Especially given that the previous parameter name was "port" and only meant a single port number, that might get a bit confusing.

include/pros/ext_adi.h
theol0403: I agree with this, as it makes more sense from a logical standpoint as well - when defining the path from the brain to the port, you start with the smart port and end with the ADI port.

include/pros/adi.hpp
```diff
@@ -23,6 +23,14 @@
 #include <cstdint>
 
 namespace pros {
+
+/** type definition for the pair of smart port and adi port for the basic adi devices */
+using ext_adi_port_pair_t = std::pair<std::uint8_t, std::uint8_t>;
```
theol0403, line 28: Pair and tuple are never included, which means the only way this compiled was because some other headers pulled them before adi.hpp got included.

include/pros/adi.hpp
```diff
@@ -32,12 +40,27 @@ class ADIPort {
...
+	 * reached:
+	 * ENXIO - The given value is not within the range of ADI Ports
+	 *
+	 * \param port
+	 *        The pair of the ADI port number (from 1-8, 'a'-'h', 'A'-'H') and the
+	 *  	  smart port number (from 1-22) to configure 
+	 * \param type
+	 *        The configuration type for the port
+	 */
+	ADIPort(ext_adi_port_pair_t port, adi_port_config_e_t type = E_ADI_TYPE_UNDEFINED);
```
theol0403, line 63: I'm not sure how much you want to favour correctness over readability, but constructors that can be called with a single argument should be marked as explicit. This prevents an integer from suddenly and mysteriously becoming an ADIPort with undefined type. 
Same goes for all the other C++ classes here.

include/pros/adi.hpp
```diff
@@ -32,12 +40,27 @@ class ADIPort {
...
 
 	virtual ~ADIPort(void) = default;
```
theol0403, line 65: Nit: `void` is redundant, including in all the other methods below. I'm not sure what the project style is but `void` parameters exist in some of the headers (motor.hpp) but not others (imu.hpp). Imo it's more readable to omit.

include/pros/adi.hpp
```diff
@@ -82,7 +105,8 @@ class ADIPort {
 
 	protected:
 	ADIPort(void);
```
theol0403, line 107: Like hotel said this should be defaulted.

src/devices/vdml_adi.c
```diff
@@ -10,371 +10,110 @@
...
-		// Button is currently pressed and was not detected as being pressed during last check
-		adi_data->digital_data.was_pressed = true;
-		return_port(INTERNAL_ADI_PORT, true);
-	}
-
-	return_port(INTERNAL_ADI_PORT, false);
+	return ext_adi_digital_get_new_press(port, INTERNAL_ADI_PORT);
 }
 
 int32_t adi_digital_write(uint8_t port, const bool value) {
```
theol0403, line 57: Why is a non-reference bool marked as const?
Same with `ext_adi.h:L418` and others.

include/pros/ext_adi.h
```diff
@@ -0,0 +1,641 @@
...
+   */
+#define get_ports(ports, smart_port, adi_port) {\
+	uint32_t uport = (uint32_t)ports; \
+	smart_port = uport & SMART_PORT_MASK; \
+	adi_port = uport >> SMART_PORT_BITS;  }
+
+
+  /**
+   * Macro description: Given a port mask, it sets the smart port to the value masked inside the int_32. 
+   *
```
theol0403, line 624:
```suggestion
```
(double check that the suggestion removes the line)

include/pros/ext_adi.h
```diff
@@ -0,0 +1,641 @@
...
+ * failed, setting errno.
+ */
+int32_t ext_adi_gyro_shutdown(ext_adi_gyro_t gyro);
+
+
+#define SMART_PORT_BITS 16
+#define SMART_PORT_MASK ((1 << SMART_PORT_BITS) - 1)
+
+  /**
+   * Macro Description: Given a port mask, it sets the smart port and adi port to the values masked inside the int_32.
```
theol0403, line 614: Minuscule nit: "macro description" is redundant and it's `int32_t` not `int_32`.

src/devices/vdml_ext_adi.c
```diff
@@ -0,0 +1,405 @@
...
+	/*
+	* validate_twowire(port_top,port_bottom)
+	* 
+	* Validates 3 things:
+	* - Ports next to each other
+	* - Both ports are initialized and valid (individually)
+	* 
+	* Returns PROS_ERR if one of these is false.
+	*/
+
```
theol0403, line 84:
```suggestion
```

src/devices/vdml_ext_adi.c
```diff
@@ -0,0 +1,405 @@
...
+#define validate_motor(device, port)                                                                      \
+	adi_port_config_e_t config = (adi_port_config_e_t)vexDeviceAdiPortConfigGet(device->device_info, port); \
+	if (config != E_ADI_LEGACY_PWM && config != E_ADI_LEGACY_SERVO) {                                       \
+		errno = EADDRINUSE;                                                                                   \
+		return PROS_ERR;                                                                                      \
+	}
+
+	/*
+	* validate_twowire(port_top,port_bottom)
```
theol0403, line 76: Remove this little section of the comment, it does not add anything.

theol0403 (approved):
I know it's been mentioned before, but just want to make sure everyone knows that every formatting nit and mistake mentioned here can be automatically detected and corrected by clang-format. It is highly recommended to install a format plugin into your editor, and configure it to format on save. There is a `.clang-format` file in this repository which contains the format guide and all rules that must be obeyed, making sure that everything has consistent format.

Even running clang-format on the latest commit of this PR still catches a few things (mostly to do with whitespace and line length). 

I like the API a lot, I don't have much to say about that other than have smart_port before adi_port in the function arguments.

I just have a few tiny nits, sorry for being so pedantic, ignore my comments as you wish :p.

include/pros/adi.hpp
theol0403: Scratch that - it should be completely removed, as it is not being used.

include/api.h
```diff
@@ -42,13 +42,14 @@
 #define PROS_VERSION_MAJOR 3
 #define PROS_VERSION_MINOR 2
 #define PROS_VERSION_PATCH 1
-#define PROS_VERSION_STRING "3.2.1"
+#define PROS_VERSION_STRING "3.2.1-dirty.128.f89fdf1"
```
theol0403, line 45: This got added back
```suggestion
#define PROS_VERSION_STRING "3.2.1"
```

include/pros/adi.hpp
nickmertin: To clarify: it's always redundant in the parameter list in C++.

HotelCalifornia: i blame @baylessj for this but changing that is out of scope of this PR

HotelCalifornia: i take it back, while we're here and refactoring this file this is a good one to just do

include/pros/ext_adi.h
mooreBrendan: where does smart port value set ENXIO in these functions?

include/pros/adi.hpp
```diff
@@ -446,15 +561,15 @@ class ADIUltrasonic : private ADIPort {
 	 * \return The distance to the nearest object in m^-4 (10000 indicates 1
 	 * meter), measured from the sensor's mounting points.
 	 */
-	using ADIPort::get_value;
+	std::int32_t get_value(void) const;
```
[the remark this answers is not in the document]
mooreBrendan: This doesn’t work because in order for the override command to compile, the original function has to be labeled as virtual.  If you label the original as virtual, than gyro::get_value also tries to override it ev,en though it has a different return type.  So, I don’t think there is a way of doing this, without sacrificing uniformity in function names, and it only adds slight maintainability.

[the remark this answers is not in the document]
theol0403: Humm, this isn't really ideal. Since you have a virtual destructor, you already have the overhead of the vtable, so there is no good reason not to make the methods virtual. They should be virtual anyways, because if you have an `ADIUltrasonic` that is being treated as a `ADIPort`, you want it to properly dispatch to the derived method. 
However, I see the problem with clashing with `ADIGyro`. Unfortunately, how it is right now is not ideal. The `ADIGyro` simply shadows and hides the derived method, which leads to problems with polymorphism and unexpected issues with the return type changing. I was worried it was going to lead to ambiguity, but at least this passes:
```cpp
pros::ADIGyro g('A');
auto x = g.get_value();
static_assert(std::is_same_v<decltype(x), double>);

pros::ADIAnalogIn i('A');
auto j = i.get_value();
static_assert(std::is_same_v<decltype(j), std::int32_t>);
```
But you can see how it is not ideal to have the return type change, it could lead to unexpected conversions and narrowing. 

What would be ideal is to have proper overriding and matching return type. I'm not sure how that would be done though, so I might have to agree with your conclusion. 
Maybe have the base `get_value` override return integers and have a `get_value_HR` for the 10th precision, idk. If you and others don't think that's a good idea that's fine.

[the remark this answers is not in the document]
nathan-moore: The inheritence is private, so I don't think this is an issue.

BTW, my understanding is that most of the cost is in the function call. The vtable data structure has some size overhead, but that wouldn't be the part I worry about. If course, neither matter here.

kunwarsahni01:
Encoder, Motor, Button, Potentiometer, and Ultrasonic, have been retested on the internal and external ports and everything is working properly.

include/pros/adi.hpp
```diff
@@ -94,15 +118,29 @@ class ADIAnalogIn : private ADIPort {
...
+	 * ENXIO - The given value is not within the range of ADI Ports
+	 *
+	 * \param port_pair
+	 *        The pair of the smart port number (from 1-22) and the
+	 * 		  ADI port number (from 1-8, 'a'-'h', 'A'-'H') to configure
+	 *
+	 * \return 1 if the operation was successful or PROS_ERR if the operation
+	 * failed, setting errno.
+	 */
+	ADIAnalogIn(ext_adi_port_pair_t port_pair);
```
theol0403, line 143: What happened to `explicit` here? I noticed you added them in but then took them out in a later commit. Same with others below.

mooreBrendan: if you put explicit on the constructors with port_pair then when you initialize the objects, you have to cast them or create a variable with the correct type first.  When you just do object({{smart, adi}}) it doesn’t automatically convert to ext_adi_port_pair_t, which causes the program to not work.

theol0403: Humm, that's strange. That's not how I remember explicit constructors working, I thought explicit would prevent something from being converted to `ADIAnalogIn` without explicit type. For example `ADIAnalogIn i = 1` does not automatically convert `1` to `ADIAnalogIn` for copy construction. However, I don't think that's the case here, the problem you described is it not being able to create a `ext_adi_port_pair_t` from `{smart, adi}`, and I would think that's the fault of explicit pair constructors. Maybe I'm wrong, but try removing a pair of curly brackets, like `object({smart, adi})` instead of `object({{smart, adi}})`.

mooreBrendan: You were right, changing from {{smart, adi}} to {smart, adi} worked.  I have re-added the explicit

theol0403: Huh, cool. I think what was maybe happening was that it was implicitly converting the inner level of braces into an `ADIAnalogIn` for the copy constructor. It was basically implicitly doing this: `ADIAnalogIn(ADIAnalogIn({smart, adi}))`, so when you made the constructor explicit it broke that.

include/pros/adi.hpp
```diff
@@ -18,42 +18,66 @@
...
+
 class ADIPort {
 	public:
 	/**
 	 * Configures an ADI port to act as a given sensor type.
 	 *
 	 * This function uses the following values of errno when an error state is
 	 * reached:
-	 * ENXIO - The given value is not within the range of ADI Ports
+	 * ENXIO - The given value is not within the range of ADI Ports, or the Smart Port is not within the valid range
```
HotelCalifornia, line 42: 3 things:

- 80-column limit is enforced for comments
- don't capitalize smart port like that
- "the given value is not within the range of ADI ports" both ADI port and smart port values are given as values

```suggestion
	 * ENXIO - Either the ADI port value or the smart port value is not within its
 		   valid range (ADI port: 1-8, 'a'-'h', or 'A'-'H'; smart port: 1-21).
```

HotelCalifornia: needless to say, this should apply for all subsequent occurrences of this line

include/pros/ext_adi.h
```diff
@@ -0,0 +1,613 @@
...
+ * \return 1 if the operation was successful or PROS_ERR if the operation
+ * failed, setting errno.
+ */
+int32_t ext_adi_ultrasonic_shutdown(ext_adi_ultrasonic_t ult);
+
+/**
+ * Reference type for an initialized gyroscope.
+ *
+ * This merely contains the port number for the gyroscope, unlike its use as an
+ * object to store encoder data in PROS 2.
```
HotelCalifornia, line 523: this should probably say "gyro" data
```suggestion
 * object to store gyro data in PROS 2.
```

nathan-moore (dismissed):
Only thing that struck me was that formatting for some of the parm parts of the function comments was off. Nothing too pressing.

HotelCalifornia (approved):
i think this looks good to go

PR #246, opened 2020-09-09
🚸 Make mutexes STL-compliant

### LINKED ISSUE #229: RFC: RAII interface for PROS Mutexes

Buzzec:
#### Summary:
Makes pros::Mutex stl compliant by adhering to [BasicLockable](https://en.cppreference.com/w/cpp/named_req/BasicLockable), [TimedLockable](https://en.cppreference.com/w/cpp/named_req/TimedLockable), and [Lockable](https://en.cppreference.com/w/cpp/named_req/Lockable).

#### Motivation:
Allows for stl locks.

##### References (optional):
Resolves #229

### DIFF
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -153,7 +153,7 @@ class Task {
 	 *        A task handle from task_create() for which to create a pros::Task
 	 *        object.
 	 */
-	void operator=(const task_t in);
+	Task& operator=(task_t in);
 
 	/**
 	 * Removes the Task from the RTOS real time kernel's management. This task
@@ -268,7 +268,7 @@ class Task {
 	 * \return The value of the task's notification value before it is decremented
 	 * or cleared
 	 */
-	std::uint32_t notify_take(bool clear_on_exit, std::uint32_t timeout);
+	static std::uint32_t notify_take(bool clear_on_exit, std::uint32_t timeout);
 
 	/**
 	 * Clears the notification for a task.
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -356,6 +356,11 @@ class Mutex {
 	 */
 	bool give(void);
 
+	/**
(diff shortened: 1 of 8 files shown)


include/pros/rtos.hpp
```diff
@@ -320,6 +321,22 @@ class Task {
 	task_t task;
 };
 
+// STL Clock compliant clock
+class ProsClock {
```
HotelCalifornia, line 325: per discussion in #127 I'm pretty sure there's a way to add newlib stubs in such a way that some of the std::chrono features Just Work

include/pros/rtos.hpp
```diff
@@ -320,6 +321,22 @@ class Task {
...
+	public:
+	typedef uint32_t rep;
+	typedef std::milli period;
+	typedef std::chrono::duration<rep, period> duration;
+	typedef std::chrono::time_point<ProsClock> time_point;
```
HotelCalifornia, line 330: prefer `using` declarations (e.g. `using period = std::milli` [is that a typo?])

HotelCalifornia:
@Buzzec just checking in---is there any further progress on this?

HotelCalifornia:
superseded by #276

PR #251, opened 2020-09-13
🔀 backmerge master onto develop for release 3.3.0

HotelCalifornia:
#### Summary:
Keep develop even with master for further development


nathan-moore (approved):
This should not be squashed. We've done it in the past, which is why some commits are pulling up. Thats also why it said the temp file stubs were added this release, even though they weren't.

PR #255, opened 2020-09-27
🐛 Bugfix/fix abandoned mutexes

nathan-moore:
#### Summary:
Rework vdml mutex taking some. A lot of these changes were focused around removing duplication around return_port. This solved several potential deadlocks where we don't release a mutex after we take it on sad paths. There is also some inclusion of various checks to help catch this case in the future.

#### Motivation:
It fixes some rare cases that can cause mutexes to be abandoned. It should also help us spot these in the future.

#### Test Plan:
A lot of these should only effect sad paths which can be hard to test, we can make sure the various adi parts works on the happy path though. 

- [ ] analog works
- [ ] digital works
- [ ] motor works
- [ ] encoder works
- [ ] ultrasonic works
- [ ] gyro works


src/devices/vdml_ext_adi.c
```diff
@@ -139,68 +159,97 @@ int32_t ext_adi_analog_calibrate(uint8_t smart_port, uint8_t adi_port) {
...
 }
 
 int32_t ext_adi_analog_read(uint8_t smart_port, uint8_t adi_port) {
+	int32_t rtn;
 	transform_adi_port(adi_port);
 	claim_port_i(smart_port - 1, E_DEVICE_ADI);
-	validate_type(device, adi_port, E_ADI_ANALOG_IN);
-	int32_t rtn = vexDeviceAdiValueGet(device->device_info, adi_port);
+	if ((rtn = validate_type(device, adi_port, E_ADI_ANALOG_IN)))
+		goto done;
```
nathan-moore, line 173: I left off the != 0 for compactness. I can add it if people think it's helpful.

This goto cleanup pattern is fairly common, and is a decent way of not duplicating the resource freeing logic.
The other way I can think of doing this is have claim_port return a handle and always call into a helper function, where we can use that to detect what resources we have to free.

include/pros/adi.h
```diff
@@ -585,7 +585,7 @@ int32_t adi_ultrasonic_get(adi_ultrasonic_t ult);
  *        next highest port following port_ping.
  *
  * \return An adi_ultrasonic_t object to be stored and used for later calls to
- * ultrasonic functions
+ * ultrasonic functions or PROS_ERR if an error occurred.
```
HotelCalifornia, line 588: What? How

include/pros/ext_adi.h
```diff
@@ -517,7 +517,7 @@ int32_t ext_adi_ultrasonic_get(ext_adi_ultrasonic_t ult);
  *        next highest port following port_ping.
  *
  * \return An adi_ultrasonic_t object to be stored and used for later calls to
- * ultrasonic functions
+ * ultrasonic functions or PROS_ERR if an error occured.
```
HotelCalifornia, line 520: Ditto

include/pros/adi.h
nathan-moore: It's for the path where there's a configuration mismatch.

nathan-moore:
I backed out of the doc changes as there's a lot of inconsistency pre-existing.

src/devices/vdml_ext_adi.c
WillXuCodes: While I'm personally fine with the goto pattern, for the sake of maintainability and to prevent confusion for future devs I'd say calling return_port() might be more concise. It's not that big of a deal but it's also not in line with how other portions of the kernel are written.

WillXuCodes:
This was fixed in a different PR.

PR #259, opened 2020-10-09
✨ Add Support for Optical Sensor

WillXuCodes:
### Summary:
Setup the C/Cpp API for the OpticalSensor.

### Motivation:
With the rotation sensor being released, PROS should support it since users already have their hands on it.

### Test Plan:

- [x] Test C/Cpp API with optical sensor


kunwarsahni01:
Tested with hardware and all functions appear to be working properly.

include/pros/optical.h
```diff
@@ -0,0 +1,271 @@
...
+namespace c {
+#endif
+
+typedef enum optical_direction_e {
+	NO_GESTURE = 0,
+	UP = 1,
+	DOWN = 2,
+	RIGHT = 3,
+	LEFT = 4,
+} optical_direction_e_t;
```
nathan-moore, line 41: consider making these lower case. All upper case is usually reserved for macros.

include/pros/optical.hpp
```diff
@@ -0,0 +1,232 @@
...
+#define _PROS_OPTICAL_HPP_
+
+#include <stdbool.h>
+#include <cstdint>
+#include "pros/optical.h"
+
+namespace pros {
+class Optical {
+	public:
+  /**
```
nathan-moore, line 29: nit: looks like you need a formatting pass.

include/pros/optical.hpp
```diff
@@ -0,0 +1,232 @@
...
+	std::int32_t disable_gesture();
+
+  /**
+	 * Gets the port number of the Optical Sensor.
+	 *
+	 * \return The Optical Sensor's port number.
+	 */
+	std::uint8_t get_port();
+
+  protected:
```
nathan-moore, line 226: We usually make this private, though I agree it really should be protected. We should probably go through and do a consistency check.

include/pros/optical.hpp
```diff
@@ -0,0 +1,232 @@
...
+	 *
+	 * This function uses the following values of errno when an error state is
+	 * reached:
+	 * ENXIO - The given value is not within the range of V5 ports (1-21).
+	 * ENODEV - The port cannot be configured as an Optical Sensor
+	 *
+	 * \return hue value if the operation was successful or PROS_ERR_F if the operation
+	 * failed, setting errno.
+	 */
+	double get_hue();
```
nathan-moore, line 57: This is a question for @HotelCalifornia. We usually make these virtual, but I'm not entirely sure why.

src/devices/vdml_optical.c
```diff
@@ -0,0 +1,136 @@
...
+	vexDeviceOpticalRgbGet(device->device_info, &rgb);
+	rtn.red = rgb.red;
+	rtn.green = rgb.green;
+	rtn.blue = rgb.blue;
+	rtn.brightness = rgb.brightness;
+	return_port(port - 1, rtn);
+}
+
+#define RAW_ERR_INIT \
+	{ .clear = OPT_GESTURE_ERR, .red = OPT_GESTURE_ERR, .green = OPT_GESTURE_ERR, .blue = OPT_GESTURE_ERR }
```
nathan-moore, line 76: Is there a reason why these aren't pros error?

version
```diff
@@ -1 +1 @@
-3.3.0
+3.3.0
```
HotelCalifornia, line 1: add the newline back so this isn't tracked as a change

src/devices/vdml_optical.c
WillXuCodes: They aren't due to PROS_ERR being int32, and these values are int8 and int16

nathan-moore: They're uin32_t on the pros side?

HotelCalifornia: ah yeah they are for this struct, but not the other one

include/pros/optical.hpp
kunwarsahni01: So should this be private for this PR?

include/pros/optical.hpp
kunwarsahni01: Should this be virtual for this PR?

nathan-moore: Private and non-virtual are the defensive choices. We can do that and see if anyone asks for the need to extend them.

nathan-moore: private and all virtual is what we have done, so just do that for now.

include/pros/optical.h
```diff
@@ -0,0 +1,265 @@
...
+ * 4 = left
+ *
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * ENXIO - The given value is not within the range of V5 ports (1-21).
+ * ENODEV - The port cannot be configured as an Optical Sensor
+ *
+ * \param port
+ *        The V5 Optical Sensor port number from 1-21
+ * \return gesture value if the operation was successful or PROS_ERR if
```
nathan-moore, line 213: > or PROS_ERR if

This looks like ub: https://stackoverflow.com/questions/366017/what-is-the-size-of-an-enum-in-c. Looks like you need to make PROS_ERR part of the enum.

nathan-moore: Practically though, if it works now it's probably fine for now, though it's a simple fix.

include/pros/optical.h
```diff
@@ -0,0 +1,265 @@
...
+ * Enable gesture detection on the sensor
+ *
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * ENXIO - The given value is not within the range of V5 ports (1-21).
+ * ENODEV - The port cannot be configured as an Optical Sensor
+ *
+ * \param port
+ *        The V5 Optical Sensor port number from 1-21
+ */
```
nathan-moore, line 243: return comment?

include/pros/optical.h
```diff
@@ -0,0 +1,265 @@
...
+	uint32_t blue;
+} optical_raw_s_t;
+
+typedef struct optical_gesture_s {
+	uint8_t udata;
+	uint8_t ddata;
+	uint8_t ldata;
+	uint8_t rdata;
+	uint8_t type;
+	uint8_t pad;
```
nathan-moore, line 57: Do you know if this is actual data or making the implicit padding explicit? If you don't, I wouldn't worry about it, as it's used for the low level "raw" apis. Also, if we have it, some docs on the individual fields would be nice.

nathan-moore (approved):
Overall LGTM

include/pros/optical.h
```diff
@@ -0,0 +1,265 @@
...
+	uint16_t count;
+	uint32_t time;
+} optical_gesture_s_t;
+
+/**
+ * Get the detected color hue
+ *
+ * This is not avaliable if gestures are being detected. Hue has a
```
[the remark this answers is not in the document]
kunwarsahni01: It returns 0 when in gesture mode.

include/pros/optical.h
kunwarsahni01: We don't have docs on the individual fields at this time.

nathan-moore (approved):

include/pros/optical.h
```diff
@@ -0,0 +1,273 @@
...
+
+/**
+ * Get the most recent gesture data from the sensor
+ *
+ * Gestures will be cleared after 500mS
+ * 0 = no gesture
+ * 1 = up (towards cable)
+ * 2 = down
+ * 3 = right
+ * 4 = left
```
HotelCalifornia, line 208: this should be documented by the enum itself, not here

HotelCalifornia (commented):
as always, my comments about the C documentation apply to the C++ header also

include/pros/optical.hpp
```diff
@@ -0,0 +1,232 @@
...
+	 * ENODEV - The port cannot be configured as an Optical Sensor
+	 *
+	 * \return The Error code encountered
+	 */
+	virtual std::int32_t disable_gesture();
+
+	/**
+	 * Gets the port number of the Optical Sensor.
+	 *
+	 * \return The Optical Sensor's port number.
```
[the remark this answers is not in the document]
kunwarsahni01: What are you referring to here? There is no `get_port` in the C file...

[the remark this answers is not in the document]
HotelCalifornia: 🤔 i swear there were two comments that both said "the error encountered" or something like that

HotelCalifornia (dismissed):
lgtm

HotelCalifornia (approved):

PR #260, opened 2020-10-15
✨ Add Zeroing and Position Setting Functions for the IMU

### LINKED ISSUE #252: pros::IMU : Add relative value functions such as set_rotation(), set_heading(), reset_rotation(), reset_heading() etc.
#### Expected Behavior:
set IMU sensor's rotation, yaw and heading to zero value; currently IMU retains the value from the prev reading. 

#### Actual Behavior:
I would like to reset IMU sensor's rotation to zero prior to every turn and don't see method to reset_rotation() like robotmesh api. Are there any methods available to reset heading, rotation etc. in the IMU without actually calibrating the IMU prior to each turn or movement...

#### Steps to reproduce:
none

#### System information:
Platform: V5
PROS Kernel Version:  3.2.1

#### Additional Information
none

WillXuCodes:
#### Summary:
Added zeroing and current value setting for readings for the IMU.

#### Motivation:
Requested via issue, might as well add this feature as it would be beneficial for writing autons.

#### References:
closes #252 

#### Test Plan:

- [x] Test Reset Functions (C API)
- [x] Test Reset Functions (C++ API)
- [x] Test Value Set Functions (C API)
- [x] Test Value Set Functions (C ++API)


include/pros/imu.hpp
```diff
@@ -196,7 +201,232 @@ class Imu {
...
+	 * ENXIO - The given value is not within the range of V5 ports (1-21).
+	 * ENODEV - The port cannot be configured as an Inertial Sensor
+	 * EAGAIN - The sensor is still calibrating
+	 *
+	 * \param  port
+	 * 				 The V5 Inertial Sensor port number from 1-21
+	 * \return 1 if the operation was successful or PROS_ERR if the operation
+	 * failed, setting errno.
+	 */
+	virtual std::int32_t reset_rotation();
```
nathan-moore, line 217: Consider making a reset all one. I image that would be a common use.

nathan-moore (commented):
A couple of other thoughts :

This is moving away from the long held tradition of having the cpp object being a view over a singleton kernel side. I don't disagree with this, but you'll have to make sure the docs are clear on this.

A complicated topic here is thread safety.  Pretty much everything is in pros, and these really should follow. While practically they aren't not thread safe, they're in the realm of technically not and should be wrapped in the mutex.  It'll be a bit of work, since imu_get_* takes (what I think is) a non-recursive mutex.

include/pros/imu.hpp
```diff
@@ -24,9 +24,14 @@
...
 	const std::uint8_t _port;
-
 	public:
-	Imu(const std::uint8_t port) : _port(port){};
+	Imu(const std::uint8_t port) : _port(port){
+		this->_heading_offset = 0; 
+		this->_rotation_offset = 0; 
+		this->_pitch_offset = 0; 
+		this->_yaw_offset = 0;
+		this->_roll_offset = 0;
```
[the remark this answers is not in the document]
WillXuCodes: sorry that's my old Java tendencies popping back up :/

WillXuCodes:
> This is moving away from the long held tradition of having the cpp object being a view over a singleton kernel side. I don't disagree with this, but you'll have to make sure the docs are clear on this.

I'm very aware of this, and I made the decision to make this feature cpp only as making the zeroing feature available in the C api would create breaking changes (since I'd have to create some sort of typedef for a new IMU object. (Unless you have other ideas for this).

WillXuCodes:
> A complicated topic here is thread safety. Pretty much everything is in pros, and these really should follow. While practically they aren't not thread safe, they're in the realm of technically not and should be wrapped in the mutex. It'll be a bit of work, since imu_get_* takes (what I think is) a non-recursive mutex.

Going to get on making these thread safe and fixing other requested changes.

HotelCalifornia:
> > This is moving away from the long held tradition of having the cpp object being a view over a singleton kernel side. I don't disagree with this, but you'll have to make sure the docs are clear on this.
> 
> I'm very aware of this, and I made the decision to make this feature cpp only as making the zeroing feature available in the C api would create breaking changes (since I'd have to create some sort of typedef for a new IMU object. (Unless you have other ideas for this).

Yes, when this feature was initially discussed we acknowledged and expected that it would be provided in the C++ API only. As will says, it would be nontrivial to add to the C API and doing so would frankly offer little benefit for the majority of our users

baylessj:
I don't mean to throw the direction of this PR totally off course, but it should be feasible to implement this in the C API first. There is a 128 byte padding adding to the device struct for each of the 21 smart devices that can be used for holding internal state. The motors use this for saving their PID constants and the ADI uses this for storing a variety of state about the connected sensors. The state that is being added to the C++ class for this functionality could be stored instead in the registry padding.

This solution is not nearly as easy or clean as storing state in the C++ class but would have the benefits of sticking to the singleton convention that Nathan mentioned and maintaining feature parity with the C API. It sounds as though the vast majority of competitors have switched to using C++ though, so I'm not sure how important maintaining the C-first API design is anymore. Wanted to clarify that it is possible though.

include/pros/imu.h
```diff
@@ -276,6 +276,198 @@ imu_status_e_t imu_get_status(uint8_t port);
...
+ * EAGAIN - The sensor is still calibrating
+ *
+ * \param  port
+ * 				 The V5 Inertial Sensor port number from 1-21
+ * \return 1 if the operation was successful or PROS_ERR if the operation
+ * failed, setting errno.
+ */
+int32_t imu_set_yaw(uint8_t port, double target);
+
+//Offset Getter Functions (Internal Usage, Not Thread Safe)
```
nathan-moore, line 457: You're adding these in the pr. If they aren't used remove them or put in an internal header.

version
```diff
@@ -1 +1 @@
-3.3.1
+3.3.2-dirty.24.6a3540a
```
baylessj, line 1: you'll want to revert this change to prevent an accidental version bump. We might want to brainstorm some solutions on keeping this from accidentally updating; I got caught by this a number of times too

WillXuCodes: yep, will do when this PR Is done. I have an issue right now where the yaw, roll, and pitch functions aren't working properly so I'm going to work that out in a bit.

WillXuCodes:
Other Noteable Changes in this PR:
-Updated license years in file headers.
-Fixed distance and optical sensor headers with the proper file name.

include/pros/imu.h
```diff
@@ -74,7 +74,7 @@ typedef struct __attribute__((__packed__)) euler_s {
  * \return 1 if the operation was successful or PROS_ERR if the operation
  * failed, setting errno.
  */
-int32_t imu_reset(uint8_t port);
+int32_t imu_calibrate(uint8_t port);
```
nathan-moore, line 77:
```suggestion
int32_t imu_recalibrate(uint8_t port);
```
would be my suggestion since a calibration is already started for you. I'm waffling on this though. I know more then a couple people call this and I don't want to break their code, but the name is kind of bad...

include/pros/imu.hpp
```diff
@@ -206,8 +206,173 @@ class Imu {
 	 *
 	 * \param  port
 	 * 				 The V5 Inertial Sensor port number from 1-21
-	 * \return The raw accelerometer values. If the operation failed, all the
-	 * structure's members are filled with PROS_ERR_F and errno is set.
+	 * \return 1 if the operation was successful or PROS_ERR if the operation
+	 * failed, setting errno.
+	 */
+	virtual std::int32_t reset_rotation() const;
```
nathan-moore, line 212: I know it's existing practice, but I was never a fan of these "constant" functions that mutated things.

include/pros/imu.hpp
```diff
@@ -235,4 +400,4 @@ class Imu {
 };
 }  // namespace pros
 
-#endif
+#endif
```
nathan-moore, line 403: nit: revert removing the new line

src/devices/vdml_imu.c
```diff
@@ -22,7 +23,11 @@
 		return_port(port - 1, err_return);                                         \
 	}
 
-int32_t imu_reset(uint8_t port) {
+typedef struct imu_reset_data { 
+	double heading_offset, rotation_offset, pitch_offset, yaw_offset, roll_offset;
```
nathan-moore, line 27: I believe that these should all be declared on a new line in our style.

src/devices/vdml_imu.c
```diff
@@ -156,3 +184,135 @@ imu_status_e_t imu_get_status(uint8_t port) {
...
+	}
+	v5_smart_device_s_t* device = registry_get_device(port - 1);
+	ERROR_IMU_STILL_CALIBRATING(port, device, PROS_ERR);
+	imu_set_yaw_offset(port - 1, target - imu_get_euler_bl(port).yaw);
+	return_port(port - 1, 1);
+}
+
+//Internal functions for getting IMU offsets (Not thread safe)
+double imu_get_heading_offset(uint8_t port){
+    return ((imu_data_s_t*)registry_get_device(port)->pad)->heading_offset;
```
nathan-moore, line 275: My advice would be to factor getting the imu_data_s_t into a helper, and add an assert that's the struct is smaller then however much memory you get.

include/pros/imu.hpp
```diff
@@ -206,8 +206,173 @@ class Imu {
...
+	 * EAGAIN - The sensor is still calibrating
+	 *
+	 * \param  port
+	 * 				 The V5 Inertial Sensor port number from 1-21
+	 * \return 1 if the operation was successful or PROS_ERR if the operation
+	 * failed, setting errno.
+	 */
+	virtual std::int32_t set_roll(double target) const;
+	/**
+	 *  Get the current offset of the IMU's rotation reading from its true value
```
nathan-moore, line 364: Did you change these docs? This doesn't make sense for get_accel.

include/pros/imu.hpp
nathan-moore: Probably a thing for later though.

src/devices/vdml_imu.c
```diff
@@ -22,7 +23,11 @@
 		return_port(port - 1, err_return);                                         \
 	}
 
-int32_t imu_reset(uint8_t port) {
+typedef struct imu_reset_data {
```
nathan-moore, line 26: Have you double checked that this fits in the registry? You may also want to pack the struct.

include/vdml/imubl.h
```diff
@@ -0,0 +1,210 @@
+/**
+ * \file pros/imubl.h
+ *
+ * IMU Backend Layer header for non-threadsafe functions for reset
+ * functions that interact with the registry.
```
nathan-moore, line 5: I'm curious what others think of this. I personally don't think it's worth it to bother exposing these. We don't really need to expose 3 ways of setting the offset... The getters I don't see a use for.

src/devices/vdml_imu.c
WillXuCodes: Struct packed, I'm pretty sure it fits in the registry (6 doubles at 8 bytes each being 48 bytes total, fitting into 128 bytes on the pad in the registry).

include/vdml/imubl.h
WillXuCodes: in the process of reworking this to not have this at all.

include/pros/imu.h
WillXuCodes: going to change this back to reset so it doesn't break user code, and changing the imu_reset function that resets all values to reset_all (or something better if you have another suggestion).

include/pros/imu.hpp
```diff
@@ -196,7 +196,172 @@ class Imu {
...
+	 * EAGAIN - The sensor is still calibrating
+	 *
+	 * \param  port
+	 * 				 The V5 Inertial Sensor port number from 1-21
+	 * \return 1 if the operation was successful or PROS_ERR if the operation
+	 * failed, setting errno.
+	 */
+	virtual std::int32_t tare_pitch() const;
+	/**
+	 * Resets the current reading of the Inertial Sensor's pitch to zero
```
nathan-moore, line 244:
```suggestion
	 * Resets the current reading of the Inertial Sensor's yaw to zero
```

src/devices/vdml_imu.c
```diff
@@ -97,15 +105,42 @@ euler_s_t imu_get_euler(uint8_t port) {
...
 
 double imu_get_pitch(uint8_t port) {
-	return imu_get_euler(port).pitch;
+	double rtn = PROS_ERR_F;
+	if (!claim_port_try(port - 1, E_DEVICE_IMU)) {
+		return rtn;
+	}
+	euler_s_t euler_values;
+	v5_smart_device_s_t* device;
+	device = registry_get_device(port - 1);
```
nathan-moore, line 114: Why split the definition and where you set it?

src/devices/vdml_imu.c
```diff
@@ -156,3 +191,106 @@ imu_status_e_t imu_get_status(uint8_t port) {
...
+	v5_smart_device_s_t* device = registry_get_device(port - 1);
+	euler_s_t euler_values;
+	vexDeviceImuAttitudeGet(device->device_info, (V5_DeviceImuAttitude*)&euler_values);
+	imu_data_s_t* data = (imu_data_s_t*)device->pad;
+	data->rotation_offset = -vexDeviceImuHeadingGet(device->device_info);
+	data->heading_offset = -vexDeviceImuDegreesGet(device->device_info);
+	data->pitch_offset = -euler_values.pitch;
+	data->roll_offset = -euler_values.roll;
+	data->yaw_offset = -euler_values.yaw;
+	return_port(port - 1, 1);
```
nathan-moore, line 209:
```suggestion
	return_port(port - 1, PROS_SUCCESS);
```

src/devices/vdml_imu.c
```diff
@@ -97,15 +105,42 @@ euler_s_t imu_get_euler(uint8_t port) {
...
 
 double imu_get_roll(uint8_t port) {
-	return imu_get_euler(port).roll;
+	double rtn = PROS_ERR_F;
+	if (!claim_port_try(port - 1, E_DEVICE_IMU)) {
+		return rtn;
+	}
+	euler_s_t euler_values;
+	v5_smart_device_s_t* device;
+	device = registry_get_device(port - 1);
```
nathan-moore, line 127: Same here

src/devices/vdml_imu.c
```diff
@@ -97,15 +105,42 @@ euler_s_t imu_get_euler(uint8_t port) {
...
 
 double imu_get_yaw(uint8_t port) {
-	return imu_get_euler(port).yaw;
+	double rtn = PROS_ERR_F;
+	if (!claim_port_try(port - 1, E_DEVICE_IMU)) {
+		return rtn;
+	}
+	euler_s_t euler_values;
+	v5_smart_device_s_t* device;
+	device = registry_get_device(port - 1);
```
nathan-moore, line 140: And here

nathan-moore (dismissed):

src/devices/vdml_imu.c
nathan-moore: For all of these that return 1

src/devices/vdml_imu.c
WillXuCodes: Honestly, I did it because I saw that somebody else did it in another portion of the code so I thought it was the precedent for the project :/. Was wondering the same thing as you.

src/devices/vdml_imu.c
WillXuCodes: Got it, will fix for previous code as well. Should I also take a look around and fix the other sensors in a seperate PR for this issue?

src/devices/vdml_imu.c
WillXuCodes: Note for future reference: PROS_SUCCESS does not exist but we should add it in PROS 4 for readability's sake.

nathan-moore:
How will this interact with imu_get_quaternion(uint8_t port) and imu_get_euler? Currently the ignore the values.

WillXuCodes:
> How will this interact with imu_get_quaternion and imu_get_euler? Currently they ignore the offsets.

Honestly, I didn't think many people used them other than people who write libraries. I guess we could implement the offsets with those too if needed.

nathan-moore:
> Honestly, I didn't think many people used them other than people who write libraries (who may prefer the raw values instead). I guess we could implement the offsets with those too if needed.

I don't think it's right to introduce a concept such as "raw values". From an outside perspective, there's the angle the imu reports, and that's it. The offset that we keep is an implementation detail. Beside, get_header and such were originally implemented to be a simpler get_euler.

include/api.h
```diff
@@ -41,8 +41,8 @@
 
 #define PROS_VERSION_MAJOR 3
 #define PROS_VERSION_MINOR 3
-#define PROS_VERSION_PATCH 1
-#define PROS_VERSION_STRING "3.3.1"
+#define PROS_VERSION_PATCH 2
+#define PROS_VERSION_STRING "3.3.2-dirty.49.63955a5"
```
HotelCalifornia, line 45: fix this

version
```diff
@@ -1 +1 @@
-3.3.1
+3.3.2-dirty.49.63955a5
```
HotelCalifornia, line 1: fix

src/main.cpp
```diff
@@ -7,13 +7,7 @@
  * "I was pressed!" and nothing.
```
HotelCalifornia, line 7: this file should not be modified

HotelCalifornia (changes requested):

HotelCalifornia:
there's also compilation errors you're going to need to resolve here

include/pros/imu.hpp
```diff
@@ -89,7 +89,7 @@ class Imu {
 	 * Get the Inertial Sensor's heading relative to the initial direction of its
 	 * x-axis
 	 *
-	 * This value is bounded by (-360,360). Clockwise rotations are represented with
+	 * This value is bounded by (0,360). Clockwise rotations are represented with
```
HotelCalifornia, line 92: same as above re: intervals

include/pros/imu.h
```diff
@@ -57,6 +57,10 @@ typedef struct __attribute__((__packed__)) euler_s {
 } euler_s_t;
 
 #define IMU_MINIMUM_DATA_RATE 5
+#define IMU_EULER_LIMIT 180
+#define IMU_HEADING_MAX 360
+
+#define DEGTORAD (M_PI / 180)
```
HotelCalifornia, line 63: unless these are used in this header file somewhere, they should probably get moved to the source file. avoid adding internal values/utilities to the public interface wherever possible

HotelCalifornia: (and yes DEGTORAD may be useful for users but in the interests of separation of concern it should at the very least not be defined in this file, if it's to be given to our users at all)
(which i think it shouldn't be, especially because this is the kind of symbol commonly defined by users anyway and to define it here now might cause conflicts)

include/pros/imu.h
```diff
@@ -465,7 +471,8 @@ int32_t imu_set_pitch(uint8_t port, double target);
 
 /**
  * Sets the current reading of the Inertial Sensor's roll to target value
- *
+ * Will default to +/- 180 if value target +/- 180.
```
HotelCalifornia, line 474: is this line supposed to be the same as the others?

src/devices/vdml_imu.c
```diff
@@ -82,13 +82,24 @@ quaternion_s_t imu_get_quaternion(uint8_t port) {
...
+	double roll = euler.roll + data->roll_offset;
+	double yaw = euler.yaw + data->yaw_offset;
+	double pitch = euler.pitch + data->pitch_offset;
+
+	if(roll > IMU_EULER_LIMIT) roll -= 2 * IMU_EULER_LIMIT;
+	if(roll < -IMU_EULER_LIMIT) roll += 2 * IMU_EULER_LIMIT;
+	if(yaw > IMU_EULER_LIMIT) yaw -= 2 * IMU_EULER_LIMIT;
+	if(yaw < -IMU_EULER_LIMIT) yaw += 2 * IMU_EULER_LIMIT;
+	if(pitch > IMU_EULER_LIMIT) pitch -= 2 * IMU_EULER_LIMIT;
+	if(pitch < -IMU_EULER_LIMIT) pitch += 2 * IMU_EULER_LIMIT;
```
HotelCalifornia, line 95: these should all be `if (...)`

src/devices/vdml_imu.c
```diff
@@ -113,6 +124,12 @@ euler_s_t imu_get_euler(uint8_t port) {
 	rtn.pitch += data->pitch_offset;
 	rtn.yaw += data->yaw_offset;
 	rtn.roll += data->roll_offset;
+	if(rtn.roll > IMU_EULER_LIMIT) rtn.roll -= 2 * IMU_EULER_LIMIT;
+	if(rtn.roll < -IMU_EULER_LIMIT) rtn.roll += 2 * IMU_EULER_LIMIT;
+	if(rtn.yaw > IMU_EULER_LIMIT) rtn.yaw -= 2 * IMU_EULER_LIMIT;
+	if(rtn.yaw < -IMU_EULER_LIMIT) rtn.yaw += 2 * IMU_EULER_LIMIT;
+	if(rtn.pitch > IMU_EULER_LIMIT) rtn.pitch -= 2 * IMU_EULER_LIMIT;
+	if(rtn.pitch < -IMU_EULER_LIMIT) rtn.pitch += 2 * IMU_EULER_LIMIT;
```
HotelCalifornia, line 132: these should all be `if (...)`

src/devices/vdml_imu.c
```diff
@@ -125,6 +142,8 @@ double imu_get_pitch(uint8_t port) {
 	v5_smart_device_s_t* device = registry_get_device(port - 1);
 	vexDeviceImuAttitudeGet(device->device_info, (V5_DeviceImuAttitude*)&euler_values);
 	rtn = euler_values.pitch + ((imu_data_s_t*)registry_get_device(port - 1)->pad)->pitch_offset;
+	if(rtn > IMU_EULER_LIMIT) rtn -= 2 * IMU_EULER_LIMIT;
+	if(rtn < -IMU_EULER_LIMIT) rtn += 2 * IMU_EULER_LIMIT;
```
HotelCalifornia, line 146: `if (...)`

src/devices/vdml_imu.c
```diff
@@ -137,6 +156,8 @@ double imu_get_roll(uint8_t port) {
 	v5_smart_device_s_t* device = registry_get_device(port - 1);
 	vexDeviceImuAttitudeGet(device->device_info, (V5_DeviceImuAttitude*)&euler_values);
 	rtn = euler_values.roll + ((imu_data_s_t*)registry_get_device(port - 1)->pad)->roll_offset;
+	if(rtn > IMU_EULER_LIMIT) rtn -= 2 * IMU_EULER_LIMIT;
+	if(rtn < -IMU_EULER_LIMIT) rtn += 2 * IMU_EULER_LIMIT;
```
HotelCalifornia, line 160: `if (...)`

src/devices/vdml_imu.c
```diff
@@ -149,6 +170,8 @@ double imu_get_yaw(uint8_t port) {
 	v5_smart_device_s_t* device = registry_get_device(port - 1);
 	vexDeviceImuAttitudeGet(device->device_info, (V5_DeviceImuAttitude*)&euler_values);
 	rtn = euler_values.yaw + ((imu_data_s_t*)registry_get_device(port - 1)->pad)->yaw_offset;
+	if(rtn > IMU_EULER_LIMIT) rtn -= 2 * IMU_EULER_LIMIT;
+	if(rtn < -IMU_EULER_LIMIT) rtn += 2 * IMU_EULER_LIMIT;
```
HotelCalifornia, line 174: `if (...)`

src/devices/vdml_imu.c
```diff
@@ -262,7 +285,9 @@ int32_t imu_set_heading(uint8_t port, double target){
 	euler_s_t euler_values;
 	vexDeviceImuAttitudeGet(device->device_info, (V5_DeviceImuAttitude*)&euler_values);
 	imu_data_s_t* data = (imu_data_s_t*)device->pad;
-	data->heading_offset = (target - vexDeviceImuDegreesGet(device->device_info)) % 360;
+	if(target > IMU_HEADING_MAX) target = IMU_HEADING_MAX;
+	if(target < 0) target = 0;
```
HotelCalifornia, line 289: you know the drill

src/devices/vdml_imu.c
```diff
@@ -275,7 +300,9 @@ int32_t imu_set_pitch(uint8_t port, double target){
 	euler_s_t euler_values;
 	vexDeviceImuAttitudeGet(device->device_info, (V5_DeviceImuAttitude*)&euler_values);
 	imu_data_s_t* data = (imu_data_s_t*)device->pad;
-	data->pitch_offset = (target - euler_values.pitch) % 90;
+	if(target > IMU_EULER_LIMIT) target = IMU_EULER_LIMIT;
+	if(target < -IMU_EULER_LIMIT) target = -IMU_EULER_LIMIT;
```
HotelCalifornia, line 304: here too

src/devices/vdml_imu.c
```diff
@@ -288,7 +315,9 @@ int32_t imu_set_roll(uint8_t port, double target){
 	euler_s_t euler_values;
 	vexDeviceImuAttitudeGet(device->device_info, (V5_DeviceImuAttitude*)&euler_values);
 	imu_data_s_t* data = (imu_data_s_t*)device->pad;
-	data->roll_offset = (target - euler_values.roll) % 90;
+	if(target > IMU_EULER_LIMIT) target = IMU_EULER_LIMIT;
+	if(target < -IMU_EULER_LIMIT) target = -IMU_EULER_LIMIT;
```
HotelCalifornia, line 319: and here

src/devices/vdml_imu.c
```diff
@@ -301,7 +330,9 @@ int32_t imu_set_yaw(uint8_t port, double target){
 	euler_s_t euler_values;
 	vexDeviceImuAttitudeGet(device->device_info, (V5_DeviceImuAttitude*)&euler_values);
 	imu_data_s_t* data = (imu_data_s_t*)device->pad;
-	data->yaw_offset = (target - euler_values.yaw) % 90;
+	data->yaw_offset = target - euler_values.yaw;
+	if(target > IMU_EULER_LIMIT) target = IMU_EULER_LIMIT;
+	if(target < -IMU_EULER_LIMIT) target = -IMU_EULER_LIMIT;
```
HotelCalifornia, line 335: here

src/devices/vdml_imu.c
```diff
@@ -313,8 +344,14 @@ int32_t imu_set_euler(uint8_t port, euler_s_t target){
...
 	imu_data_s_t* data = (imu_data_s_t*)device->pad;
-	data->pitch_offset = (target.pitch - euler_values.pitch) % 90;
-	data->roll_offset = (target.roll - euler_values.roll) % 90;
-	data->yaw_offset = (target.yaw - euler_values.yaw) % 90;
+	if(target.pitch > IMU_EULER_LIMIT) target.pitch = IMU_EULER_LIMIT;
+	if(target.pitch < -IMU_EULER_LIMIT) target.pitch = -IMU_EULER_LIMIT;
+	if(target.yaw > IMU_EULER_LIMIT) target.yaw = IMU_EULER_LIMIT;
+	if(target.yaw < -IMU_EULER_LIMIT) target.yaw = -IMU_EULER_LIMIT;
+	if(target.roll > IMU_EULER_LIMIT) target.roll = IMU_EULER_LIMIT;
+	if(target.roll < -IMU_EULER_LIMIT) target.roll = -IMU_EULER_LIMIT;
```
HotelCalifornia, line 352: all of these

HotelCalifornia (commented):
some minor style things but other than those this looks good to me

include/pros/imu.h
WillXuCodes: yes, I'm trying to say that if the target value is above or below +180 and -180 respectively, it will default to +180 or -180.

HotelCalifornia: right, but the wording here is inexplicably different from the others

WillXuCodes: All three euler values have their bounds at -180 and 180, and each setter function has the same feature.

HotelCalifornia (approved):
lgtm

PR #276, opened 2021-01-23
🚸 Make mutexes STL-compliant

### LINKED ISSUE #229: RFC: RAII interface for PROS Mutexes

HotelCalifornia:
#### Summary:

Makes pros::Mutex stl compliant by adhering to [BasicLockable](https://en.cppreference.com/w/cpp/named_req/BasicLockable), [TimedLockable](https://en.cppreference.com/w/cpp/named_req/TimedLockable), and [Lockable](https://en.cppreference.com/w/cpp/named_req/Lockable).

#### Motivation:

Allows the use of various STL locks with pros::Mutex

##### References (optional):

supersedes #246, which has had no response from the contributor for months.
resolves #229 

#### Test Plan:

```cpp
// adapted from example found at https://en.cppreference.com/w/cpp/thread/lock_guard
#include "pros/apix.h"
#include <mutex>

int g_i = 0;
pros::Mutex g_i_mutex;  // protects g_i
 
void safe_increment(void* id) {
	const std::lock_guard<pros::Mutex> lock(g_i_mutex);
	++g_i;
 
	std::cout << static_cast<const char*>(id) << ": " << g_i << '\n';
 
	// g_i_mutex is automatically released when lock
	// goes out of scope
}

void join(pros::Task& t) {
	pros::c::task_notify_when_deleting(t, nullptr, 1, pros::E_NOTIFY_ACTION_INCR);

	while (true) {
		if (pros::Task::notify_take(true, TIMEOUT_MAX)) {
			return;
		}
	}
}

void opcontrol() {
	std::cout << "main: " << g_i << '\n';
 
	pros::Task t1(safe_increment, (void*)"t1");
	pros::Task t2(safe_increment, (void*)"t2");

	join(t1);
	join(t2);
	
	std::cout << "main: " << g_i << '\n';
}
```


HotelCalifornia:
and if any of our fine community members want to test it, here's the latest artifact
[kernel@3.3.2-pr276.1825.zip](https://github.com/purduesigbots/pros/files/5859583/kernel%403.3.2-pr276.1825.zip)

src/rtos/rtos.cpp
```diff
@@ -99,11 +99,31 @@ using namespace pros::c;
...
+  }
+
+  void Mutex::unlock(void) {
+	  give();
+  }
+
+  bool Mutex::try_lock(void) {
+	  try_lock_for(std::chrono::duration<long, std::ratio<1>>{1});
+	  try_lock_until(Clock::time_point{std::chrono::duration<long, std::ratio<1>>{1}});
+	  return take(0);
```
nathan-moore, line 127: Can you explain this logic? Why is the mutex trying to be taken 3 times?

include/pros/rtos.hpp
```diff
@@ -320,7 +321,27 @@ class Task {
 	task_t task;
 };
 
+// STL Clock compliant clock
+struct Clock {
```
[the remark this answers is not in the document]
HotelCalifornia: yes

include/pros/rtos.hpp
```diff
@@ -356,8 +377,41 @@ class Mutex {
 	 */
 	bool give(void);
 
-	private:
-	std::shared_ptr<std::remove_pointer_t<mutex_t>> mutex;
+	/**
+	 * BasicLockable compliance lock function https://en.cppreference.com/w/cpp/named_req/BasicLockable
+	 * Do not use directly, use pros::Mutex::take instead
```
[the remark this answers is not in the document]
HotelCalifornia: well i assume it's a sort of mix between the advice cpreference gives ("in general don't interact directly with a mutex, use a lock around it") and telling people who don't use STL locks which function they should be using. but beyond that conjecture i have no idea

src/rtos/rtos.cpp
HotelCalifornia: i have a feeling it's an attempt to artificially recreate the behaviour described in

> This function is allowed to fail spuriously and return false even if the mutex is not currently locked by any other thread.

which [someone explains](https://stackoverflow.com/a/48305054) probably really means one of two things:

1. the mutex is already locked elsewhere...
2. some platform-specific feature interrupts or prevents the locking attempt, and control is returned to the caller who can decide whether to retry.

in which case i think we really can just cut this down to `take(0)`... input appreciated

WillXuCodes: Original source of the fail spuriously: https://en.cppreference.com/w/cpp/thread/mutex/try_lock

Another thing to note about this, is that the same source says that the function "Returns immediately.", so we could use that to justify not waiting for those attempts and just cutting it down to take(0);

HotelCalifornia: agreed. I think the default behaviour of a freeRTOS mutex matches the desired pretty closely anyway (i.e. if something prevents the locking attempt it will fail properly... probably)

HotelCalifornia:
latest artifact 
[kernel@3.3.2-pr276.1827.zip](https://github.com/purduesigbots/pros/files/5861086/kernel%403.3.2-pr276.1827.zip)

WillXuCodes:
```C++
#include "pros/apix.h"
#include <mutex>
#include "main.h"

int i = 0;
pros::Mutex inc_mutex;
void inc() {
	const std::lock_guard<pros::Mutex> lock(inc_mutex);
	i++;
}

void opcontrol() {
	
	uint32_t start = pros::millis();

      for(int j = 0; j < 6000; j++){
		  pros::Task t1(inc);
		  pros::Task t2(inc);
		  pros::delay(1);
      }
      pros::delay(10000);

	std::cout << "i value: " << i << std::endl;
	std::cout << "Elapsed Time: " << pros::millis() - start << std::endl;

	pros::lcd::initialize();
	pros::lcd::set_text(1, "Hello PROS User!");

	pros::lcd::register_btn1_cb(on_center_button);
}
```

Test code that I used, the output was consistently :

```
i value: 12000
Elapsed Time: 16000
```

WillXuCodes (approved):
(See comment above, if it wasn't working the value would be somewhere < 12000 between compilations)

WillXuCodes (approved):
Works as expected after formatting fixes

PR #277, opened 2021-01-24
🚸Change pros-cli-v5 to pros-cli

WillXuCodes:
Changed instance of pros-cli-v5 to pros-cli to reflect changes in cli repo.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -25,7 +25,7 @@ jobs:
   - task: PipAuthenticate@1
     inputs:
       artifactFeeds: 'pros-cli'
-  - bash: pip install --verbose pros-cli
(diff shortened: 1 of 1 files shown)


HotelCalifornia:
for some reason the wheel in our artifact feed is not installing the `prosv5` console script as it does when i install it locally

HotelCalifornia:
Looks like there's an issue with the pip dependency resolver that prevents
- us from using version ranges to specify the package
- pip from accepting prerelease/release candidates as satisfying version ranges

(tracking at pypa/pip#9619).

that said, since we released 3.2.0 to pypi we have sidestepped the issues. in the future we will want to make sure this action only uses versions ≥ 3, but we can safely merge this for now

HotelCalifornia (approved):
lgtm

PR #281, opened 2021-02-28
 ✨ Add microsecond timing

### LINKED ISSUE #280: Sub-millisecond timing
#### Expected Behavior:
There should be an API (and the requisite backend machinery) to get sub-millisecond timing. This may involve reducing the quantum below 1 ms, which may or may not also yield scheduling efficiency benefits.

#### System information:
Platform: V5
PROS Kernel Version: all

TixoRebel:
#### Summary:
Adds `pros::micros()` to return the time since program start in microseconds.

I noticed `pros::millis()` didn't use `vexSystemTimeGet()` but I'm not familiar enough with FreeRTOS to know what it's doing and if that'd be better then wrapping `vexSystemHighResTimeGet()`.

#### Motivation:
There is currently no sub-millisecond timing. 

##### References:
Closes #280

#### Test Plan:

- [x] Call `pros::micros()` and ensure it returns the time since program start in microseconds

### DIFF
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -368,6 +368,13 @@ class Mutex {
 using pros::c::millis;
 
 /**
+ * @brief Gets the number of microseconds since PROS initialized.
+ * 
+ * \return The number of microseconds since PROS initialized
+ */
+using pros::c::micros;
+
+/**
  * Delays a task for a given number of milliseconds.
  *
  * This is not the best method to have a task execute code at predefined
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -368,7 +368,7 @@ class Mutex {
 using pros::c::millis;
 
 /**
- * @brief Gets the number of microseconds since PROS initialized.
+ * Gets the number of microseconds since PROS initialized.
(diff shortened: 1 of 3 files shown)


nickmertin:
Duplicate of #280 ?

HotelCalifornia:
@kunwarsahni01 @WillXuCodes if one of youse can test this quickly we can retarget it onto the release branch and ship this out with that

HotelCalifornia (changes requested):
https://github.com/purduesigbots/pros/blob/a07944b29a364d0f0a450c97bd10c667e70762a7/include/pros/rtos.hpp#L368

you should bring this to the pros namespace like we do with `millis`

include/pros/rtos.hpp
```diff
@@ -367,6 +367,13 @@ class Mutex {
  */
 using pros::c::millis;
 
+/**
+ * @brief Gets the number of microseconds since PROS initialized.
```
HotelCalifornia, line 371:
```suggestion
 * Gets the number of microseconds since PROS initialized.
```

HotelCalifornia (approved):
lgtm

PR #282, opened 2021-02-28
✨ Allow user code to query registry status

TixoRebel:
#### Summary:
Move the decelerations of `registry_get_bound_type` and `registry_get_plugged_type` from `registry.h` to `apix.h` to expose them to users.

#### Motivation:
Currently there is no way for user code to query the bound and plugged status of each port. This is useful to make sure devices are ready to use before they are accessed.

#### Test Plan:

- [x] Ensure that registry_get_bound_type and registry_get_plugged_type return the expected values

### DIFF
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -417,6 +417,36 @@ int registry_bind_port(uint8_t port, v5_device_e_t device_type);
  */
 int registry_unbind_port(uint8_t port);
 
+/*
+ * Returns the type of device registered to the port.
+ *
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * ENXIO - The given value is not within the range of V5 ports (1-21).
+ *
+ * \param port
+ *        The V5 port number from 1-21
+ *
+ * \return The type of device that is registered into the port (NOT what is
+ * plugged in)
+ */
+v5_device_e_t registry_get_bound_type(uint8_t port);
+
+/*
+ * Returns the type of the device plugged into the port.
+ *
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * ENXIO - The given value is not within the range of V5 ports (1-21).
+ *
+ * \param port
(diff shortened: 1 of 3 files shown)


WillXuCodes:
LGTM, I don't see any downside to doing this and feel like this would be beneficial for debugging or autogenerating port diagrams. Would be kind of interesting to see what users will use this for! (other than your guys's rust stuff lol)

WillXuCodes (approved):

nathan-moore (approved):

WillXuCodes:
All we're waiting for on this one is for somebody to test it with hardware.

WillXuCodes:
So it seems that our initial documentation was slightly off, the port parameter is zero indexed, could you make the documentation comment say something about this? Thanks.

WillXuCodes:
Other than that everything seems to be working.

TixoRebel:
I see that the docs for `registry_bind_port` are also incorrect so I'll correct those too. Right now `registry_unbind_port` does actually take 1 based indexes though it's not used anywhere in the kernel, should I update it to be 0 indexed to be consistent the rest of the registry functions?

WillXuCodes:
> I see that the docs for `registry_bind_port` are also incorrect so I'll correct those too. Right now `registry_unbind_port` does actually take 1 based indexes though it's not used anywhere in the kernel, should I update it to be 0 indexed to be consistent with the rest of the registry functions?

Yep, that would be greatly appreciated. I'm not sure if any others disagree w/ me but I personally think that registry_unbind_port isn’t really something the average user would use... and we might as well keep it consistent to the rest of the API with the port number.
The only downside that it’s a breaking change to whoever does use it

WillXuCodes:
If nobody disagrees with me within 24 H I'll merge it after you change it.

WillXuCodes (approved):
LGTM

PR #286, opened 2021-04-19
✨Added Delay to IMU Reset

belericant:
#### Summary:
The calibration flag is not set immediately after a call to `imu_reset` necessitating user code to add a delay before polling the status of the IMU. By adding a 20 ms delay after calling `vexDeviceImuReset` in `imu_reset`, the calibration flag is correctly set once the function returns.

#### Motivation:
`imu_get_status` should reflect that the IMU is calibrating immediately upon the return of `imu_reset`.

#### Test Plan:
- [x] find the delay necessary for the calibration flag to be set
- [x] make sure `imu_reset` and `imu_get_status` still work

The delay between `vexDeviceImuReset` and the calibration flag being set was found by running 

```c
void initialize() {
  imu_reset(IMU_PORT);
  uint32_t start_time = millis();
  while (!(imu_get_status(IMU_PORT) & E_IMU_STATUS_CALIBRATING)) {

  }
  uint32_t delta = millis() - start_time;

  FILE* file = fopen("/usd/logging.csv", "a");
  char str[100];
  sprintf(str, "%d, ", delta);
  fputs(str, file);
  fclose(file);
}
```

The resulting log file after running the program 100 times showed that the delay was a consistent 12 ms.

### DIFF
--- a/include/pros/imu.h
+++ b/include/pros/imu.h
@@ -61,7 +61,8 @@ typedef struct __attribute__((__packed__)) euler_s {
 /**
  * Calibrate IMU
  *
- * This takes approximately 2 seconds, and is a non-blocking operation.
+ * Calibration takes approximately 2 seconds, but this function only blocks
+ * until the IMU status flag is set properly to E_IMU_STATUS_CALIBRATING.
  *
  * This function uses the following values of errno when an error state is
  * reached:
--- a/include/pros/imu.hpp
+++ b/include/pros/imu.hpp
@@ -31,7 +31,8 @@ class Imu {
 	/**
 	 * Calibrate IMU
 	 *
-	 * This takes approximately 2 seconds, and is a non-blocking operation.
+	 * Calibration takes approximately 2 seconds, but this function only blocks
+ 	 * until the IMU status flag is set properly to E_IMU_STATUS_CALIBRATING.
 	 *
 	 * This function uses the following values of errno when an error state is
 	 * reached:
--- a/src/devices/vdml_imu.c
+++ b/src/devices/vdml_imu.c
@@ -40,7 +40,7 @@ int32_t imu_reset(uint8_t port) {
 	ERROR_IMU_STILL_CALIBRATING(port, device, PROS_ERR);
 	vexDeviceImuReset(device->device_info);
 	// delay for vexos to set calibration flag
-	delay(20);
+	while(!(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING)) delay(1);
 	return_port(port - 1, 1);
 }
 
--- a/src/devices/vdml_imu.c
+++ b/src/devices/vdml_imu.c
@@ -40,7 +40,7 @@ int32_t imu_reset(uint8_t port) {
 	ERROR_IMU_STILL_CALIBRATING(port, device, PROS_ERR);
 	vexDeviceImuReset(device->device_info);
 	// delay for vexos to set calibration flag
-	while(!(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING)) delay(1);
+	while(!(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING)) delay(5);
 	return_port(port - 1, 1);
 }
 
--- a/src/devices/vdml_imu.c
+++ b/src/devices/vdml_imu.c
@@ -39,6 +39,8 @@ int32_t imu_reset(uint8_t port) {
 	claim_port_i(port - 1, E_DEVICE_IMU);
 	ERROR_IMU_STILL_CALIBRATING(port, device, PROS_ERR);
 	vexDeviceImuReset(device->device_info);
+	// delay for vexos to set calibration flag
+	delay(20);
 	return_port(port - 1, 1);
 }
 


Buzzec:
The current [documentation](https://github.com/purduesigbots/pros/blob/866cfefcf7b089ca28e9e0bdfcb14cd126ecbe42/include/pros/imu.h#L61) for `imu_reset` states 
```
This takes approximately 2 seconds, and is a non-blocking operation.
```
This change would make this a blocking operation so the documentation needs to be changed to reflect the new behavior. I think the documentation is enough already and is user error if they don't put a delay before using the imu after a reset.

belericant:
The change to `imu_reset` isn't meant to block for the entirety of calibration. The current example for the documentation is:
```c
#define IMU_PORT 1

void initialize() {
  imu_reset(IMU_PORT);
  int time = millis();
  int iter = 0;
  while (imu_get_status(IMU_PORT) & E_IMU_STATUS_CALIBRATING) {
    printf("IMU calibrating... %d\n", iter);
    iter += 10;
    delay(10);
  }
  // should print about 2000 ms
  printf("IMU is done calibrating (took %d ms)\n", iter - time);
}
```
This does not work currently because on the return of `imu_reset` a subsequent call to `imu_get_status` will still show the IMU as done calibrating which is not user error as the expected behavior after `imu_reset` is that the IMU status return that the IMU is calibrating.

WillXuCodes (changes requested):
Given that the 12 ms delay could change depending on the amount of motors attached, I recommend a more variable approach:

`
while(!(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING)) delay(1);
`

HotelCalifornia:
> Given that the 12 ms delay could change depending on the amount of motors attached, I recommend a more variable approach:
> 
> `while(!(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING)) delay(1);`

this fundamentally changes the nature of the calibration function from nonblocking to blocking. as discussed above, the point of this change is to allow the system time for the calibrating status flag to update so that users can query status as we recommend in the documentation, NOT to block for the entirety of calibration.

WillXuCodes:
> > Given that the 12 ms delay could change depending on the amount of motors attached, I recommend a more variable approach:
> > `while(!(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING)) delay(1);`
> 
> this fundamentally changes the nature of the calibration function from nonblocking to blocking. as discussed above, the point of this change is to allow the system time for the calibrating status flag to update so that users can query status as we recommend in the documentation, NOT to block for the entirety of calibration.

Right, hence the **!**(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING) is intended to wait until the calibrating status flag has been updated properly before exiting the function scope. If we wanted to block the entirety of calibration, the **not** would not be there.

nathan-moore:
@belericant in what cases are you calling imu reset? Is that use case covered by the tare functions?

belericant:
> @belericant in what cases are you calling imu reset baring doing it initially? Is that use case covered by the tare functions?

I think most use cases of reset are only in the init portion of the program. It's just the reset function can cause many issues later on since code that waits on the IMU to be calibrated runs prematurely and many users don't figure this out because the example in the docs has this very issue and is assumed to be correct.



> > > Given that the 12 ms delay could change depending on the amount of motors attached, I recommend a more variable approach:
> > > `while(!(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING)) delay(5);`
> > 
> > 
> > this fundamentally changes the nature of the calibration function from nonblocking to blocking. as discussed above, the point of this change is to allow the system time for the calibrating status flag to update so that users can query status as we recommend in the documentation, NOT to block for the entirety of calibration.
> 
> Right, hence the **!**(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING) is intended to wait until the calibrating status flag has been updated properly before exiting the function scope. If we wanted to block the entirety of calibration, the **not** would not be there.

This is probably the more elegant solution. I just stuck with one delay because I found in testing that the time needed between reset being called and the flag being updated was consistent across all runs.

WillXuCodes:
Neat, well the sooner you can make the suggested changes and all that, we will test and see if the change works with your test code.

WillXuCodes (changes requested):
Could you make the delay 5 ms? As well as update the non-blocking comment in the function header.

belericant:
> Could you make the delay 5 ms? As well as update the non-blocking comment in the function header.

Which comment are you referring to?

WillXuCodes:
> > Could you make the delay 5 ms? As well as update the non-blocking comment in the function header.
> 
> Which comment are you referring to?

https://github.com/purduesigbots/pros/blob/develop/include/pros/imu.h#L64 

and the respective commend in the hpp file. (Maybe say something such as "and is a blocking operation until the IMU status flag is set properly."

WillXuCodes:
image
Just tested the PR, the delay works. However, it must be noted that I had to add a short delay before running reset or the entire program would hang. This is however **irrelevant** to this PR due to it being a known issue that happens on every V5 coding environment.

WillXuCodes (approved):

WillXuCodes:
With testing done and showing similar values to @belericant's testing, this PR looks good to me!

PR #290, opened 2021-05-14
✨ Added rotation data rate function

kunwarsahni01:
#### Summary:
Adds new data rate set functions for rotation sensor


WillXuCodes (approved):
LGTM, I'd ask someone to test something like this normally but there's *not* exactly a good way to do so.

PR #293, opened 2021-05-16
🚸 add pros::IMU alias

### LINKED ISSUE #288: Add Alias for pros::IMU as pros::Imu
Issue is what it says on label, the rationale behind this is because newer users get confused with the API or even mispellings of the fully uppercase IMU type.

AyushShukla3:
#### Summary:
Create the pros::Imu alias for pros::IMU in imu.h and imu.hpp

#### Motivation:
Reduce confusion with newer users attempting to use pros::Imu instead of pros;:IMU

##### References (optional):
closes #288


include/api.h
```diff
@@ -40,9 +40,9 @@
 #endif /* __cplusplus */
 
 #define PROS_VERSION_MAJOR 3
-#define PROS_VERSION_MINOR 4
-#define PROS_VERSION_PATCH 0
-#define PROS_VERSION_STRING "3.4.0"
+#define PROS_VERSION_MINOR 3
+#define PROS_VERSION_PATCH 2
+#define PROS_VERSION_STRING "3.3.2-dirty.10.5abe157"
```
HotelCalifornia, line 45: Don't check this file into features or bugfixes

include/pros/imu.h
```diff
@@ -28,6 +28,8 @@ namespace pros {
 namespace c {
 #endif
 
+typedef struct IMU Imu;
```
HotelCalifornia, line 31: does this even compile? there are no such structs defined in the C API. I think it is sufficient to provide that alias for the C++ API only.

version
```diff
@@ -1 +1 @@
-3.4.0
+3.3.2-dirty.10.5abe157
```
HotelCalifornia, line 1: this file should also not be checked into changes like this

HotelCalifornia (changes requested):

HotelCalifornia:
also if you edit the description above to say "closes #288" it will automatically close that issue when this is merged

WillXuCodes (changes requested):
Newline on Version

include/pros/imu.hpp
```diff
@@ -444,6 +444,9 @@ class Imu {
 	 */
 	virtual bool is_calibrating() const;
 };
+
+	using IMU = Imu;
```
HotelCalifornia, line 448: should not be indented here

WillXuCodes (dismissed):
LGTM

HotelCalifornia (approved):

PR #294, opened 2021-05-30
🔥 Deprecate PID setting functions

### LINKED ISSUE #291: Remove Motor PID Setting Functions
Given that the Motor PID setting functions break the motors, we should remove these functions from the working API.

The way we'd do this is just to have these functions do nothing when called, and print a deprecation warning or something.

AyushShukla3:
#### Summary:
Removed the functionality of pid setting functions and added deprecation warnings to them. The functions included are:

for the c implementation--
`pros::c::motor_set_pos_pid`
`pros::c::motor_set_pos_pid_full`
`pros::c::motor_set_vel_pid`
`pros::c::motor_set_vel_pid_full`

for the C++ implementation--
`pros::Motor::set_pos_pid`
`pros::Motor::set_pos_pid_full`
`pros::Motor::set_vel_pid`
`pros::Motor::set_vel_pid_full`

#### Motivation:
As stated in #291 motors can be broken using the pid setting functions therefore they should be removed.

##### References (optional):
closes #291

#### Test Plan:
Unfortunately, I do not have access to a brain, so I tested these functions by calling them in main.cpp in a main() method.

For each method I wrote--
```C++
int main() {
	// set_pos_pid
	pros::Motor motor(1);
	pros::motor_pid_s_t pid = pros::Motor::convert_pid(KF, KP, KI, KD);
	motor.set_pos_pid(pid);
}
```
And at compile time we receive the following warning--
```
src/main.cpp:16:23: warning: 'virtual int32_t pros::Motor::set_pos_pid(pros::motor_pid_s_t) const' is deprecated [-Wdeprecated-declarations]
  motor.set_pos_pid(pid);
                       ^
In file included from ./include/api.h:69:0,
                 from ./include/main.h:37,
                 from src/main.cpp:1:
./include/pros/motors.hpp:618:23: note: declared here
  virtual std::int32_t set_pos_pid(const motor_pid_s_t pid) const;
                       ^~~~~~~~~~~
```
All of the other methods printed the same warnings.

### DIFF
--- a/src/devices/vdml_motors.cpp
+++ b/src/devices/vdml_motors.cpp
@@ -112,11 +112,17 @@ motor_gearset_e_t Motor::get_gearing(void) const {
 }
 
 motor_pid_full_s_t Motor::get_pos_pid(void) const {
+	#pragma GCC diagnostic push
+  	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 	return motor_get_pos_pid(_port);
+	#pragma GCC diagnostic pop
 }
 
 motor_pid_full_s_t Motor::get_vel_pid(void) const {
+	#pragma GCC diagnostic push
+  	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 	return motor_get_vel_pid(_port);
+	#pragma GCC diagnostic pop
 }
 
 std::int32_t Motor::get_raw_position(std::uint32_t* const timestamp) const {
@@ -196,12 +202,18 @@ std::int32_t Motor::set_gearing(const motor_gearset_e_t gearset) const {
 }
 
 motor_pid_s_t Motor::convert_pid(double kf, double kp, double ki, double kd) {
+	#pragma GCC diagnostic push
+  	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 	return motor_convert_pid(kf, kp, ki, kd);
+	#pragma GCC diagnostic pop
 }
 
 motor_pid_full_s_t Motor::convert_pid_full(double kf, double kp, double ki, double kd, double filter, double limit,
                                            double threshold, double loopspeed) {
+	#pragma GCC diagnostic push
+  	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 	return motor_convert_pid_full(kf, kp, ki, kd, filter, limit, threshold, loopspeed);
+	#pragma GCC diagnostic pop
 }
 
 std::int32_t Motor::set_pos_pid(const motor_pid_s_t pid) const {
--- a/src/devices/vdml_motors.cpp
+++ b/src/devices/vdml_motors.cpp
@@ -205,19 +205,31 @@ motor_pid_full_s_t Motor::convert_pid_full(double kf, double kp, double ki, doub
 }
 
 std::int32_t Motor::set_pos_pid(const motor_pid_s_t pid) const {
+	#pragma GCC diagnostic push
+  	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 	return motor_set_pos_pid(_port, pid);
+	#pragma GCC diagnostic pop
 }
 
 std::int32_t Motor::set_pos_pid_full(const motor_pid_full_s_t pid) const {
+	#pragma GCC diagnostic push
+  	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 	return motor_set_pos_pid_full(_port, pid);
+	#pragma GCC diagnostic pop
 }
 
 std::int32_t Motor::set_vel_pid(const motor_pid_s_t pid) const {
+	#pragma GCC diagnostic push
+  	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 	return motor_set_vel_pid(_port, pid);
+	#pragma GCC diagnostic pop
 }
 
 std::int32_t Motor::set_vel_pid_full(const motor_pid_full_s_t pid) const {
+	#pragma GCC diagnostic push
+  	#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 	return motor_set_vel_pid_full(_port, pid);
+	#pragma GCC diagnostic pop
 }
 
 std::int32_t Motor::set_zero_position(const double position) const {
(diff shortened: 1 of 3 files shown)


src/devices/vdml_motors.c
```diff
@@ -31,19 +31,13 @@ static V5_DeviceMotorPid get_pos_pid(uint8_t port) {
 	return ((motor_data_s_t*)registry_get_device(port)->pad)->pos_pid;
 }
 
-static void set_pos_pid(uint8_t port, V5_DeviceMotorPid pos) {
-	motor_data_s_t* data = (motor_data_s_t*)registry_get_device(port)->pad;
-	data->pos_pid = pos;
-}
+static void set_pos_pid(uint8_t port, V5_DeviceMotorPid pos) { /*deprecated*/ }
```
HotelCalifornia, line 34: Ah, the point of deprecation is that none of the functionality changes. Rather users are warned that the functionality will be going away in the future. All changes to this file and the c++ source file should be reverted.

WillXuCodes: To be fair, I think in this specific instance given the dangers of the functionality we should remove it. Sure, it's a breaking change but a very warranted one nonetheless.

HotelCalifornia: if you want to remove them now, just remove them now instead of doing this halfway thing.

my understanding was that we were following what all best practices for API maintenance dictate, which is to use deprecation as a warning that the function will not exist in the future. not that it currently does something different than what's documented.

this change is unlikely to affect many (if any) users anyway, so I personally think there's little harm in just marking as deprecated and leaving the functionality as is

WillXuCodes: > if you want to remove them now, just remove them now instead of doing this halfway thing.
> 
> my understanding was that we were following what all best practices for API maintenance dictate, which is to use deprecation as a warning that the function will not exist in the future. not that it currently does something different than what's documented.
> 
> this change is unlikely to affect many (if any) users anyway, so I personally think there's little harm in just marking as deprecated and leaving the functionality as is

I just have a feeling that despite marking them deprecated and leaving the functionality as is there is bound to be some stupid high school student that decides to run the code regardless and ends up breaking their v5 motors. Given that this outcome even sparsely happening from time to time is terrible, I would still give the go ahead to make the API removal. 

We should definently make it clear to our library writers (specifically rust and okapi) that we are doing this and they should remove any uses of these functions before upgrading their kernels (it'd cause a compilation error anyways).

theol0403: Not that I have much say, but I vote to remove the feature completely. However, if that is not an option, I agree with hotel to keep the functionality with deprecation. With the current change, it is still breaking, since it affects users who use the pid set. Then, it will be impossible to set the pid, yet the method still exists, showing a warning instead of a compile error.

It would make sense to deprecate now and remove in pros 4, but I think removing the feature now is a good option. You could then also remove stuff like `motor_pid_s_t` which is bloating pros, but doing nothing.

src/devices/vdml_motors.c
```diff
@@ -275,9 +270,10 @@ int32_t motor_set_pos_pid(uint8_t port, const motor_pid_s_t pid) {
 	out.kd = pid.kd;
 	set_pos_pid(port - 1, out);
 	vexDeviceMotorPositionPidSet(device->device_info, &out);
-	return_port(port - 1, 1);
```
theol0403, line 278: I feel like either way, it would stay consistent to clean out these methods, because not only is the code useless, it might actually be doing something unintended.

WillXuCodes: agreed... I was also told by VEX that there are some weird behaviors with these functions (such as trying to set the value equal to 0 will effectively do nothing to the current value) so having the functionality gutted would arguably make the behavior of these functions more consistent lol

theol0403:
You should remove everything related, like `motor_pid_full_s_t` and anything else now unused in the kernel. Doing a search for "pid" in the project might be a good way to make sure you get everything.

nathan-moore:
@AyushShukla3 I thought it was agreed upon that we'd depreciate and not remove? Usually you set the depreciation for a bit for a bit before the break. There also isn't much rush here to remove them since these functions have been here for years, what's a bit longer?

AyushShukla3:
My bad, I read that wrong for some reason, I'll fix it up tomorrow.

AyushShukla3:
I've added a message warning possibly breaking motors using the function, so the deprecation warning looks like this:
```
Compiled src/main.cpp [WARNINGS]
src/main.cpp: In function 'int main()':
src/main.cpp:15:23: warning: 'virtual int32_t pros::Motor::set_pos_pid(pros::motor_pid_s_t) const' is deprecated: Changing these values is not supported by VEX and may lead to permanent motor damage. [-Wdeprecated-declarations]
  motor.set_pos_pid(pid);
                       ^
In file included from ./include/api.h:69:0,
                 from ./include/main.h:37,
                 from src/main.cpp:1:
./include/pros/motors.hpp:618:23: note: declared here
  virtual std::int32_t set_pos_pid(const motor_pid_s_t pid) const;
                       ^~~~~~~~~~~
```

src/devices/vdml_motors.cpp
```diff
@@ -205,19 +205,31 @@ motor_pid_full_s_t Motor::convert_pid_full(double kf, double kp, double ki, doub
 }
 
 std::int32_t Motor::set_pos_pid(const motor_pid_s_t pid) const {
+	#pragma GCC diagnostic push
```
nathan-moore, line 208: Alternatively, you could do this all in one pragma/pop

src/devices/vdml_motors.cpp
```diff
@@ -205,19 +205,31 @@ motor_pid_full_s_t Motor::convert_pid_full(double kf, double kp, double ki, doub
 }
```
nathan-moore, line 205: Might as well depreciate these as well, since they're useless without the set Pids.

nathan-moore (dismissed):

src/devices/vdml_motors.cpp
HotelCalifornia: ah that's a good point

AyushShukla3:
So I added warnings to the following functions that are useless without pid setting functions, likely to be removed when the pid setting functions get removed:
`pros::c::motor_convert_pid`
`pros::c::motor_convert_pid_full`
`pros::c::motor_get_pos_pid`
`pros::c::motor_get_vel_pid`
`pros::Motor::convert_pid`
`pros::Motor::convert_pid_full`
`pros::Motor::get_pos_pid`
`pros::Motor::get_vel_pid`

nathan-moore (approved):

PR #317, opened 2021-10-16
✨ Add C style task create for any void() callable

TixoRebel:
#### Summary:
Adds a C style task create that accepts any void() callable like the new(ish) C++ constructors.

#### Motivation:
I'm not a fan of how the C++ API ties task creation to task object creation so prefer to use the C API but would really like the ability to pass any void() callable, like a lambda with captures.

#### Test Plan:
Run the following test to verify functionality:
```cpp
#include "main.h"

void initialize() {
	const char *s1 = "test1", *s2 = "test2";
	pros::task_t handle = pros::Task::create([&] {
		std::cout << s1 << std::endl;
	}, "task1");
	std::cout << pros::c::task_get_name(handle) << std::endl;

	pros::Task task2{[&] {
		std::cout << s2 << std::endl;
	}, "task2"};
	std::cout << task2.get_name() << std::endl;
}

void disabled() {}

void competition_initialize() {}

void autonomous() {}

void opcontrol() {}
```

- [x] Verify functionality

### DIFF
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -102,14 +102,59 @@ class Task {
 	 *
 	 */
 	template <class F>
+	static task_t create(F&& function, std::uint32_t prio = TASK_PRIORITY_DEFAULT,
+			 std::uint16_t stack_depth = TASK_STACK_DEPTH_DEFAULT, const char* name = "") {
+		static_assert(std::is_invocable_r_v<void, F>);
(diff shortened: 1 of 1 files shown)


TixoRebel:
> does this give us an opportunity to reduce repeated code? can we use this static function in the equivalent object constructor?

Yeah, I've now switched it over to using the static function by calling the `Task(task_t)` constructor.

PR #324, opened 2021-10-28
✨ Add Dedicated ADI Potentiometer Class

### LINKED ISSUE #301: Add dedicated class for potentiometer
With the release of the new potentiometer, we should support a potentiometer class and methods that allow users to configure the following:
- What type of potentiometer it is (new or old, will effect what degree display the 0-4095 values represent).
- Display the degree value.
- Get the raw value without breaking people's code (keep the get_value functions).

The goal of this is to prevent confusion for users not familiar with the difference between the 2 potentiometers to believe that they both have the same range, and a change that I believe would simply be helpful for people as well.

AyushShukla3:
#### Summary:
A wrapper for the potentiometer was created for the C++ API. The new potentiometer functions allows users to create potentiometer object specifying the potentiometer version (potentiometer_v1 vs potentiometer_v2) which will affect the degree value that is printed when they use the new ADIPotentiometer::get_value() function. Init functions were added to the C API to specify potentiometer versions in place of a constructor and allows users to call adi_potentiometer_get() to get values in degrees converted to degrees by potentiometer version.

#### Motivation:
The new potentiometer class allows users to create a potentiometer object and get values converted to degrees depending on the version of the potentiometer. The goal is to reduce confusion between the different versions which have different angles of rotation.

##### References (optional):
Closes #301 

#### Test Plan:
With a brain and a potentiometer v1, I will test the following:

In the C API:
- [x]  Run `adi_potentiometer_type_init()` specifying potentiometer v1 and calling `adi_potentiometer_get` (should output degree values 0-250).
- [x]  Run `adi_potentiometer_type_init()` specifying potentiometer v2 and calling `adi_potentiometer_get` (should output degree values 0-330).
- [x] Run `adi_potentiometer_init` without specifying potentiometer type (test default init function) (should output degree values 0-250).

In the C++ API:
- [ ] Create an ADIPotentiometer object specifying potentiometer v1 and calling `ADIPotentiometer::get_value()` (should output degree values 0-250).
- [ ] Create an ADIPotentiometer object specifying potentiometer v2 and calling `ADIPotentiometer::get_value()` (should output degree values 0-330).
- [ ]  Create ADIPotentiometer object without specifying potentiometer type (test default `adi_potentiometer_type_e_t` value).
- [ ]  Test inherited `ADIAnalogIn` methods with an ADIPotentiometer object.
- [x]  Confirm `ADIAnalogIn::get_adi_port()`, `ADIAnalogIn::get_smart_port()`, `ADIAnalogIn::set_adi_port()`, `ADIAnalogIn::set_smart_port()` are not callable with an ADIPotentiometer object.

Code:
```C++
//old potentiometer c init
pros::c::adi_potentiometer_t pot1 = pros::c::adi_potentiometer_type_init('A', pros:: E_ADI_POT_EDR);
while (true) {
	printf("pot value: %f\n", pros::c::adi_potentiometer_get_angle(pot1));
	pros::delay(200);
}

//new potentiometer c init
pros::c::adi_potentiometer_t pot2 = pros::c::adi_potentiometer_type_init('A', pros:: E_ADI_POT_V2);
while (true) {
	printf("pot value: %f\n", pros::c::adi_potentiometer_get_angle(pot2));
	pros::delay(200);
}

//default potentiometer c init
pros::c::adi_potentiometer_t pot3 = pros::c::adi_potentiometer_init('A');
while (true) {
	printf("pot value: %f\n", pros::c::adi_potentiometer_get_angle(pot3));
	pros::delay(200);
}

//old potentiometer c++ constructor
pros::ADIPotentiometer pot4('A', pros:: E_ADI_POT_EDR);
while (true) {
	printf("pot value: %f\n", pot4.get_angle());
	pros::delay(200);
}

//new potentiometer c++ constructor
pros::ADIPotentiometer pot5('A', pros:: E_ADI_POT_V2);
while (true) {
	printf("pot value: %f\n", pot5.get_angle());
	pros::delay(200);
}

//potentiometer c++ default constructor (old potentiometer)
pros::ADIPotentiometer pot6('A');
while (true) {
	printf("pot value: %f\n", pot6.get_angle());
	pros::delay(200);
}
```


Jerrylum:
image

1. and 2.  You should use `claim_port_i(smart_port - 1, E_DEVICE_ADI);` instead, since `smart_port` start from 0. See no.3
2. and 3. Duplicated `claim_port`,
4. and 5. the code should not always return 1
6. The formula is wrong, also it should be 333

```cpp
double ext_adi_potentiometer_get_value_degrees(ext_adi_potentiometer_t potentiometer) {
	double rtn;
	int32_t result;
	uint8_t smart_port, adi_port;
	get_ports(potentiometer, smart_port, adi_port);

	result = ext_adi_analog_read(smart_port, adi_port);

	if (result < 0 || result > 4095)
		return result; // error code

	adi_data_s_t* const adi_data = &((adi_data_s_t*)(device->pad))[potentiometer];

	if (adi_data->potentiometer_data.potentiometer_version) { // XXX: should not use bool
		rtn = result * 333 / 4095.0;
	} else {
		rtn = result * 250 / 4095.0;
	}

	return rtn;
}
```

WillXuCodes:
> image
> 
> Note:
> 
> ```
> 1. and 2. You should use `claim_port_i(smart_port - 1, E_DEVICE_ADI);` instead, since `smart_port` start from 0. See no.3
> 2. and 3. Duplicated `claim_port`,
> 4. The code should not always return 1
> 6. The formula is wrong, also it should be 333
> ```
> 
> You can try my code. I fixed some of the problems:
> 
> ```c++
> double ext_adi_potentiometer_get_value_degrees(ext_adi_potentiometer_t potentiometer) {
> 	double rtn;
> 	int32_t result;
> 	uint8_t smart_port, adi_port;
> 	get_ports(potentiometer, smart_port, adi_port);
> 
> 	result = ext_adi_analog_read(smart_port, adi_port);
> 
> 	if (result < 0 || result > 4095)
> 		return result; // error code
> 
> 	adi_data_s_t* const adi_data = &((adi_data_s_t*)(device->pad))[potentiometer];
> 
> 	if (adi_data->potentiometer_data.potentiometer_version) { // XXX: should not use bool
> 		rtn = result * 333 / 4095.0;
> 	} else {
> 		rtn = result * 250 / 4095.0;
> 	}
> 
> 	return rtn;
> }
> ```

I agree for the most part, I'll go over Ayush's code with him and show that he needs to do the mutex claiming and releasing within a new ext_pot function (which is the only thing missing from your code above really).

Jerrylum:
The whole point is, the mutex claiming and releasing are already inside the `ext_adi_analog_read` method so I don't have to do it again. Anyway, I think you will have a better idea. Keep going!

WillXuCodes:
> The whole point is, the mutex claiming and releasing are already inside the `ext_adi_analog_read` method so I don't have to do it again. Anyway, I think you will have a better idea. Keep going!

Yeah, I believe a better method would be to use the vexos function that `ext_adi_analog_read` uses in order to get the value, so we can still claim/release the mutex within that same function (more thread safe).

include/pros/adi.hpp
```diff
@@ -222,9 +222,17 @@ class ADIAnalogIn : private ADIPort {
...
 	 */
 	using ADIPort::get_value;
+
+	std::uint8_t get_adi_port() const;
+
+	std::uint8_t get_smart_port() const;
+	
+	void set_adi_port(std::uint8_t adi_port);
+
+	void set_smart_port(std::uint8_t smart_port);
```
WillXuCodes, line 232: I understand you need these for accessing private members in derived classes, but is it possible to make these private? I don't think we should add these to the public API.

src/devices/vdml_ext_adi.c
```diff
@@ -408,3 +411,32 @@ int32_t ext_adi_gyro_shutdown(ext_adi_gyro_t gyro) {
...
+	transform_adi_port(adi_port);
+	claim_port_i(smart_port, E_DEVICE_ADI);
+
+	adi_data_s_t* const adi_data = &((adi_data_s_t*)(device->pad))[adi_port];
+	adi_data->potentiometer_data.potentiometer_type = potentiometer_type;
+
+	return_port(smart_port - 1, merge_adi_ports(smart_port - 1, adi_port + 1));
+}
+
+double ext_adi_potentiometer_get(ext_adi_potentiometer_t potentiometer) {
```
WillXuCodes, line 425:
```suggestion
double ext_adi_potentiometer_get_angle(ext_adi_potentiometer_t potentiometer) {
```

(do the same with all relevant areas), c++ call should be .get_angle().

include/pros/adi.h
```diff
@@ -69,6 +69,11 @@ typedef enum adi_port_config_e {
 	E_ADI_ERR = PROS_ERR
 } adi_port_config_e_t;
 
+typedef enum adi_potentiometer_type_e { 
+	OLD_POTENTIOMETER = false,
+	NEW_POTENTIOMETER = true
+} adi_potentiometer_type_e_t;
+
```
WillXuCodes, line 76:
```suggestion
typedef enum adi_potentiometer_type_e { 
	E_ADI_POT_EDR = 0,
	E_ADI_POT_V2
} adi_potentiometer_type_e_t;

```

This enum does not follow how the others are styled, and we shouldn't keep these as booleans in case more potentiometer types are released by vex.

src/devices/vdml_adi.cpp
```diff
@@ -140,4 +157,29 @@ double ADIGyro::get_value() const {
...
+	get_ports(_port, _smart_port, _adi_port);
+	set_smart_port(_smart_port);
+	set_adi_port(_adi_port);
+}
+
+ADIPotentiometer::ADIPotentiometer(ext_adi_port_pair_t port_pair, adi_potentiometer_type_e_t potentiometer_type) : ADIAnalogIn(std::get<1>(port_pair)) { 
+ 	std::int32_t _port = ext_adi_potentiometer_init(port_pair.first, port_pair.second, potentiometer_type);
+	std::uint8_t _smart_port, _adi_port;
+ 	_smart_port = get_smart_port();
+	_adi_port = get_adi_port();
```
WillXuCodes, line 175: What's the point of get_smart_port and get_adi_port if the values returned from them are going to be overwritten anyways?

include/pros/adi.hpp
WillXuCodes: Also see below about get_adi_port and get_smart_port

src/devices/vdml_ext_adi.c
```diff
@@ -408,3 +411,32 @@ int32_t ext_adi_gyro_shutdown(ext_adi_gyro_t gyro) {
...
+	int32_t rtn;
+	uint8_t smart_port, adi_port;
+	get_ports(potentiometer, smart_port, adi_port);
+	transform_adi_port(adi_port);
+	claim_port_i(smart_port, E_DEVICE_ADI);
+	validate_type(device, adi_port, E_ADI_ANALOG_IN);
+
+	adi_data_s_t* const adi_data = &((adi_data_s_t*)(device->pad))[potentiometer];
+
+	if (adi_data->potentiometer_data.potentiometer_type) {
```
WillXuCodes, line 435: turn this into switch case and use the enum values for the cases.

WillXuCodes (changes requested):

src/devices/vdml_ext_adi.c
```diff
@@ -408,3 +411,39 @@ int32_t ext_adi_gyro_shutdown(ext_adi_gyro_t gyro) {
...
+	switch (adi_data->potentiometer_data.potentiometer_type) {
+		case E_ADI_POT_EDR:
+			rtn = vexDeviceAdiValueGet(device->device_info, adi_port) * 333 / 4095.0;
+			break;
+		case E_ADI_POT_V2:
+			rtn = vexDeviceAdiValueGet(device->device_info, adi_port) * 250 / 4095.0;
+			break;
+		default:
+			errno = ENXIO;
+			return PROS_ERR;
```
WillXuCodes, line 445: set rtn instead of returning PROS_ERR directly here, or you won't be returning the mutex.

WillXuCodes (changes requested):

src/devices/vdml_ext_adi.c
WillXuCodes: Maybe also set errno to EINVAL.

BillWyeman:
One issue with get_angle is it assumes angle is in direct proportion with the raw value from the port. This is rarely true, though most of the time its close enough. I'm not sure if this is a significant issue but some teams may believe the values from this function are closer to the truth than they actually are, which may cause frustration if the difference causes issues for them.

WillXuCodes:
> One issue with get_angle is it assumes angle is in direct proportion with the raw value from the port. This is rarely true, though most of the time its close enough. I'm not sure if this is a significant issue but some teams may believe the values from this function are closer to the truth than they actually are, which may cause frustration if the difference causes issues for them. 

Yeah I totally understand where you're coming from, though I do want to note that those teams are still free to use the older API. This change was also brought up mostly to help differentiate the meaning of the raw return values of the two types of VEX potentiometers now available for purchase, and is there as an option for teams who may prefer to get a more user friendly degree value.

WillXuCodes (dismissed):
image

Finished up testing and bugfixes on this branch, I understand why we set up some of the ADI stuff we did originally the way we did but it's actually kind of cursed. We will look over this in PROS 4 with whatever changes we also need.

src/devices/vdml_ext_adi.c
```diff
@@ -408,3 +411,39 @@ int32_t ext_adi_gyro_shutdown(ext_adi_gyro_t gyro) {
...
+	validate_type(device, adi_port, E_ADI_ANALOG_IN);
+
+	adi_data_s_t* const adi_data = &((adi_data_s_t*)(device->pad))[adi_port];
+
+	switch (adi_data->potentiometer_data.potentiometer_type) {
+		case E_ADI_POT_EDR:
+			rtn = vexDeviceAdiValueGet(device->device_info, adi_port) * 250 / 4095.0;
+			break;
+		case E_ADI_POT_V2:
+			rtn = vexDeviceAdiValueGet(device->device_info, adi_port) * 330 / 4095.0;
```
Jerrylum, line 442: should be 333

Jerrylum (commented):
some changes

include/pros/adi.h
```diff
@@ -689,6 +697,66 @@ int32_t adi_gyro_reset(adi_gyro_t gyro);
...
+ * \return An adi_potentiometer_t object containing the given port, or PROS_ERR if the
+ * initialization failed.
+ */
+adi_potentiometer_t adi_potentiometer_type_init(uint8_t port, adi_potentiometer_type_e_t potentiometer_type);
+
+/**
+ * Gets the current potentiometer angle in tenths of a degree.
+ *
+ * The original potentiometer rotates 250 degrees thus returning an angle between 0-250 degrees.
+ * Potentiometer V2 rotates 333 degrees thus returning an angle between 0-333 degrees.
```
WillXuCodes, line 746:
```suggestion
 * The original potentiometer rotates 250 degrees thus returning an angle between 0-250 degrees.
 * Potentiometer V2 rotates 330 degrees thus returning an angle between 0-330 degrees.
```

include/pros/adi.hpp
```diff
@@ -658,6 +658,108 @@ class ADIGyro : private ADIPort {
...
+	 * \param potentiometer_type
+ 	 *        An adi_potentiometer_type_e_t enum value specifying the potentiometer version type
+	 */
+	ADIPotentiometer(ext_adi_port_pair_t port_pair, adi_potentiometer_type_e_t potentiometer_type = E_ADI_POT_EDR);
+
+	/**
+	 * Gets the current potentiometer angle in tenths of a degree.
+	 *
+	 * The original potentiometer rotates 250 degrees thus returning an angle between 0-250 degrees.
+	 * Potentiometer V2 rotates 333 degrees thus returning an angle between 0-333 degrees.
```
WillXuCodes, line 699:
```suggestion
	 * Potentiometer V2 rotates 330 degrees thus returning an angle between 0-330 degrees.
```

WillXuCodes (changes requested):
Comments should be 330, not 333

WillXuCodes (approved):
LGTM

PR #335, opened 2021-12-16
🐛 gcc exception table section name typo

mattfbacon:
#### Summary:
`gcc_except_table` was misspelled as `gcc_execpt_table`.

#### Motivation:
The linking is improper if the exception table is not included properly.

#### Test Plan:

- [ ] Verify that the exception table is now in the binary


HotelCalifornia (approved):

HotelCalifornia:
good catch

PR #340, opened 2022-01-12
:sparkles: Add Device class, get_port and get_type method

Jerrylum:
#### Summary:
Since we have get_port method in many device classes, we should have more. Therefore, I made a super class `Device` for get_port. It can also open the door for more device-wise functions to be implemented in the future.

In addition, I made `get_type` method in `Device`, just like `get_config` method in `ADIPort`. It returns the device's expected type.

#### Motivation:
Love, yes, it is.

##### References (optional):
#188 

#### Test Plan:
[ ] Run the test case

```cpp
// TODO
```


BillWyeman:
If this is implemented, it would be a great opportunity to implement a `timestamp()` (vexDeviceGetTimestampByIndex) function for all smart devices too

WillXuCodes:
So while I still want this to be a thing, I think this PR is better done once we finish up PROS 4. The reason I've been hesitant to merge this is because as we've realized with VEXLink, sometimes the plugged device *won't* line up with the intended device that we want to eventually register. There's also a good amount of merge conflicts surrounding this, but we appreciate you starting the preliminary work on this and we'll definently use your work as a baseline for when we do this again for PROS 4 right before we go into LVGL porting!

PR #341, opened 2022-01-14
🐛 Fix rotation sensor race condition

epotapov:
#### Summary:
Created a constructor with a block to prevent race conditions. Waits for the sensor initialization to complete to set the reverse flag.

#### Motivation:
https://www.vexforum.com/t/rotation-sensor-bug-workaround-on-vexos-1-1-0/96577

#### Test Plan:
It builds.

### DIFF
--- a/include/pros/rotation.h
+++ b/include/pros/rotation.h
@@ -180,7 +180,7 @@ int32_t rotation_set_reversed(uint8_t port, bool value);
  */
 int32_t rotation_reverse(uint8_t port);
 
-int32_t rotation_set_reversed_blocking(uint8_t port, bool value);
+int32_t rotation_init_with_reverse_flag(uint8_t port, bool reverse_flag);
 
 /**
  * Get the Rotation Sensor's reversed flag
--- a/include/pros/rotation.h
+++ b/include/pros/rotation.h
@@ -190,8 +190,8 @@ int32_t rotation_reverse(uint8_t port);
  *
  * \param  port
  * 				 The V5 Rotation Sensor port number from 1-21
- * \param reverse_flag
- *               Determines if the Rotation Sensor is reversed or not.
+ * \param  reverse_flag
+ * 				 Determines if the Rotation Sensor is reversed or not.
  * 
  * \return 1 if the operation was successful or PROS_ERR if the operation
  * failed, setting errno.
--- a/include/pros/rotation.h
+++ b/include/pros/rotation.h
@@ -180,6 +180,8 @@ int32_t rotation_set_reversed(uint8_t port, bool value);
  */
 int32_t rotation_reverse(uint8_t port);
 
+int32_t rotation_set_reversed_blocking(uint8_t port, bool value);
+
 /**
  * Get the Rotation Sensor's reversed flag
  * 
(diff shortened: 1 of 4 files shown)


src/devices/vdml_rotation.c
```diff
@@ -73,6 +75,26 @@ int32_t rotation_set_reversed(uint8_t port, bool value) {
...
+        timeoutCount += 5;
+        claim_port_i(port - 1, E_DEVICE_ROTATION);
+        if (timeoutCount >= ROTATION_RESET_TIMEOUT) {
+            port_mutex_give(port - 1);
+            errno = EAGAIN;
+            return PROS_ERR;
+        }
+        device = device; // suppressing compiler warning
+    } while(vexDeviceAbsEncStatusGet(device->device_info) == 0);
+    vexAbsEncReverseFlagSet(port - 1, reverse_flag);
```
WillXuCodes, line 94: We should be setting the flag before waiting for the flag status to update properly.

WillXuCodes (changes requested):

src/devices/vdml_rotation.c
```diff
@@ -73,6 +75,26 @@ int32_t rotation_set_reversed(uint8_t port, bool value) {
 	return_port(port - 1, 1);
 }
 
+int32_t rotation_init_with_reverse_flag(uint8_t port, bool reverse_flag) {
```
WillXuCodes, line 78: I know originally we wanted an initialization function, but I think a more accurate description would be "rotation_set_reversed_blocking" and update the header accordingly.

include/pros/rotation.h
```diff
@@ -180,6 +180,8 @@ int32_t rotation_set_reversed(uint8_t port, bool value);
  */
 int32_t rotation_reverse(uint8_t port);
 
+int32_t rotation_init_with_reverse_flag(uint8_t port, bool reverse_flag);
```
WillXuCodes, line 183: Be sure to add header docs

WillXuCodes (changes requested):

include/pros/rotation.h
```diff
@@ -180,6 +180,24 @@ int32_t rotation_set_reversed(uint8_t port, bool value);
...
+ *
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * ENXIO - The given value is not within the range of V5 ports (1-21).
+ * ENODEV - The port cannot be configured as an Rotation Sensor
+ *
+ * \param  port
+ * 				 The V5 Rotation Sensor port number from 1-21
+ * \param reverse_flag
+ *               Determines if the Rotation Sensor is reversed or not.
```
WillXuCodes, line 194: Add a tab or fix the spacing here.

src/devices/vdml_rotation.cpp
```diff
@@ -13,6 +13,11 @@
 #include "pros/rotation.hpp"
 
 namespace pros {
+    
+Rotation::Rotation(const std::uint8_t port, const bool reverse_flag) : _port(port) {
+    pros::c::rotation_init_reverse(port, reverse_flag);
```
WillXuCodes, line 18: minor task but it should be done, could we fix the spaces vs tabs issue with everything in this file?

WillXuCodes (changes requested):

WillXuCodes (approved):
LGTM :+1:

PR #257, opened 2020-10-02
✨ Add Support for the Distance Sensor

kunwarsahni01:
#### Summary:
Created a new API for the Distance Sensor

#### Motivation:
PROS should support the Distance Sensor because it's a VRC-legal electronic part and PROS supports all the VRC-legal electronics parts.

#### Test Plan:
- [x] Tested with hardware and all seems to function properly.

PR #345, opened 2022-01-22
:memo: Fix file headers typo

Jerrylum:
#### Summary:
Stop it

#### Motivation:
Send helps

##### References (optional):
#257 #259 

#### Test Plan:
uh

image


WillXuCodes (approved):
sad that it turns out every sensor is not an inertial, but glad that it's corrected nonetheless

PR #348, opened 2022-02-14
✨Add motor stop function

omegaStag:
#### Summary:
Adds a function to set the motor voltage to zero.

#### Motivation:
There isn't already a motor stop function

#### Test Plan:
Make motor run and use function to stop motor

### DIFF
--- a/include/pros/motors.h
+++ b/include/pros/motors.h
@@ -57,7 +57,12 @@ namespace c {
 int32_t motor_move(uint8_t port, int32_t voltage);
 
 /**
- * Sets the velocity for the motor to 0, stopping the motor with respect to set brake mode.
+ * Stops the motor using the currently configured brake mode.
+ * 
+ * This function sets motor velocity to zero, which will cause it to act
+ * according to the set brake mode. If brake mode is set to MOTOR_BRAKE_HOLD,
+ * this function may behave differently than calling motor_move_absolute(port, 0)
+ * or motor_move_relative(port, 0).
  *
  * This function uses the following values of errno when an error state is
  * reached:
--- a/include/pros/motors.h
+++ b/include/pros/motors.h
@@ -57,7 +57,7 @@ namespace c {
 int32_t motor_move(uint8_t port, int32_t voltage);
 
 /**
- * Sets the voltage for the motor to 0, stopping the motor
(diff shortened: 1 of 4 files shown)


WillXuCodes (approved):
Looks good

HotelCalifornia (changes requested):
I thought the point of this was to add a function that stops a motor respecting the set brake mode? as-is this PR doesn't do that. instead of internally calling move_voltage(0) it should call e.g. move_velocity(0).

also i'd recommend calling it motor_brake/pros::Motor::brake instead of motor_stop/pros::Motor::stop to make it clear that's what's happening.

include/pros/motors.h
```diff
@@ -56,6 +56,22 @@ namespace c {
  */
 int32_t motor_move(uint8_t port, int32_t voltage);
 
+/**
+ * Sets the velocity for the motor to 0, stopping the motor with respect to set brake mode.
```
HotelCalifornia, line 60:
```suggestion
 * Stops the motor using the currently configured brake mode.
 * 
 * This function sets motor velocity to zero, which will cause it to act
 * according to the set brake mode. If brake mode is set to MOTOR_BRAKE_HOLD,
 * this function may behave differently than calling motor_move_absolute(port, 0)
 * or motor_move_relative(port, 0).
```

include/pros/motors.hpp
```diff
@@ -185,6 +185,18 @@ class Motor {
 	 */
 	virtual std::int32_t move_voltage(const std::int32_t voltage) const;
 
+	/**
+	 * Sets the velocity for the motor to 0, stopping the motor with respect to set brake mode.
```
HotelCalifornia, line 189:
```suggestion
	 * Stops the motor using the currently configured brake mode.
	 *
	 * This function sets motor velocity to zero, which will cause it to act
	 * according to the set brake mode. If brake mode is set to MOTOR_BRAKE_HOLD,
	 * this function may behave differently than calling move_absolute(0)
	 * or move_relative(0).
```

HotelCalifornia (commented):
some updates and clarifications I thought would be useful in the documentation

WillXuCodes (approved):
Works on hardware, feel free to merge later tonight nearer to 3.6 release

PR #350, opened 2022-03-06
✨Add Task Join Method

omegaStag:
#### Summary:
``task_join(task_t task)`` method to have the program wait for the completion (and subsequent deletion of a task) before moving on to the rest of the program

#### Motivation:
Analogous to ``std::thread::join`` functionality in C++

#### Test Plan:
Write tasks that outputs a message and use task_join to call it in opcontrol- something like this?

```
#include "main.h"
#include "pros/apix.h"

void task_one(void* ignore) {
	lcd_print(1, "%s running", task_get_name(NULL));
	task_delay(1000);
	lcd_print(2, "End of %s", task_get_name(NULL));
}

void task_two(void* ignore) {
	lcd_print(3, "%s running", task_get_name(NULL));
	task_delay(1000);
	lcd_print(4, "End of %s", task_get_name(NULL));
}

void opcontrol() {
	task_t task1= task_create(task_one, NULL, TASK_PRIORITY_DEFAULT, TASK_STACK_DEPTH_DEFAULT, "Task One");
	task_t task2 = task_create(task_two, NULL, TASK_PRIORITY_DEFAULT, TASK_STACK_DEPTH_DEFAULT, "Task Two");
	lcd_set_text(0, "Running tasks.");
        task_join(task1);
        task_join(task2);
        task_delay(1000);
        lcd_set_text(5, "End of tasks.");
}
```

### DIFF
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -280,15 +280,15 @@ class Task {
 	std::uint32_t notify();
 
 	/**
-	 * Waits until specified task is complete and deleted, then continues to execute the program.
-	 * Analogous to std::thread::join in C++.
+ 	 * Utilizes task notifications to wait until specified task is complete and deleted,
+ 	 * then continues to execute the program. Analogous to std::thread::join in C++.
 	 *
 	 * See https://pros.cs.purdue.edu/v5/tutorials/topical/notifications.html for
 	 * details.
 	 *
 	 * \return void
 	 */
-	std::void_t join();
+	void join();
 
 	/**
 	 * Sends a notification to a task, optionally performing some action. Will
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -280,6 +280,17 @@ class Task {
 	std::uint32_t notify();
 
 	/**
+	 * Waits until specified task is complete and deleted, then continues to execute the program.
+	 * Analogous to std::thread::join in C++.
+	 *
+	 * See https://pros.cs.purdue.edu/v5/tutorials/topical/notifications.html for
+	 * details.
+	 *
+	 * \return void
+	 */
+	std::void_t join();
+
+	/**
 	 * Sends a notification to a task, optionally performing some action. Will
 	 * also retrieve the value of the notification in the target task before
 	 * modifying the notification value.
--- a/include/rtos/task.h
+++ b/include/rtos/task.h
@@ -1751,24 +1751,6 @@ int32_t task_notify(task_t xTaskToNotify);
 
 /**
  * task. h
- * <PRE>void task_join( task_t taskToJoin );</PRE>
- *
- * configUSE_TASK_NOTIFICATIONS must be undefined or defined as 1 for this macro
- * to be available.
- * 
- * Waits until specified task is complete and deleted, then continues to execute the program.
- * Analogous to std::thread::join in C++.
- *
- * @param xTaskToJoin The handle of the task to wait on.
- * 
- * \defgroup task_join task_join
- * \ingroup TaskNotifications
- */
-
-void task_join(task_t xTaskToJoin);
-
-/**
- * task. h
  * <PRE>void vTaskNotifyGiveFromISR( task_t xTaskHandle, int32_t *pxHigherPriorityTaskWoken );
  *
  * configUSE_TASK_NOTIFICATIONS must be undefined or defined as 1 for this macro
--- a/include/rtos/task.h
+++ b/include/rtos/task.h
@@ -1751,6 +1751,24 @@ int32_t task_notify(task_t xTaskToNotify);
 
 /**
  * task. h
+ * <PRE>void task_join( task_t taskToJoin );</PRE>
+ *
+ * configUSE_TASK_NOTIFICATIONS must be undefined or defined as 1 for this macro
+ * to be available.
+ * 
+ * Waits until specified task is complete and deleted, then continues to execute the program.
+ * Analogous to std::thread::join in C++.
+ *
+ * @param xTaskToJoin The handle of the task to wait on.
+ * 
+ * \defgroup task_join task_join
+ * \ingroup TaskNotifications
+ */
+
+void task_join(task_t xTaskToJoin);
+
+/**
+ * task. h
  * <PRE>void vTaskNotifyGiveFromISR( task_t xTaskHandle, int32_t *pxHigherPriorityTaskWoken );
  *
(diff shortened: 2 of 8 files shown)


nickmertin:
Two general comments:
1. Errors should be reported via a return value and `errno` as with other PROS functions.
2. It should be explicitly documented that this makes use of task notifications, so that users do not run into hard-to-debug issues when they try to join a task and then get woken up by an unrelated task notification.

include/pros/rtos.hpp
```diff
@@ -279,6 +279,17 @@ class Task {
...
+	/**
+	 * Waits until specified task is complete and deleted, then continues to execute the program.
+	 * Analogous to std::thread::join in C++.
+	 *
+	 * See https://pros.cs.purdue.edu/v5/tutorials/topical/notifications.html for
+	 * details.
+	 *
+	 * \return void
+	 */
+	std::void_t join();
```
HotelCalifornia, line 291: std::void_t is definitely not correct here.

include/rtos/task.h
```diff
@@ -1749,6 +1749,24 @@ int32_t task_notify_wait( uint32_t ulBitsToClearOnEntry, uint32_t ulBitsToClearO
...
+ * to be available.
+ * 
+ * Waits until specified task is complete and deleted, then continues to execute the program.
+ * Analogous to std::thread::join in C++.
+ *
+ * @param xTaskToJoin The handle of the task to wait on.
+ * 
+ * \defgroup task_join task_join
+ * \ingroup TaskNotifications
+ */
```
HotelCalifornia, line 1766: no need to copy freeRTOS style documentation here, the important part is in pros/rtos.h

and actually this declaration isn't necessary at all...

src/rtos/tasks.c
```diff
@@ -4745,6 +4745,19 @@ uint32_t uxReturn;
 #endif /* configUSE_TASK_NOTIFICATIONS */
 /*-----------------------------------------------------------*/
 
+/*-----------------------------------------------------------*/
+
+#if( configUSE_TASK_NOTIFICATIONS == 1 )
+
+	void task_join(task_t task) {
+    	task_notify_when_deleting(task, NULL, 1, E_NOTIFY_ACTION_INCR);
```
HotelCalifornia, line 4753: please fix indentation of this function

HotelCalifornia (dismissed):

HotelCalifornia:
> 1. Errors should be reported via a return value and `errno` as with other PROS functions.

not sure how to reconcile this with the C++ API where I believe the goal is to mirror std::thread::join behavior... I suppose the right way to do it would be to convert C API errors into exceptions but since that's not how the rest of the API works we might as well not bother right now.

> 2. It should be explicitly documented that this makes use of task notifications, so that users do not run into hard-to-debug issues when they try to join a task and then get woken up by an unrelated task notification.

I'm wondering if there's a way we can make this work... Like setting a specific bit in the notification value (and documenting that). problem is the way notify_take allegedly works, wherein the value is either cleared or decremented...

include/pros/rtos.hpp
omegaStag: wait then what is (i am not familiar with c++)

HotelCalifornia:
actually re: error handling: I take back what I said. I don't think we really need to be that concerned with it. this function is meant effectively as an extension of the freeRTOS tasks API rather than as a part of the PROS API (semantically speaking) so void is ok imo

src/rtos/tasks.c
```diff
@@ -4745,6 +4745,19 @@ uint32_t uxReturn;
 #endif /* configUSE_TASK_NOTIFICATIONS */
 /*-----------------------------------------------------------*/
 
+/*-----------------------------------------------------------*/
+
+#if( configUSE_TASK_NOTIFICATIONS == 1 )
+
+	void task_join(task_t task) {
+    		task_notify_when_deleting(task, NULL, 1, E_NOTIFY_ACTION_INCR);
+    		while (!pros::Task::notify_take(true, 20));
```
WillXuCodes, line 4754: be sure to use the C equivalent of this (task_notify_take)

WillXuCodes (changes requested):
c above

WillXuCodes:
Currently, the test code hangs. I'm going to look into it tommorow.

WillXuCodes:
This join implementation works on one thread at a time, but does not work when there are two threads that need to be joined (or maybe it doesn't work on threads that are already finished)

WillXuCodes (commented):
Tested with hardware, code still doesn't work as intended. I think we might have to just go with the hard check in the loop to make sure that the task isn't deleted already.

WillXuCodes (approved):
Works on hardware, free to merge tonight closer to 3.6 release.

src/rtos/tasks.c
```diff
@@ -4745,6 +4745,26 @@ uint32_t uxReturn;
...
+	void task_join(task_t task) {
+		if(!task) return;
+		TaskStatus_t xTaskDetails;
+		do
+		{
+			vTaskGetInfo(task, &xTaskDetails, pdTRUE, E_TASK_STATE_INVALID);
+			if(xTaskDetails.eCurrentState == E_TASK_STATE_DELETED) { 
+				break;
+			}
+		} while (!task_notify_take(true, 20));
```
HotelCalifornia, line 4761: does this mean the caller has to set up the task_notify_when_deleting call before invoking this function? task_join should probably handle that before it goes into the loop

src/rtos/tasks.c
```diff
@@ -4745,6 +4745,24 @@ uint32_t uxReturn;
...
+#if( configUSE_TASK_NOTIFICATIONS == 1 )
+
+	void task_join(task_t task) {
+		if(!task) return;
+		TaskStatus_t xTaskDetails;
+		task_notify_when_deleting(task, NULL, 1, E_NOTIFY_ACTION_INCR);
+		do
+		{
+			vTaskGetInfo(task, &xTaskDetails, pdTRUE, E_TASK_STATE_INVALID);
+		} while (!task_notify_take(true, 20) && xTaskDetails.eCurrentState != E_TASK_STATE_DELETED);
```
WillXuCodes, line 4759:
```suggestion
		} while (!task_notify_take(true, 20) || xTaskDetails.eCurrentState != E_TASK_STATE_DELETED);
```

WillXuCodes (changes requested):

WillXuCodes (approved):
1 star, not enough ranch

PR #360, opened 2022-04-13
🐛Fix Link Port Claim Failures

WillXuCodes:
#### Summary:
There's a case where the only radio on the brain is the VEXLink radio, and even with an override the icon on the brain shows as a "generic serial device" but the VEXos still believes that it's a radio. This PR fixes it by allowing VEXLink devices to also be radios. 

- [ ] Test case below works:
- [ ] Test more thoroughly with cases other than the code below:

```C++
void initialize() {
    pros::c::delay(50);
    pros::c::link_init_override(20, "TEST", pros::E_LINK_TRANSMITTER);

    while (!pros::c::link_connected(20)) {
      pros::c::delay(20);
    }
}

```

### DIFF
--- a/include/api.h
+++ b/include/api.h
@@ -68,7 +68,6 @@
 #include "pros/distance.hpp"
 #include "pros/gps.hpp"
 #include "pros/imu.hpp"
-#include "pros/link.hpp"
 #include "pros/llemu.hpp"
 #include "pros/misc.hpp"
 #include "pros/motors.hpp"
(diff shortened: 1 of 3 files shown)


WillXuCodes:
Closing as we're taking a new approach due to CPU0's device registration issues with multiple radios: similar to VEXCode we're going to *officially* support one link per brain (which is realistically all that is needed for competition).

Issue #362, opened 2022-05-17
undefined reference to getentropy


mattfbacon:
#### Expected Behavior:
Project's cold library compiles and links properly

#### Actual Behavior:
Error at link time:

```
Creating cold package with libc,libm,libpros,okapilib [ERRORS]
/usr/lib/gcc/arm-none-eabi/12.1.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/12.1.0/../../../../arm-none-eabi/lib/thumb/v7+fp/softfp/libstdc++.a(random.o): in function `std::(anonymous namespace)::__libc_getentropy(void*)':
/build/arm-none-eabi-gcc/src/gcc-12.1.0/libstdc++-v3/src/c++11/random.cc:179: undefined reference to `getentropy'
/usr/lib/gcc/arm-none-eabi/12.1.0/../../../../arm-none-eabi/bin/ld: /usr/lib/gcc/arm-none-eabi/12.1.0/../../../../arm-none-eabi/lib/thumb/v7+fp/softfp/libstdc++.a(random.o): in function `std::random_device::_M_init(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
/build/arm-none-eabi-gcc/src/gcc-12.1.0/libstdc++-v3/src/c++11/random.cc:452: undefined reference to `getentropy'
make: *** [common.mk:229: bin/cold.package.elf] Error 1
/usr/lib/gcc/arm-none-eabi/12.1.0/../../../../arm-none-eabi/bin/ld: ./firmware/libc.a(lib_a-arc4random.o): in function `_rs_stir':
arc4random.c:(.text._rs_stir+0x8): undefined reference to `getentropy'
collect2: error: ld returned 1 exit status
```

#### Steps to reproduce:

100% reproducible. Create a new project with `prosv5 conductor new-project` and build it with `prosv5 make`.

#### System information:
Platform: V5
PROS Kernel Version: 3.6.0

#### Additional Information

OS: Arch Linux

HotelCalifornia:
yeah this happens because arch updates the toolchain faster than arm does (and also we haven't updated our copy of libc.a in like 3 years).

the workaround for you is to define

```c
int getentropy(void *buffer, size_t length) {
    return -1;
}
```

[...]

PR #363, opened 2022-05-18
🐛Add getentropy to newlib stubs

mattfbacon:
#### Summary:
Add a stub implementation of the `getentropy` function to the newlib stubs included in libpros.

#### Motivation:
Currently, if the arm-none-eabi toolchain's libc is too new, it will emit a reference to getentropy, which results in a linker error

##### References

Closes https://github.com/purduesigbots/pros/issues/362

#### Test Plan:

- [ ] Make sure that a new project with the patched libpros.a builds on Arch Linux

### DIFF
--- a/src/system/newlib_stubs.c
+++ b/src/system/newlib_stubs.c
@@ -46,6 +46,10 @@ unsigned sleep( unsigned period ) {
 	return 1;
 }
 
+int getentropy(void *_buffer, size_t _length) {
+	return -1;
+}
+
 // HACK: this helps confused libc++ functions call the right instruction. for
 // info see https://github.com/purduesigbots/pros/issues/153#issuecomment-519335375
 void __sync_synchronize(void) {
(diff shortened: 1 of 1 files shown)


mattfbacon:
Unless V5 API includes a getrandom syscall (are you even allowed to tell me if it does??)

WillXuCodes:
[kernel@3.6.1-pr363.3463.zip](https://github.com/purduesigbots/pros/files/8712811/kernel%403.6.1-pr363.3463.zip)

This is the artifact produced, let me know if it works on your machine @mattfbacon

mattfbacon:
It works! Thanks.

src/system/newlib_stubs.c
```diff
@@ -46,6 +46,10 @@ unsigned sleep( unsigned period ) {
 	return 1;
 }
 
+int getentropy(void *_buffer, size_t _length) {
+	return -1;
```
HotelCalifornia, line 50: this should additionally set `errno = ENOSYS`

HotelCalifornia (changes requested):
other than my one comment it looks good. since this was tested and confirmed working on an arch machine we can get this merged as soon as that change is made

WillXuCodes (approved):

WillXuCodes (approved):

HotelCalifornia (approved):

PR #366, opened 2022-06-03
 ✨Reorganize API with ADI and V5 Namespaces

WillXuCodes:
#### Summary:
This PR (for PROS 4 Development branch) is intended to add the v5 and ADI namespaces to help reorganize our API in a cleaner way that is more "C++ like".

#### Motivation:
We will need to work on doxygen for the new C++ API so I'm opening this PR now so it can be worked on.


src/devices/vdml_adi.cpp
```diff
@@ -15,148 +15,148 @@
 #include "vdml/port.h"
 
 namespace pros {
+namespace adi{
```
HotelCalifornia, line 18: missing space

PR #369, opened 2022-06-07
🔥 PROS 4: Decouple LVGL Directly from Kernel

WillXuCodes:
#### Summary:
Decouples LVGL from Kernel, and adds weak symbols to keep main.cpp compileable.
Also adds weak symbol for lvgl_init() for template implementation.

- [x] Make sure lvgl_init() actually works in initialize, with template implementation in a normal user program.


WillXuCodes:
Potential Training Task: Somebody who is not familiar with git can attempt to solve the merge conflict on this branch.

PR #376, opened 2022-06-13
🐛 fix linker script improper use of keyword as region name

### LINKED ISSUE #364: arm-none-eabi-binutils 2.38 linker script syntax error
OS: Arch Linux
PROS version: 3.6.0

When running `prosv5 make` and/or `make`:

```
Creating cold package with libc,libm,libpros,okapilib [ERRORS]
/usr/lib/gcc/arm-none-eabi/12.1.0/../../../../arm-none-eabi/bin/ld:./firmware/v5-common.ld:30: syntax error
collect2: error: ld returned 1 exit status
make: *** [common.mk:229: bin/cold.package.elf] Error 1
```

The syntax looks fine to me. Not sure what's tripping up the linker here.

Aang099:
Gnu ld 2.38 now uses MEMORY as a keyword, while the pros linker scripts uses it as the name of a region, causing an error during linking that prevents any project from building. This PR contains a simple fix for that by renaming all instances of MEMORY to RAM. 

closes #364

### DIFF
--- a/firmware/v5-common.ld
+++ b/firmware/v5-common.ld
@@ -27,15 +27,15 @@ SECTIONS
    *(.vfp11_veneer)
    *(.ARM.extab)
    *(.gnu.linkonce.armextab.*)
-} > MEMORY
-
+} > RAM
+ 
 .init : {
    KEEP (*(.init))
-} > MEMORY
+} > RAM
 
 .fini : {
    KEEP (*(.fini))
-} > MEMORY
+} > RAM
 
 .rodata : {
    __rodata_start = .;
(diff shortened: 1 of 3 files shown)


firmware/v5-common.ld
```diff
@@ -27,46 +27,46 @@ SECTIONS
    *(.vfp11_veneer)
    *(.ARM.extab)
    *(.gnu.linkonce.armextab.*)
-} > MEMORY
-
+} > RAM
+
```
HotelCalifornia, line 31: remove extraneous space

HotelCalifornia (commented):
lgtm, aside from the one extra space

firmware/v5-common.ld
Aang099: Done

WillXuCodes (approved):
Sauce

PR #382, opened 2022-06-19
✨Add MutexVar

nickmertin:
#### Summary:
Add a new class template, `MutexVar<Var>`, which wraps a mutex and a variable of type `Var`, allowing access to the protected variable only via a `MutexVarLock`, which is a specialized RAII handle representing exclusive access.

#### Motivation:
This is inspired by and implements the same pattern as [Rust's `Mutex` type](https://doc.rust-lang.org/std/sync/struct.Mutex.html). It provides a means of using the compiler to enforce a contract of exclusive access to a variable via a mutex, rather than relying on the programmer to ensure that they bracket accesses with `take()` and `give()` calls. The RAII pattern also ensures that it is exception-safe.

#### Test Plan:

- [ ] Test the following program:

```cpp
MutexVar<bool> flag{false};

void opcontrol() {
  Task([] {
    for (int i = 0; i < 100; i++) {
      *flag.lock() = true;
      delay(10);
    }
  });

  while (true) {
    if (*flag.lock()) {
      std::cout << "Flag" << std::endl;
    }
    delay(1);
  }
}
```

### DIFF
--- a/include/pros/rtos.hpp
+++ b/include/pros/rtos.hpp
@@ -624,7 +624,7 @@ class MutexVar {
 	MutexVarLock<Var> lock() {
 		while (!mutex.take(TIMEOUT_MAX))
 			;
-		return {{mutex, var}};
+		return {mutex, var};
 	}
 };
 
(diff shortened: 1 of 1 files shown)


WillXuCodes:
Getting an error here with your test code: 
image

nickmertin:
Oops, fixed that typo.

WillXuCodes (approved):

Issue #375, opened 2022-06-10
PROS 4: Update to C++20


karmanyaahm:
Hello,
Are there any plans to update default projects to C++20 rather than 17?  I'm doing an OkapiLib feature, and it would be really nice to have everyone using c++20 for std::format

HotelCalifornia:
not currently but i've documented a way to change it here:

https://github.com/hotelcalifornia/libhotel#usage

this would also be a very easy feature to PR here, if you want to take a stab at it. I'd been meaning to add this but never got around to it

HotelCalifornia:
also looking at the readme again i think the if statements can be replaced with ```mk
C_STANDARD?=gnu11
CXX_STANDARD?=gnu++17
```

HotelCalifornia:
> Are there any plans to update default projects to C++20 rather than 17?

lastly, to answer this specifically: probably not until PROS 4

WillXuCodes:
Yeah we could totally do this for PROS 4. I'll work on it in the coming few days.

WillXuCodes:
> using c++20 for std::format

Sorry to break it to you, but I don't think GCC supports std::format in any form at the moment. Although we will be upgrading for PROS 4, I recommend you look for other options for whatever you are trying to accomplish.

karmanyaahm:
Yes, thanks for taking care of this. I have already implemented this using sprintf and a fixed buffer
 https://github.com/OkapiLib/OkapiLib/pull/487/files#diff-cd6a0e232dbe657c242a1f904fc3c7f1b9e03f1811e4d9c77c7128d774cec54bR18-R32

WillXuCodes:
This is done for PROS 4.

PR #384, opened 2022-06-22
✨PROS 4: Update to C++20

WillXuCodes:
#### Summary:
Retarget building to C++20 for PROS 4
#375 
#### Test Plan:
Compile with  std::format or other C++20 features and make sure they work, implement whatever we need afterwards to facilitate this if need be.


Makefile
```diff
@@ -16,6 +16,9 @@ EXTRA_INCDIR=$(FWDIR)/libv5rts/sdk/vexv5/include
 # Directories to be excluded from all builds
 EXCLUDE_SRCDIRS+=$(SRCDIR)/tests
 
+C_STANDARD?=gnu11
```
HotelCalifornia, line 19: these should be `C_STANDARD=` etc

common.mk
```diff
@@ -24,8 +24,8 @@ wlprefix=-Wl,$(subst $(SPACE),$(COMMA),$1)
 LNK_FLAGS=--gc-sections --start-group $(strip $(LIBRARIES)) -lgcc -lstdc++ --end-group -T$(FWDIR)/v5-common.ld
 
 ASMFLAGS=$(MFLAGS) $(WARNFLAGS)
-CFLAGS=$(MFLAGS) $(CPPFLAGS) $(WARNFLAGS) $(GCCFLAGS) --std=gnu11
-CXXFLAGS=$(MFLAGS) $(CPPFLAGS) $(WARNFLAGS) $(GCCFLAGS) --std=gnu++17
+CFLAGS=$(MFLAGS) $(CPPFLAGS) $(WARNFLAGS) $(GCCFLAGS) --std=$(C_STANDARD)
```
HotelCalifornia, line 27: continued from previous comment, common.mk should then set `C_STANDARD?=gnu11` etc

noam987:
I think the azure pipeline needs to use a newer version of GCC. It is currently using the Q3 2019 release, while C++ 20 didn't release until December 2020. I would recommend migrating it to a 2021 or 2022 version

omegaStag (approved):
tested some C++20 features such as ``__VA_OPT__`` or ``char8_t`` on hardware, though clangd linter was kicking up a bit of a fuss it still compiled and ran on the brain

i guess if anyone has issues trying to use C++20 stuff tell them to check the C++ feature table for GCC first

PR #385, opened 2022-06-23
Pros4 enum/struct classes

### LINKED ISSUE #353: PROS 4: Change Various Structs/Enums to pros namespace, instead of being in pros::c, change all enums to be enum classes
Currently, we have many structs and enums (namely in imu.h and gps.h) that should probably be in the pros namespace rather than the pros::c namespace. This is to have a better distinction between what is part of the "pros::c" API vs default C++ pros API. Since the PROS API is primarily C++, we want to minimize interaction with the ::c namespace for users of the C++ only API. 

However, I'd like to get some input on this tbh. May or may not be a good change and it's definently something we want as a PROS 4 thing to prevent user confusion on PROS 3. 

Good representation of what I'm talking about
https://github.com/purduesigbots/pros-docs/pull/268

aaryan-gautam:
#### Summary:

I have moved the structs in pros::c to pros namespace in both the imu.h and gps.h. The respective modifications in the files vdml_imu.cpp and vdml_gps.cpp have also been made to appropriately match the modified namespace. The enum in imu.h was changed to enum class.

#### Motivation:

These changes make PROS4 a more C++ focused API.

##### References (optional):

Closes #353 

#### Test Plan:

Run CI tests to ensure there are no breaking changes

- [ ] test item

### DIFF
--- a/include/pros/gps.h
+++ b/include/pros/gps.h
@@ -26,6 +26,13 @@
 extern "C" {
 namespace pros {
 	
+typedef struct __attribute__((__packed__)) gps_status_s {
+	double x;      ///< X Position (meters)
+	double y;      ///< Y Position (meters)
+	double pitch;  ///< Percieved Pitch based on GPS + IMU
+	double roll;   ///< Percieved Roll based on GPS + IMU
+	double yaw;    ///< Percieved Yaw based on GPS + IMU
+} gps_status_s_t;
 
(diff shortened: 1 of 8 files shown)


HotelCalifornia:
changed the base to target develop-pros-4 instead of develop, please rebase your local work onto that branch and force push here

Issue #398, opened 2022-08-01
imu_reset appears to normally time out


Octogonapus:
#### Expected Behavior:
`pros::c::imu_reset` should wait until the IMU has set its calibration flag. This calibration process takes approximately 2 seconds. The `imu_reset` function should also time out after an excessive amount of time has passed, though this should not be the normal operating mode.

#### Actual Behavior:
The `imu_reset` function appears to normally time out. It has a 1 second timeout, and the IMU calibration process is documented to take around 2 seconds, so this is logical.

It would be helpful if someone with an IMU could create a MWE that simply resets the IMU using the VDML directly and measures how long the reset takes.

#### System information:

latest V5 kernel

Octogonapus:
Once someone with an IMU can confirm the true calibration time, we can consider adjusting the `imu_reset` timeout.

WillXuCodes:
So the code in question I believe is this: 
```C++
do {
		port_mutex_give(port - 1);
		task_delay(5);
		timeoutCount += 5;
		claim_port_i(port - 1, E_DEVICE_IMU);
		if (timeoutCount >= IMU_RESET_TIMEOUT) {
			port_mutex_give(port - 1);
			errno = EAGAIN;
			return PROS_ERR;
		}
		device = device; // suppressing compiler warning
	} while(!(vexDeviceImuStatusGet(device->device_info) & E_IMU_STATUS_CALIBRATING));
```

This isn't intended to be a blocking call to wait until the IMU is done calibrating, but to wait until the hardware is finished setting the "calibrating" flag. This was to prevent an issue where the IMU wasn't doing this correctly and checking the IMU status immediately after resetting was leading to the status being reported incorrectly as "not calibrating". 

However, it's really strange that it's not setting the flag now... we should test this with hardware and figure out what's going on.

WillXuCodes:
TLDR: 1 second is not how long we wait for the IMU to fully reset, but rather how long we are waiting for the IMU/hardware to set the status flag correctly.

Octogonapus:
> So the code in question I believe is this:

Indeed, it is.

> This isn't intended to be a blocking call to wait until the IMU is done calibrating, but to wait until the hardware is finished setting the "calibrating" flag.

Oh, my understanding was that those two are the same thing. What is the proper way to check if calibration has completed?

WillXuCodes:
[...]

PR #399, opened 2022-08-03
✨Add Blocking Reset Option for IMU

WillXuCodes:
### Summary:
Added an option in the C and C++ API for a calibration function to block while resetting instead of forcing the user to do it every time.
#### C:
- `pros::c::reset_blocking();`

#### C++:
Added a default parameter to Imu::reset that doesn't block by default, but will block if `.reset(true);` is called.
- `pros::Imu::reset(bool blocking = false);`

### Motivation:
Improve the PROS API and its usability

### References:
#398 

### Test Plan:
This should be a short hardware test:
- [x] Test on hardware, make sure both forms of reset/calibration work as intended.

### DIFF
--- a/include/pros/imu.h
+++ b/include/pros/imu.h
@@ -88,6 +88,27 @@ typedef struct __attribute__((__packed__)) euler_s {
  */
 int32_t imu_reset(uint8_t port);
 
+/**
+ * Calibrate IMU and Blocks while Calibrating
+ *
+ * Calibration takes approximately 2 seconds and blocks during this period, 
+ * with a timeout for this operation being set a 3 seconds as a safety margin.
+ * Like the other reset function, this function also blocks until the IMU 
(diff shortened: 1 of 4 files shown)


BennyBot (approved):
C API Tests.
Test:
```
pros::c::imu_reset(10);
printf("IMU Value: %lf\n", pros::c::imu_get_rotation(10));
```
Result:
`IMU Value: inf`

Test:
```
pros::c::imu_reset_blocking(10);
printf("IMU Value: %lf\n", pros::c::imu_get_rotation(10));
```
Result:
`IMU Value: -0.001618`

C++ API Tests.
Test:
```
pros::Imu imu(10);
imu.reset();
printf("IMU Value: %lf\n", imu.get_heading());
```
Result:
`IMU Value: inf`

Test:
```
pros::Imu imu(10);
imu.reset(true);
printf("IMU Value: %lf\n", imu.get_heading());
```
Result:
`IMU Value: 0.018558`

LGTM

PR #400, opened 2022-08-07
✨Add Native Motor Group Support

### LINKED ISSUE #388: Add native support for motor groups
At the current moment, most third party libraries require Okapi to be a dependency due to our lack of motor groups within the native API. One thing we should do to help foster the growth of other third party libraries outside of Okapi should be the native support of motor groups. 

This could be a PROS 3 feature, or a PROS 4 feature but I'd advocate for this being a PROS 3 feature so people like our software peeps working on ARMS can decouple Okapi.

This would most likely be a C++ only feature, due to its objected oreinted nature.

Thoughts?

noam987:
#### Summary:
Adds Motor groups to control multiple motors simultaneously.

#### Motivation:
Help foster the growth of third-party libraries by adding native support for motor groups. 

##### References (optional):
Closes #388 
#### Test Plan:
WIP

- [x] Create a motor group
- [ ] Call each setter/move method of the motor group and compare the behavior to calling each motor individually. (More detailed Code coming soon).


include/pros/motors.hpp
```diff
@@ -20,7 +20,10 @@
 #define _PROS_MOTORS_HPP_
 
 #include <cstdint>
+#include <vector>
+#include <initializer_list>
 
+#include "pros/rtos.hpp"
```
WillXuCodes, line 26: Don't think this is necessary

WillXuCodes (changes requested):
Will have more comments later

include/pros/motors.hpp
```diff
@@ -851,9 +846,482 @@ class Motor {
...
+ * Macro to free the motor group mutex with a vector of error as the error value
+ * 
+ */
+#define give_motor_group_mutex_vector(error) \
+	if (!_motor_group_mutex.give()) {					 \
+		out.clear();														 \
+		for (Motor m : _motors) {								 \
+			out.push_back(error);									 \
+		}																				 \
+	}
```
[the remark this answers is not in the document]
noam987: Moved

src/devices/vdml_motors.cpp
```diff
@@ -13,9 +13,72 @@
...
+		else {                                                    \
+			out = PROS_ERR;                                         \
+		}                                                         \
+	}
+
+/**
+ * Macro to take the motor group mutex with a vector of error as the error value
+ * 
+ */
+#define take_motor_group_mutex_vector(error) \
```
WillXuCodes, line 55: claim_mg_mutex_vector(error)

src/devices/vdml_motors.cpp
```diff
@@ -13,9 +13,72 @@
...
+		out.resize(_motors.size());							 \
+		out.shrink_to_fit();                     \
+		return out;                              \
+	}
+
+/**
+ * Macro to free the motor group mutex with a vector of error as the error value
+ * 
+ */
+#define give_motor_group_mutex_vector(error) \
```
WillXuCodes, line 70: mg_mutex_give_vector(error)

WillXuCodes (changes requested):

src/devices/vdml_motors.cpp
```diff
@@ -13,9 +13,72 @@
...
+	if (!_motor_group_mutex.give()) {            \
+		return error;                         	   \
+	}
+
+/**
+ * Macro to loop through a function call for each motor in a motor group
+ * with an error value of PROS_ERR
+ * 
+ */
+#define mg_foreach(func_call, motors)                       	\
```
WillXuCodes, line 41: instead of having "out" as something we expect, can we have this as a macro parameter?

src/devices/vdml_motors.cpp
```diff
@@ -259,6 +322,335 @@ std::int32_t Motor::set_reversed(const bool reverse) const {
 std::int32_t Motor::set_voltage_limit(const std::int32_t limit) const {
 	return motor_set_voltage_limit(_port, limit);
 }
+Motor_Group::Motor_Group(const std::initializer_list<Motor> motors) : _motors(motors), _motor_group_mutex(pros::Mutex()) {}
```
WillXuCodes, line 325: TODO: add constructor that allows an initializer list full of numbers, and reverse numbers if the port number is negative.

WillXuCodes (changes requested):
Changes requested, pls see above. (Only thing to do before 3.7.0 release)

WillXuCodes:
Another thing to do is to replace all instances of "1" being returned to be PROS_SUCCESS (i've backmerged develop so it should work at this point). Another thing is that a motor group with 0 motors should be a valid motor group, so we should be returning "1" if no work is done w/ the output.

WillXuCodes:
Also, this code was tested with hardware in its current form without the requested constructors and it worked.

WillXuCodes (approved):
Hardware Tested, LGTM

PR #410, opened 2022-09-02
🐛Add `.cache/` to .gitignore to prevent extra files from being committed

Unionjackjz1:
#### Summary:
Added `.cache/` to .gitignore

#### Motivation:
This stops a bajillion files appearing in git.


WillXuCodes (approved):
sauce

PR #427, opened 2022-10-04
✨Add ADI Addressable LED Support

### LINKED ISSUE #424: 3.7.2: Add LED support
Waiting for firmware update

omegaStag:
#### Summary:
Adds support for the VEX ADI LED

#### Motivation:
Self-explanatory

##### References (optional):
Closes #424 

#### Test Plan:
Test code is in main.cpp. Plug in LEDs to port and run.

- [ ] Test code works?


src/devices/vdml_ext_adi.c
```diff
@@ -443,3 +444,64 @@ double ext_adi_potentiometer_get_angle(ext_adi_potentiometer_t potentiometer) {
...
+	else if (buffer_length < 1)
+	{
+		buffer_length = 1;
+	}
+	uint32_t buf[buffer_length];
+	for (int i = 0; i < buffer_length; i++) {
+		buf[i] = 0;
+	}
+	return ext_adi_led_set_buffer(led, buf, buffer_length, 0);
+}
```
WillXuCodes, line 507: Add newline at end of file

src/devices/vdml_ext_adi.c
```diff
@@ -443,3 +444,64 @@ double ext_adi_potentiometer_get_angle(ext_adi_potentiometer_t potentiometer) {
...
+	uint8_t smart_port, adi_port;
+	get_ports(led, smart_port, adi_port);
+	transform_adi_port(adi_port);
+	claim_port_i(smart_port, E_DEVICE_ADI);
+	validate_type(device, adi_port, smart_port - 1, E_ADI_DIGITAL_OUT);
+
+	if(buffer == NULL) {
+        errno = EINVAL;
+        return PROS_ERR;
+    }
```
WillXuCodes, line 480: Fix tab indentation,

src/devices/vdml_ext_adi.c
```diff
@@ -443,3 +444,64 @@ double ext_adi_potentiometer_get_angle(ext_adi_potentiometer_t potentiometer) {
...
+	if (offset < 0) {
+		offset = 0;
+	}
+	else if (offset >= MAX_LED)
+	{
+		offset = MAX_LED - 1;
+	}
+	//TBD: what is the behavior of vexAdiAddrLedSet with a buffer too long (beyond INT_MAX?)
```
WillXuCodes, line 488: The buffer is simply a pointer to an address which is why we need to pass in the size (hence why the LED func requires a size), so this is not an issue.

src/devices/vdml_ext_adi.c
```diff
@@ -443,3 +444,64 @@ double ext_adi_potentiometer_get_angle(ext_adi_potentiometer_t potentiometer) {
...
+	vexAdiAddrLedSet(device->device_info, adi_port, buffer, offset, buffer_length, 0);
```
WillXuCodes, line 489: we might want to sanity check the buffer length relative to the offset too.

src/devices/vdml_adi.cpp
```diff
@@ -159,4 +159,34 @@ double ADIPotentiometer::get_angle() const {
 	return ext_adi_potentiometer_get_angle(merge_adi_ports(temp_smart, _adi_port));
 }
 
+ADILed::ADILed(std::uint8_t adi_port) : ADIPort(adi_port) {
```
WillXuCodes, line 162: For the adi, we might want to consider making the buffer something we initialize with a size, and having the buffer be a class member. This would reduce overhead and automate the buffer creation. Of course, we could always also have overloads (like you already have) for C-api arrays.

src/devices/vdml_adi.c
```diff
@@ -128,3 +128,23 @@ adi_potentiometer_t adi_potentiometer_type_init(uint8_t port, adi_potentiometer_
...
+	return (bool)ext_adi_led_get_state((ext_adi_led_t)led);
+}
+
+int32_t adi_led_set_buffer(adi_led_t led, uint32_t* buffer, uint32_t buffer_length, uint32_t offset) {
+	return ext_adi_led_set_buffer((ext_adi_led_t)led, buffer, buffer_length, offset);
+}
+
+int32_t adi_led_clear_buffer(adi_led_t led, uint32_t buffer_length) {
+	return ext_adi_led_clear_buffer((ext_adi_led_t)led, buffer_length);
+}
```
WillXuCodes, line 150: newline at end of file required

src/devices/vdml_adi.c
```diff
@@ -128,3 +128,23 @@ adi_potentiometer_t adi_potentiometer_type_init(uint8_t port, adi_potentiometer_
 double adi_potentiometer_get_angle(adi_potentiometer_t potentiometer) {
 	return ext_adi_potentiometer_get_angle((ext_adi_potentiometer_t)potentiometer);
 }
+
+adi_led_t adi_led_init(uint8_t port) {
+	return (adi_led_t)ext_adi_led_init(INTERNAL_ADI_PORT, port);
+}
+
+int32_t adi_led_set_state(adi_led_t led, bool value) {
```
WillXuCodes, line 136: can we change this to turn on/turn off? This would be more intuitive to the user

(or turn_on(adi_led_t led) and turn_off(adi_led_t led) and turn(adi_led_t led, bool on) )

WillXuCodes (changes requested):
See above

Richard-Stump:
For the C++ api, couldn't we have the led object automatically manage the buffer? When the ADILED object is created, the class can automatically generate an RGB buffer from the number of LEDS the strip has, then the we can have an overload that lets the user access individual elements of the led strip:
```cpp
pros::ADILed ledStrip(port, num_leds);      // Create the led strip object
ledStrip[0] = 0xFF00FF;                              // Set the first pixel to purple
ledStrip.update();                                       // Send the rgb values to the strip
```
This would be safer for the end user (which is more C++ like), and easier. Would could probably also provide functions for setting a range of pixels to the same value.

SylvieMayer (commented):
`ext_adi_led_init()` appears to work perfectly fine, but `ext_adi_led_set()` and `ext_adi_led_set_pixel()` do not. 

I'm getting weird issues with program hanging, it's not making sense. 

I can get lights to turn on by calling `ext_adi_led_init()` with 1-indexed port numbers, then using the SDK functions directly to set the lights to on.

WillXuCodes (approved):
As shown in a previous PR, this seems to work on hardware now.

SylvieMayer:
AWESOME

PR #438, opened 2022-10-10
🐛 Add ext_ to missing CPP functions for adi C++ 

WillXuCodes:
#### Summary:
Says what it does

#### Test Plan:
See if this was an issue in the first place with effected functions, and then see if this PR changes anything.

### DIFF
--- a/src/devices/vdml_adi.cpp
+++ b/src/devices/vdml_adi.cpp
@@ -102,11 +102,11 @@ ADIEncoder::ADIEncoder(ext_adi_port_tuple_t port_tuple, bool reversed) : ADIPort
 }
 
 std::int32_t ADIEncoder::reset() const {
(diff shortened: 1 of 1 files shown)


WillXuCodes:
> I have zero idea if this is helpful or means anything, but I noticed that in `vdml_ext_adi.c` that the three wire devices that `return_port(smart_port...)` instead of `return_port(smart_port - 1...)` are the same ones that you have changed things with here. Encoder, Gyro, Ultrasonic.

yeah it's an inconsistent thing which makes adding stuff to ADI (like you saw in LED) a little bit of a pain that I've already opened an issue for in #440

Richard-Stump:
Tested with single motors and motor groups with various reversed/non-reversed combinations and everything seems to work.

PR #441, opened 2022-10-11
✨Add MotorGroup alias for Motor_Group

evanmiller620:
#### Summary:
Added one line that hopefully does what it is meant to do

#### Motivation:
Was my first project

#### Test Plan:
Try using MotorGroup and see if it functions as Motor_Group

- [x] test item

### DIFF
--- a/include/pros/motors.hpp
+++ b/include/pros/motors.hpp
@@ -842,7 +842,7 @@ class Motor {
 	 * ENODEV - The port cannot be configured as a motor
 	 *
 	 * \return The motor's voltage limit in V or PROS_ERR if the operation failed,
-	 * setting errno.Motor_Group
+	 * setting errno.
 	 */
 	virtual std::int32_t get_voltage_limit(void) const;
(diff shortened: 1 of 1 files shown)


include/pros/motors.hpp
```diff
@@ -842,7 +842,7 @@ class Motor {
 	 * ENODEV - The port cannot be configured as a motor
 	 *
 	 * \return The motor's voltage limit in V or PROS_ERR if the operation failed,
-	 * setting errno.
+	 * setting errno.Motor_Group
```
WillXuCodes, line 845:
```suggestion
	 * setting errno.
```
accidental I presume?

WillXuCodes (changes requested):

include/pros/motors.hpp
evanmiller620: Oops yes I'll fix that

WillXuCodes (approved):

PR #442, opened 2022-10-11
Fixed C and C++ addrled APIs so that they actually work now

SylvieMayer:
Fixed port decrement/increment issue in the c api's init and set functions, as well as the c++ api's addrled ctor

fixed issue with ext_adi_led_set always returning PROS_ERR because (... || buffer_length) is always true

tested changes on actual hardware to make sure they work

#### Summary:
There were several issues with the C and C++ addrled API still. 

The primary issue was that in the C API initialization function, the smart port number needed to be decremented by one. 

```cpp
return_port(smart_port - 1, merge_adi_ports(smart_port, adi_port + 1));
```
changed to 
```cpp
return_port(smart_port - 1, merge_adi_ports(smart_port - 1, adi_port+1));
```

Now that `ext_adi_led_t` has the correct ports itself, I could fix the increment/decrement issues in the set functions (both `ext_adi_led_set` and `ext_adi_led_set_pixel`

```cpp
validate_type(device, adi_port, smart_port - 1, E_ADI_DIGITAL_OUT);
```
changed to 
```cpp
claim_port_i(smart_port, E_DEVICE_ADI);
validate_type(device, adi_port, smart_port, E_ADI_DIGITAL_OUT); 
```

I also changed the same thing in `return_port`
```cpp
return_port(smart_port - 1, rtv); 
```
changed to 
```
return_port(smart_port, rtv); 
```

In addition, in `ext_adi_led_set` there was a completely unrelated issue with this if/else structure

```cpp
if (buffer_length > MAX_LED) {
	buffer_length = MAX_LED;
}
else if (buffer == NULL || buffer_length < 1 || buffer_length)
{
	errno = EINVAL;
	return PROS_ERR;
}
```

If `buffer_length` isn't greater than `MAX_LED` (bad for other reasons), it goes to evaluate `(buffer == NULL || buffer_length < 1 || buffer_length)`. buffer_length is an integer, so it will always return true. This meaning that if the `buffer_length` is valid in the first place, `ext_adi_led_set` will ALWAYS return `PROS_ERR`. 

This was the extent of the fixes to the C API, there was another related issue in the C++ API. 

```cpp
ADILed::ADILed(ext_adi_port_pair_t port_pair, std::uint32_t length) : ADIPort(std::get<1>(port_pair)) {
	std::int32_t _port = ext_adi_led_init(port_pair.first, port_pair.second);
	get_ports(_port, _smart_port, _adi_port);
	_smart_port++; // for inherited functions this is necessary
	if (length < 1) {
		length = 0;
	}
	if (length > MAX_LED) {
		length = MAX_LED;
	}
	_buffer.resize(length, 0);
}
```
changed to 
```cpp
ADILed::ADILed(ext_adi_port_pair_t port_pair, std::uint32_t length) : ADIPort(std::get<1>(port_pair)) {
	std::int32_t _port = ext_adi_led_init(port_pair.first, port_pair.second);
	get_ports(_port, _smart_port, _adi_port);
	if (length < 1) {
		length = 0;
	}
	if (length > MAX_LED) {
		length = MAX_LED;
	}
	_buffer.resize(length, 0);
}
```

Removing the increment of `_smart_port`. 

Completely unrelated to all above issues, there was also a problem in `ext_adi_led_set_all`. 

```cpp
int32_t ext_adi_led_set_all(ext_adi_led_t led, uint32_t* buffer, uint32_t buffer_length, uint32_t color) {
	memset(buffer, color, buffer_length * sizeof(uint32_t));
	return ext_adi_led_set(led, buffer, buffer_length);
}
```

`memset` tries to set every value of the buffer to a single provided byte, but we're passing it a `uint32_t`, which is 4 bytes long. This is a problem. I fixed this problem by replacing `memset` with a for loop. 

```cpp
int32_t ext_adi_led_set_all(ext_adi_led_t led, uint32_t* buffer, uint32_t buffer_length, uint32_t color) {
	for(int i = 0;i<buffer_length;i++){
      buffer[i] = color;
  	}
	return ext_adi_led_set(led, buffer, buffer_length);
}
```

#### Motivation:
I really really really want to be able to use my addrleds 🥺

#### Test Plan:
- [ ] Test C++ API with a simple demo project
```cpp
void opcontrol() {
    pros::ADILED led1 = pros::ADILED({1,1}, 8);
    pros::delay(200); // Delay to give ports time to configure. THIS IS NEEDED!!!!!! don't know how long though
    uint32_t color = 0xB00B69;
    led1.set_all(color);
    led1.update();
}
```
image
This results in the first 8 LEDs on the strip plugged into {smart port 1, ADI port A} turning a nice bright magenta color, as expected.

- [ ] Test C API with a simple demo project
```cpp
void opcontrol() {
    uint32_t maxled = 8;
    uint32_t ledbuffer[64];
    for(int i = 0;i<maxled;i++){
        ledbuffer[i] = 0xBADA55;
    }
    pros::c::ext_adi_led_t led1 = pros::c::ext_adi_led_init(1,1);
    pros::delay(200); // Delay to give ports time to configure. THIS IS NEEDED!!!!!! don't know how long though
    pros::c::ext_adi_led_set(led1, ledbuffer, 8);
}
```

image
This again results in exactly what we expect, the first 8 LEDs turning on to a bright, light green.


version
```diff
@@ -1 +1 @@
-3.7.1
+3.7.2-dirty.37.b6a7853
```
ayushshukla0, line 1: Revert this also

src/devices/vdml_ext_adi.c
```diff
@@ -457,44 +458,46 @@ ext_adi_led_t ext_adi_led_init(uint8_t smart_port, uint8_t adi_port) {
 	transform_adi_port(adi_port);
 	claim_port_i(smart_port - 1, E_DEVICE_ADI);
 	vexDeviceAdiPortConfigSet(device->device_info, adi_port, (V5_AdiPortConfiguration)E_ADI_DIGITAL_OUT); 
-	return_port(smart_port - 1, merge_adi_ports(smart_port, adi_port + 1));
+	return_port(smart_port - 1, merge_adi_ports(smart_port - 1, adi_port+1));
```
WillXuCodes, line 461:
```suggestion
	return_port(smart_port - 1, merge_adi_ports(smart_port - 1, adi_port + 1));
```
Format fix

src/devices/vdml_ext_adi.c
```diff
@@ -457,44 +458,46 @@ ext_adi_led_t ext_adi_led_init(uint8_t smart_port, uint8_t adi_port) {
...
 	get_ports(led, smart_port, adi_port);
-	claim_port_i(smart_port - 1, E_DEVICE_ADI);
 	transform_adi_port(adi_port);
-	validate_type(device, adi_port, smart_port - 1, E_ADI_DIGITAL_OUT);//subtracted 1 from smart_port
+	claim_port_i(smart_port, E_DEVICE_ADI);
+	validate_type(device, adi_port, smart_port, E_ADI_DIGITAL_OUT);
 	if (buffer_length > MAX_LED) {
 		buffer_length = MAX_LED;
 	}
-	else if (buffer == NULL || buffer_length < 1 || buffer_length)
```
WillXuCodes, line 472: good catch lmao... wtf was I thinking 🤦

version
WillXuCodes, line 1: revert pls
```suggestion
3.7.1

```

include/api.h
```diff
@@ -41,8 +41,8 @@
 
 #define PROS_VERSION_MAJOR 3
 #define PROS_VERSION_MINOR 7
-#define PROS_VERSION_PATCH 1
-#define PROS_VERSION_STRING "3.7.1"
+#define PROS_VERSION_PATCH 2
+#define PROS_VERSION_STRING "3.7.2-dirty.37.b6a7853"
```
WillXuCodes, line 45:
```suggestion
#define PROS_VERSION_PATCH 1
#define PROS_VERSION_STRING "3.7.1"
```
revert pls

src/devices/vdml_ext_adi.c
```diff
@@ -457,44 +458,46 @@ ext_adi_led_t ext_adi_led_init(uint8_t smart_port, uint8_t adi_port) {
...
 	buffer[pixel_position] = color;
 	uint32_t rtv = (uint32_t)vexDeviceAdiAddrLedSet(device->device_info, adi_port, buffer, 0, buffer_length, 0);
 	return_port(smart_port - 1, rtv); 
 }
 
 int32_t ext_adi_led_set_all(ext_adi_led_t led, uint32_t* buffer, uint32_t buffer_length, uint32_t color) {
-	memset(buffer, color, buffer_length * sizeof(uint32_t));
+	for(int i = 0;i<buffer_length;i++){
+      buffer[i] = color;
+  	}
```
WillXuCodes, line 500: the idea behind doing a memset instead of a for loop is because memset is much more atomic than a for loop... are we sure memset does not work here and what was the issue with it?

ayushshukla0 (changes requested):
We are trying to become more strict with formatting but its not up yet. Thanks for the contribution but make sure to follow the formatting rules in the `.clang-format` file or even better run the formatter in vsc using alt+shift+f (option+shift+f on mac) which will do the formatting for you following the rules we set in `.clang-format`.

WillXuCodes (changes requested):

src/devices/vdml_ext_adi.c
```diff
@@ -13,6 +13,7 @@
 #include <errno.h>
 #include <math.h>
 #include <stdio.h>
+#include <string.h>
```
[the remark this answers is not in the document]
SylvieMayer: I don't think so, I believe vscode added that for some silly reason by itself. It can be removed.

include/api.h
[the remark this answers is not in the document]
SylvieMayer: yes

src/devices/vdml_ext_adi.c
SylvieMayer: I tried playing with it quite a bit and it didn't work. 

fwiw, jpearman does it this same way with for loops in the vexcode implementation of addrled.

SylvieMayer:
I think I fixed all the formatting issues?

ayushshukla0:
> I think I fixed all the formatting issues?

still some things left check the file diff

SylvieMayer:
> > I think I fixed all the formatting issues?
> 
> still some things left check the file diff

good now, i think?

ayushshukla0:
Indent it one more. I suggest running a formatter instead.

ayushshukla0 (approved):
LGTM

WillXuCodes (approved):

PR #443, opened 2022-10-12
✨ Add Reverseable Motors to Default API

R11G:
#### Summary:
Change Motor constructors' port param from uint8_t to int8_t
Add functionality for users to input a negative port into Motor constructors to reverse motors at that port. 

#### Test Plan:
See if motors spin in reverse after passing in a negative port parameter.

### DIFF
--- a/include/pros/motors.hpp
+++ b/include/pros/motors.hpp
@@ -46,16 +46,16 @@ class Motor {
 	 * \param encoder_units
 	 *        The motor's encoder units
 	 */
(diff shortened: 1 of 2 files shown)


WillXuCodes:
Could you also refractor this to motor groups as well? You can see what I mean if you go down to the motor group constructors.

Richard-Stump (approved):
Tested with both motors and motor groups. Motor groups were tested with various combinations of reversed and non-reversed motors and everything seems to work as expected.

PR #449, opened 2022-10-19
🐛 Fix global pros::Rotation objects creating a data abort exception

Richard-Stump:


WillXuCodes:
Status: need to find out the issue with the init function instead of changing the constructor to what it is right now.

WillXuCodes (approved):
honestly this is a good bandaid solution... we need to figure out what the real underlying issue is sometime though.

PR #451, opened 2022-10-25
🔥PROS 4: Remove deprecated motor functions

AndrewChen12:
#### Summary:
Removed deprecated motor functions

#### Motivation:
Cleaning out code that should no longer be used

#### Test Plan:
Try to run deprecated functions, should give errors


noam987 (approved):
LGTM

noam987:
Hardware tested. Code does not compile if using deprecated functions. It does if deprecated functions are not called

Issue #428, opened 2022-10-04
PROS 4: Add overloads for std::ostream& operator<< (std::ostream&, ...) for all devices 


Richard-Stump:
Add operator overloads that allow all devices to be used with C++ ostreams for debugging. This would allow users to print the state of devices easily by doing:
```cpp
pros::Motor myMotor(...);
std::cout << motor << std::endl;
```

PR #452, opened 2022-10-25
✨Add Stream Output Overload for Multiple Devices

Gracelu128:
#### Summary:
Add print op overload for gps, rotation, motors, distance, optical, imu, and adi.

#### Motivation:
The print op overload will print out an object and help with debugging

##### References (optional):
#428 

#### Test Plan:
-Create objects and see if they print the correct results


WillXuCodes:
Hardware testing updated: Motors do not work in terms of printing out anything. We need to check if this is a blocking issue, or if we're writing too much to our stdout buffer. I think we might also need to consider if the std::endl's we put here are causing the issue too.

WillXuCodes (changes requested):
remove .DS_Store from this commit please, add it to .gitignore for us too.

WillXuCodes (approved):
LGTM

PR #459, opened 2022-11-02
🐛 Fix bug with LED non-extended constructor

WillXuCodes:
#### Summary:
Says what it does in the title

#### Test Plan:
- Try using this constructor and initializing/using a LED strip from the lab and seeing if it works. 

Props to @SylvieMayer for assisting in finding this.


WillXuCodes:
Hardware testing apparently failed, need to go figure out the real culprit here.

WillXuCodes (commented):
https://user-images.githubusercontent.com/54247087/205020112-bfcf539b-4881-48cc-95fa-9457c5cba420.mov

W H Y  D O E S 
T H IS 

WO R K

Issue #450, opened 2022-10-20
Add the array overload to PROS 3


Richard-Stump:
This is a feature we would like to implement in PROS4; however, MotorGroups have yet to be added to Pros 4, so they cannot be implemented to Pros 4. Since this feature is a small one that could be very useful for MotorGroups, it can be implemented into Pros 3 and easily ported into Pros 4. This would be fine to do since it is a small quality of life improvement, and we have #449 to push out when we get that fixed.

The gist of this feature is that if the user has a MotorGroup, they should be able to treat it like an array of motors:
```cpp
pros::MotorGroup group({12, 13, -2, 3});

pros::Motor& motorA = group[3]; // motorA now references the last motor in the group.
pros::Motor& motorB = group[0]; // motorB mow references the first motor in the group.
```
The index passed should simply index into the list of motors that the motor groups.

Additionally, MotorGroups should have a method, `size()` which returns the number of motors in the group.

These features would allow for easy iteration over a motor group with a standard for loop, and without having to use many vectors for each of the properties needed.
```cpp
for(int i = 0; i < group.size(); i++) {
      group[i].set_reversed(true);
      group[i].set_velocity(12);
}
```
This is redundant to be used like this, but this could maybe be used in more useful ways.

noam987:
closed in #463

PR #463, opened 2022-11-10
✨Added functionality for size() and [] operator

AlexHunton2:
#### Summary:
Added functionality for MotorGroups to be treated more like arrays than classes. With these changes the user should be able to index a MotorGroup similar to an array, and get its size with a size() method. This should allow MotorGroups to be looped through.

#### Motivation:
#450 

#### Test Plan:
Requires hardware testing. Asked to add this PR ASAP in order to keep ahead on schedule. Will hardware test and add information about that here when completed.

### DIFF
--- a/include/pros/motors.hpp
+++ b/include/pros/motors.hpp
@@ -1026,7 +1026,7 @@ class Motor_Group {
 	 * \param i
 	 *        The index value in the motor group.
 	 *
-	 * \return the appropriate Motor reference, return null and set errno on error.
+	 * \return the appropriate Motor reference, return first motor and set errno on error.
 	 */ 
 	pros::Motor& operator[](int i);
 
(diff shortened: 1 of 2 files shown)


include/pros/motors.hpp
```diff
@@ -1020,6 +1020,25 @@ class Motor_Group {
...
+
+
+	/**
+	 * Indexes Motor in the Motor_Group in the same way as an array.
+	 * 
+	 *
+	 * \return the size of the vector containing motors
+	 */ 
+	std::int32_t size();
+
```
WillXuCodes, line 1041: Please add proper ERRNO documentation and other relevant information as seen in other comments.

WillXuCodes (changes requested):

noam987:
Passed hardware testing

WillXuCodes (approved):

Issue #455, opened 2022-10-31
PROS 4: Add operator=() overload to classes


Richard-Stump:
Since we are making PROS 4 more object oriented, it was pointed out that implementing the assignment operator on the V5 devices would be useful. This operator would be used like:
```cpp
pros::Motor motor1(12);
pros::Motor motor2 = motor1;
```
This makes a copy of the original motor. The second motor has the same port, and copies settings such as reversed, gearset, etc. Here is an example implementation for the V5 motor:

```cpp
pros::Motor& pros::Motor::operator=(const pros::Motor& other) {
  _port = other._port;
  _reversed = other._reversed;
  return *this;
}
```

PR #468, opened 2022-11-16
✨ PROS 4: Motor '=' Operator Copy Constructor

aaryan-gautam:
#### Summary:

Added operator overloading for motor object.
#### Motivation:

These changes will improve user experience by making it easier to assign the properties of a device object to other device objects. 
##### References (optional):

#455 
#### Test Plan:

Initialize a motor object and then initialize another motor object by assigning it to the other object.

- [ ] test item
`pros::Motor motor1(12);
pros::Motor motor2 = motor1;`

### DIFF
--- a/src/devices/vdml_motors.cpp
+++ b/src/devices/vdml_motors.cpp
@@ -298,6 +298,14 @@ std::ostream& operator<<(std::ostream& os, const pros::Motor& motor) {
 	return os;
 }
 
+pros::Motor& pros::Motor::operator=(const pros::Motor& other) {
+  _port = other._port;
+  _reverse = other._reverse;
+  _gearset = other._gearset;
+  _encoder_units = other._encoder_units;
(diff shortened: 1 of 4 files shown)


src/devices/vdml_motors.cpp
```diff
@@ -298,6 +298,14 @@ std::ostream& operator<<(std::ostream& os, const pros::Motor& motor) {
 	return os;
 }
 
+pros::Motor& pros::Motor::operator=(const pros::Motor& other) {
+  _port = other._port;
+  _reverse = other._reverse;
+  _gearset = other._gearset;
+  _encoder_units = other._encoder_units;
+  return *this;
+}
```
WillXuCodes, line 307: Since this is a setter... doesn't this need to be mutex protected?

Richard-Stump (approved):

PR #470, opened 2022-11-24
✨ PROS 3.8: Add additional ctors to pros::Motor_Group, refine existing ctors

djava:
#### Summary:
1. Adds the following functions:
  a. `Motor_Group::Motor_Group(const std::vector<pros::Motor>& motors)`
  b. `Motor_Group::Motor_Group(const std::initializer_list<std::int8_t> motor_ports)`
2. Removes the `explicit` tag from the `Motor_Group(const std::initializer_list<Motor>)` ctor
3. Changes the `Motor_Group(const std::vector<std::int8_t>)` ctor to `Motor_Group(const std::vector<std::int8_t>&)` (passes by const-ref instead of by value)
4. Changes the definition of the `Motor_Group(const std::vector<std::int8_t>&)` to use a range-based `for` loop and `std::vector::emplace_back` instead of `std::vector::push_back`

#### Motivation:
1. a. Allowing the user to pass a `std::vector<Motor>&` removes the restriction that `Motor_Group`s could only be constructed with literal brace-enclosed lists of `Motor`s that existed because `std::initializer_list` is designed to exist only as an rvalue.  Now the options for constructing a `Motor_Group` with `Motor`s are far more flexible.
  b. Allows for a more efficient construction of `Motor_Groups` using a literal brace-enclosed list of port numbers than only accepting `std::vector<uint8_t>` for ports
2. I see no real reason that a brace-enclosed list of `Motor` shouldn't be implicitly convertible to `Motor_Group`.  This change seems like it could create more ergonomic function calls in places, without any real downsides.
3. Avoids the copy, just more efficient and idiomatic with no downsides.
4. `emplace_back(port)` constructs the `Motor`s in-place inside the vector instead of creating temporary `Motor`s and copying them into the vector, as `push_back(Motor(port))` does.  Also uses range-based `for` loop to be more idiomatic.
 
##### References (optional):
Original inspiration for 1: https://discord.com/channels/197777408198180864/198658294007463936/1044109726675390544

#### Test Plan:
- [x] Make sure the kernel compiles and `Motor_Group`s work as expected.  I don't foresee any issues with functionality as long as it compiles.


version
```diff
@@ -1 +1 @@
-3.7.2
+3.2.2-commit.140.b259f7f
```
WillXuCodes, line 1: Mind reverting this? 
```suggestion
3.7.2

```

include/api.h
```diff
@@ -40,9 +40,9 @@
 #endif /* __cplusplus */
 
 #define PROS_VERSION_MAJOR 3
-#define PROS_VERSION_MINOR 7
+#define PROS_VERSION_MINOR 2
 #define PROS_VERSION_PATCH 2
-#define PROS_VERSION_STRING "3.7.2"
+#define PROS_VERSION_STRING "3.2.2-commit.140.b259f7f"
```
WillXuCodes, line 45: Revert

include/pros/motors.hpp
```diff
@@ -859,8 +859,11 @@ class Motor {
 
 class Motor_Group {
 	public:
-	explicit Motor_Group(const std::initializer_list<Motor> motors);
-	explicit Motor_Group(const std::vector<std::int8_t> motor_ports);
+			 Motor_Group(const std::initializer_list<Motor> motors);
```
WillXuCodes, line 862:
```suggestion
	Motor_Group(const std::initializer_list<Motor> motors);
```

WillXuCodes (changes requested):
Looks good, just some formatting fixes please.

djava:
Done

Richard-Stump:
@djava We don't think this can get merged due to breaking changes in the constructors.
image
The constructors that take vectors have been changed to take references to vectors, and the constructor that takes an initializer list of motors is now non-explicit. 

These changes are good fine PROS 4, but we cannot accept these into PROS 3 since they break the ABI. This will break any templates that are not compiled against newer kernels.

djava:
I will revert that one to passing by value

Richard-Stump:
> I will revert that one to passing by value

Thank you. We will make them references in PROS 4 though.

djava:
That was a lot of commits to do that 🥴 oops

Richard-Stump:
Looks good! I'm still trying to figure out if the `explicit` keyword breaks ABI or not.

Richard-Stump:
I will try to apply this kernel to an out of date template to ensure that these changes aren't ABI breaking.

djava:
> Looks good! I'm still trying to figure out if the `explicit` keyword breaks ABI or not.

I'm not certain but I find it hard to imagine that it would

Richard-Stump:
Yeah, I don't imagine it would, but we didn't think that switching the motor's constructors to signed would break much, but it did.

WillXuCodes (approved):
Proven to not break ABI. Approved

PR #480, opened 2022-12-16
🐛Make display_initialize() a weak symbol for the liblvgl template

Richard-Stump:
This is a simple fix that makes display_initilialize() a weak symbol so that the implementation in the liblvgl repo can be used. 

The default implementation does not do anything. If the liblvgl template is present, the implementation in that template will be used.

### DIFF
--- a/src/system/startup.c
+++ b/src/system/startup.c
@@ -22,7 +22,7 @@ extern void rtos_initialize();
 extern void vfs_initialize();
 extern void system_daemon_initialize();
 extern void graphical_context_daemon_initialize(void);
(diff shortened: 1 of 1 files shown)


Richard-Stump:
Branched from

WillXuCodes:
Does this change actually work for fixing it? Have you tested it?

Richard-Stump:
It stops the black screen issue, and now gives a white screen. I made this change because I knew it would need to be done to get everything sorted out. I'm currently working on the changes to get LLEMU working again.

Issue #473, opened 2022-12-02
✨Add ability to change optical sensor integration time


BennyBot:
#### Expected Behavior:
https://www.vexforum.com/t/v5-optical-sensor-refresh-rate/109632/4?u=bennyboy

#### Actual Behavior:
We are unable to do this in PROS

PR #481, opened 2022-12-17
✨ Add Integration Time Setting/Getting to Optical

WillXuCodes:
#### Summary:
Added Integration Time Setting/Getting to Optical, C and C++, bounds are 712 ms max and 3 ms min. 

#### References:
#473 

#### Test Suace:
- [x] Set Integration to something reasonable (20 ms)
- [x] Set Integration to something too low (2 ms)
- [x] Set Integration to something reasonable (713 ms)
- [x] Get Integration time before setting sensor (no set)
- [x] Get Integration time after setting sensor (20 ms set)
- [x] Drink ranch

### DIFF
--- a/include/pros/optical.h
+++ b/include/pros/optical.h
@@ -265,6 +265,39 @@ int32_t optical_enable_gesture(uint8_t port);
  */
 int32_t optical_disable_gesture(uint8_t port);
 
+/**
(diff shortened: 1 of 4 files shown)


BennyBot (approved):
Tested on hardware and appears to be working.

PR #485, opened 2023-01-12
✨ PROS 4: Motor '=' Operator Copy Constructor

Richard-Stump:
Summary:
Added operator overloading for motor object.

Motivation:
These changes will improve user experience by making it easier to assign the properties of a device object to other device objects.

References (optional):
https://github.com/purduesigbots/pros/issues/455

Test Plan:
Initialize a motor object and then initialize another motor object by assigning it to the other object.

 test item
pros::Motor motor1(12); pros::Motor motor2 = motor1;

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -12,7 +12,7 @@ void on_center_button() {
 	if (pressed) {
 		pros::lcd::set_text(2, "I was pressed!");
 	} else {
-		pros::lcd::clear_line(2);
+		pros::lcd::clear_line(2); 
 	}
 }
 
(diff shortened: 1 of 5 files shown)


src/main.cpp
```diff
@@ -12,7 +12,7 @@ void on_center_button() {
 	if (pressed) {
 		pros::lcd::set_text(2, "I was pressed!");
 	} else {
-		pros::lcd::clear_line(2);
+		pros::lcd::clear_line(2);
```
WillXuCodes, line 15: h-why

Richard-Stump:
Closed because the default copy assignment works with the object oriented feature being merged.

Lost my sanity trying to figure out C++ semantics before Will found this out.

Issue #466, opened 2022-11-15
PROS 4: Add ADIPneumatics Class


Richard-Stump:
Since Pneumatics have become so widespread for competitions, it would be nice to have a dedicated class to them. It would essentially be like the ADIDigitalOut class, but with different names for the functions and keep track of the state of the piston. 

## Constructors:
```
ADIPneumatic(int adi_port, bool initial_state = false);
ADIPneumatic(pros::port_pair pair, bool initial_state = false);
```

## Methods:
* `void extend()` - Extends the piston, if not done already
* `void retract()` - Retracts the piston, if not done already
* `void toggle()` - Puts the piston into the opposite state. If it is retracted, it extends, and vice versa. 
* `bool getState()` - Returns whether the piston is extended or not

PR #486, opened 2023-01-13
✨ PROS 4: Add ADIPneumatics Class

Gracelu128:
#### Summary:
Add the Pneumatics Class. Functions in this class are extend(), retract(), toggle(), and get_state(). 

extend() - Extends the piston, if not already extended.
retract() - Retracts the piston, if not already retracted.
toggle() - Puts the piston into the opposite state of its current state.
get_state() - Returns the current state of the piston.

#### References:
#466

#### Motivation:
Pneumatics class will help adjust and know the state of the piston.

#### Test Plan:
Test the functions of the class.

- [x] extend() 
- [x] retract()
- [x] toggle()
- [x] get_state()

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -11,7 +11,7 @@ SPACE := $() $()
 COMMA := ,
 
 C_STANDARD?=gnu11
-CXX_STANDARD?=gnu++20
+CXX_STANDARD?=gnu++2a
 
 DEPDIR := .d
 $(shell mkdir -p $(DEPDIR))
--- a/common.mk
+++ b/common.mk
@@ -11,7 +11,7 @@ SPACE := $() $()
 COMMA := ,
 
 C_STANDARD?=gnu11
-CXX_STANDARD?=gnu++2a
+CXX_STANDARD?=gnu++20
 
 DEPDIR := .d
 $(shell mkdir -p $(DEPDIR))
--- a/include/api.h
+++ b/include/api.h
@@ -39,10 +39,10 @@
 #include <unistd.h>
 #endif /* __cplusplus */
 
-#define PROS_VERSION_MAJOR 4
-#define PROS_VERSION_MINOR 0
-#define PROS_VERSION_PATCH 0
-#define PROS_VERSION_STRING "4.0.0"
+#define PROS_VERSION_MAJOR 3
+#define PROS_VERSION_MINOR 7
+#define PROS_VERSION_PATCH 3
+#define PROS_VERSION_STRING "3.7.3"
 
 #include "pros/adi.h"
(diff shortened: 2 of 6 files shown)


Richard-Stump (commented):
These changes look good to me. If it works on an actual robot when tested, we can get his approved.

Richard-Stump:
I tested it, and the `toggle()` and `get_state()` methods work as expected. The extend() and retract() functionality didn't work as expected, but I pushed a change to try and fix that. I'm not too convinced that this will wok on pneumatics with wiring different from the bot I tested. I think this needs more testing.

src/devices/vdml_adi.cpp
```diff
@@ -211,6 +211,33 @@ double Potentiometer::get_angle() const {
 	return ext_adi_potentiometer_get_angle(merge_adi_ports(temp_smart, _adi_port));
 }
 
+Pneumatics::Pneumatics(std::uint8_t adi_port, bool initial_state) 
+: DigitalOut(adi_port), initial_state(initial_state), state(initial_state) {
+}
+
+Pneumatics::Pneumatics(ext_adi_port_pair_t port_pair, bool initial_state) 
+: DigitalOut(std::get<1>(port_pair)), initial_state(initial_state), state(initial_state) {
```
Richard-Stump, line 219: This does not allow users to use the pneumatics with the ADI port expander. This should just call the constructor for the DigitalOut by passing port_pair rather than getting the 2nd element in the pair.

Richard-Stump (changes requested):

common.mk
```diff
@@ -11,7 +11,7 @@ SPACE := $() $()
 COMMA := ,
 
 C_STANDARD?=gnu11
-CXX_STANDARD?=gnu++20
+CXX_STANDARD?=gnu++2a
```
WillXuCodes, line 14: revert pls

include/api.h
```diff
@@ -39,10 +39,10 @@
...
 #endif /* __cplusplus */
 
-#define PROS_VERSION_MAJOR 4
-#define PROS_VERSION_MINOR 0
-#define PROS_VERSION_PATCH 0
-#define PROS_VERSION_STRING "4.0.0"
+#define PROS_VERSION_MAJOR 3
+#define PROS_VERSION_MINOR 7
+#define PROS_VERSION_PATCH 3
+#define PROS_VERSION_STRING "3.7.3"
```
WillXuCodes, line 45: not sure if we need to revert this tbh since it's more accurate lmao

WillXuCodes (changes requested):

PR #488, opened 2023-01-13
⬆️ Change makefiles to use 'pros' instead of 'prosv5'

Richard-Stump:
#### Summary:
This PR changes the makefiles to use the command `pros` instead of `prosv5`

#### Motivation:
When permissions issues arise on *nix systems, users have to chmod `pros` since that's what people typically use and `prosv5` since that is what the build system uses. This change would remove the need to set permissions on the latter. Additionally, we do not support the cortex anymore, so there is little reason for the build system to use `prosv5`

#### Test Plan:
- [ ] Ensure the `pros make` still works
- [ ] Ensure that `pros make template` still works
- [ ] Ensure that `pros make clean` still works

I have tested the various commands regarding the make command, and they still work on my system.


omegaStag:
image

image

image

all three commands work on wsl

WillXuCodes (approved):

PR #492, opened 2023-01-20
✨ PROS 4: Add Virtual Device Class 

evanmiller620:
#### Summary:
Created a virtual device class with two functions: get_port() and get_type(). Which returns the port and the device type of the smart device, respectively. 

#### Motivation:
Groups functions that are in all smart devices so that they inherit from the Device class.

#### Test Plan:
Passed hardware tests:)


include/pros/distance.hpp
```diff
@@ -127,8 +121,13 @@ class Distance {
      */
 	friend std::ostream& operator<<(std::ostream& os, pros::Distance& distance);
 
+	/**
+     * Returns the type of device
```
Richard-Stump, line 125: This is indented using spaces. Please fix to be indented with tabs instead

include/api.h
```diff
@@ -39,10 +39,10 @@
...
 #endif /* __cplusplus */
 
-#define PROS_VERSION_MAJOR 4
-#define PROS_VERSION_MINOR 0
-#define PROS_VERSION_PATCH 0
-#define PROS_VERSION_STRING "4.0.0"
+#define PROS_VERSION_MAJOR 3
+#define PROS_VERSION_MINOR 7
+#define PROS_VERSION_PATCH 2
+#define PROS_VERSION_STRING "3.7.2-commit.96.ffd0367"
```
Richard-Stump, line 45: Please change the version to 3.7.3 and the version string to match

Richard-Stump (changes requested):

WillXuCodes (changes requested):

WillXuCodes (approved):
everything ready to go.

PR #499, opened 2023-01-31
✨Adding ctors to Motor_Group

aaryan-gautam:
#470

### DIFF
--- a/include/pros/motors.hpp
+++ b/include/pros/motors.hpp
@@ -859,8 +859,10 @@ class Motor {
 
 class Motor_Group {
 	public:
-	explicit Motor_Group(const std::initializer_list<Motor> motors);
-	explicit Motor_Group(const std::vector<std::int8_t> motor_ports);
+	Motor_Group(const std::initializer_list<Motor> motors);
+	explicit Motor_Group(const std::vector<pros::Motor>& motors);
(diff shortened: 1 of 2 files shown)


djava:
How is this different from my pr?

WillXuCodes:
Be sure to credit @djava as a co-author, I think he's just opening this on PROS 4 instead of PROS 3.

WillXuCodes:
also... needs doxygen comments

djava:
Pros 4 motor groups are totally different though, shouldn't that just go in #489?

WillXuCodes:
> Pros 4 motor groups are totally different though, shouldn't that just go in #489?

yeah after #489 gets merged we'll actually take a look at this, ideally this work should've been done on that PR branch but this works too.

noam987:
This is for PROS 3.8 This Is taking @djava's PR and ensuring it is fully backwards compatible with PROS 3.7. While we could have requested changes on your PR,  with the timeline of pros 3.8 I felt it was simpler to have one of our members to do it that way. There definitely should be credit given to @djava . (I haven't reviewed this yet, but that was the intention when assigning this task)

WillXuCodes:
Closed as duplicate to #470

Issue #418, opened 2022-09-21
PROS 3.7.2: cobs_encode() name clashing at runtime


Richard-Stump:
A user on [VTOW](https://discordapp.com/channels/197777408198180864/198658294007463936/1020145283163762738) discovered that implementing his own `cobs_encode` function caused his program to crash at runtime, with the stack trace showing only the address 0x3805820.

I believe the issue is that our implementation of `cobs_encode` in src/common/cobs.c is a public symbol, so when the hot and cold packages link at runtime, the links their `cobs_encode` calls to our implementation. The main thing is that the symbols for our implementation should be stripped.

This should be investigated further to see if `cobs_encode_measure` also causes this issue, and if any other functions in the kernel cause this issue.

PR #505, opened 2023-02-03
🐛 PROS 4: fix cobs_encode() name clashing at runtime

R11G:
#### Summary:
Change symbol for cobs_encode() to not public.

##### References (optional):
Issue #418 

#### Test Plan:
Ensure a function named cobs_encode() in a pros project calls that function when called instead of the pros one
Ensure other functions in pros doesn't have the same issue


noam987:
Closed. Not currently in the works for PROS 4.0

Issue #495, opened 2023-01-24
PROS 3.8: get_voltage() for motor groups


Richard-Stump:
Motor groups do not currently have a method to get their voltage. The user can get the voltage of individual motors in the group using the array operator; however, their is no way to get it as a whole. Since there is a method to set the voltage of all the motors in the group (`move_voltage`), there should be a method to get the voltages.

New methods for motor groups:
```cpp
std::vector<std::uint32_t> get_voltages(void);
std::vector<std::uint32_t> get_voltage_limits(void);
```

Richard-Stump:
I'm combining #474 with this once since they are related: 

Expected Behavior:
A get_raw_positions() method that returns a std::vector<std::int32_t> of the raw encoder positions of a motor group

Actual Behavior:
The method does not exist

Additional Information
Check other motor methods that may need motor group implementations e.g. get_temperatures(), get_powers(), etc.

PR #506, opened 2023-02-03
✨ PROS 3.8: Missing Functions for Motor Groups

Gracelu128:
#### Summary:
Adding following functions to motor groups:
1. get_voltages() - get voltages of all motors
2. get_voltage_limits() - get voltage limits of all motors
3. get_raw_positions() - returns a std::vectorstd::int32_t of the raw encoder positions of a motor group

Also check other motor methods that may need motor group implementations e.g. get_temperatures(), get_powers(), etc.

#### Motivation:
The users can get the voltages and voltage limits of all the motors at once instead getting voltage of individual motors in the group using the array operator. Also consider other motor methods that need to be implemented for motor groups.

##### References (optional):
Issue #495 

#### Test Plan:

- [ ] get_voltages()
- [ ] get_voltage_limits()
- [ ] get_raw_positions()

### DIFF
--- a/include/pros/motors.hpp
+++ b/include/pros/motors.hpp
@@ -1053,7 +1053,7 @@ class Motor_Group {
 	 * \return The encoder position of the motor in ticks or PROS_ERR_F if the operation
 	 * failed, setting errno.
 	 */
-	std::vector<std::int32_t> get_raw_positions(void);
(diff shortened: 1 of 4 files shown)


WillXuCodes (approved):
Hardware tested, LGTM

noam987:
LGTM. Can we merge this in ASAP

PR #509, opened 2023-02-03
🐛Fix templates bundling PROS Kernel headers

BennyBot:
#### Summary:
Changes the Makefile to only allow files inside of `include/LIBNAME` to be included in the packaged zip file.

#### Motivation:
Currently, if this line is left unchanged, templates will include every file inside of the `include` directory in their packaged zip file. This includes pros headers and other templates. This is not good, as currently applying templates will attempt to overwrite files of the same name, which could lead to different kernel versions messing up branchline.

#### Test Plan:
Testing creating a template without these changes, and saw that all of `include/pros` was included in the zip file's `include` directory.
Made these changes to that project's Makefile, and then built the template again. The pros headers were no longer in the zip file.

### DIFF
--- a/template-Makefile
+++ b/template-Makefile
@@ -25,7 +25,9 @@ EXCLUDE_COLD_LIBRARIES:=
 
 # Set this to 1 to add additional rules to compile your project as a PROS library template
 IS_LIBRARY:=0
-# TODO: CHANGE THIS!
+# TODO: CHANGE THIS! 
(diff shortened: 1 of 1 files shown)


WillXuCodes:
This might be breaking, so we really need to warn template makers about this.

WillXuCodes:
Reopen on PROS 4 and we'll do it.

Issue #518, opened 2023-02-07
VDML Device Error Logging 


Richard-Stump:
In the future it would be great to have a log of device disconnects, mismatched ports, etc. that users can access. PROS 3 had features to display these things on the screen, but this feature was removed. The code remains in `#if 0` blocks in PROS 4 and could maybe serve as a basis for this feature.

WillXuCodes:
Shower Thought: This could probably be implemented by dumping some sort of log file to the SD card. 

Seems pretty straight forward to implement with the old VDML logging code just redirected to said file, the main 2 decision decisions to consider would be: 

1. Figuring out what to name the file (could probably just use the compilation timestamp) 
2. FIguring out whether to enable this by default or have a function for the user to enable ( with a provided file name for this )? 

My takes: 
1. See no. 2
2. Have this be something user enabled, with the initialization function specifying the output file. 

( RFC still active here ofc, this is not a "shall" design sort of comment )

PR #519, opened 2023-02-07
💀 PROS 4: Remove dead VDML logging code

Richard-Stump:
#### Summary:
This disables (not removes) the logging code that the VDML had. Since PROS 4 no longer shows messages for device disconnects, mismatches, etc. this code is no longer needed. It was not completely removed since we want to add better logging in the future. See #518 

#### Test Plan:
- [x] Run the kernel on the brain and make sure it doesn't crash
- [x] Unplug a device while the kernel is running to ensure that an event like this doesn't do anything weird.


Richard-Stump:
This was tested and nothing weird or funky occured.

Issue #515, opened 2023-02-07
PROS 4: Rotation Class ctors not explicit


Richard-Stump:
## Task:
The constructors for the rotation class are not declared with the `explicit` keyword. This will prevent copies or conversions of the class from happening implicitly.

It would be good to check the other device classes to make sure that their constructors are marked with the `explicit` keyword as well. 

## Testing:
Make sure implicit copies/conversions do not work:
```cpp
pros::Rotation rot = pros::Rotation(1); // invalid
```

PR #520, opened 2023-02-08
✨PROS 4: Make constructors explicit

ryan-doan:
#### Summary:

Add "explicit" to devices' constructors.

#### Motivation:

Prevent implicit copies and conversion of classes.

##### References (optional):

PROS 4: Rotation Class ctors not explicit https://github.com/purduesigbots/pros/issues/515

#### Test Plan:

- [x] Compile
- [ ] Check for exceptions to explicit and constructors missing explicit
- [ ] pros::Rotation rot = pros::Rotation(1); should be invalid (might be prudent to check every constructors).


Richard-Stump:
Please switch this branch to develop-pros-4. This is a pros 4 task.

PR #522, opened 2023-02-08
✨Pros 4/feature/make classes explicit

ryan-doan:
#### Summary:

Add "explicit" to devices' constructors.

#### Motivation:

Prevent implicit copies and conversion of classes.

##### References (optional):

PROS 4: Rotation Class ctors not explicit https://github.com/purduesigbots/pros/issues/515

#### Test Plan:

- [x] Compile
- [ ] Check for exceptions to explicit and constructors missing explicit
- [ ] pros::Rotation rot = pros::Rotation(1); should be invalid (might be prudent to check every constructors).


include/pros/rtos.hpp
```diff
@@ -669,4 +669,4 @@ using pros::c::delay;
 }  // namespace rtos
 }  // namespace pros
 
-#endif  // _PROS_RTOS_HPP_
+#endif  // _PROS_RTOS_HPP_
```
WillXuCodes, line 672: add back newline please
```suggestion
#endif  // _PROS_RTOS_HPP_

```

version
```diff
@@ -1 +1 @@
-4.0.0
+4.0.0
```
WillXuCodes, line 1:
```suggestion
4.0.0

```

WillXuCodes (changes requested):

Issue #472, opened 2022-11-30
PROS 3/4: Paths with parenthesis break the build system


Richard-Stump:
When dealing with paths that contain parenthesis, the make file fails to build the timestamp file properly. This was discovered by a user on VTOW ([Relevant Message](https://discordapp.com/channels/197777408198180864/198658294007463936/1046395975205388328))

Specifically, the line that pipes the c code to create the timestamp object is the one that breaks. It breaks due to the `char const * const _PROS_COMPILE_DIRECTORY` variable. I suspect this is due to the way quotes work between makefiles, bash, and C code.

PR #530, opened 2023-02-17
🐛 Prevent paths with parenthesis and spaces from breaking

aaryan-gautam:
#### Summary:

Added a wildcard function when the `char const * const _PROS_COMPILE_DIRECTORY` variable is assigned.
##### References (optional):

#472 
#### Test Plan:

Tested with
Create a new pros project using `pros c new <project name>`
Create a project name with spaces and a project with parenthesis
Compile and run without any errors

image

image


WillXuCodes:
(guess I was wrong srry lol)

PR #532, opened 2023-02-21
✨PROS4: Base Class for Motors

### LINKED ISSUE #490: Motor_Groups should fulfill the Container named requirements
Building on #450 and #463, for `pros::Motor_Group` to be fully treatable as a standard container, it should fulfill at the very least the standard's [`Container` named requirements](https://en.cppreference.com/w/cpp/named_req/Container).

Additionally, Motor_Group could likely also fulfill all the named requirements that `std::vector` does (with the exception of [`AllocatorAwareContainer`](https://en.cppreference.com/w/cpp/named_req/AllocatorAwareContainer)), since its underlying data structure is a vector.  Those additional named requirements would be [`SequenceContainer`](https://en.cppreference.com/w/cpp/named_req/SequenceContainer), [`ContiguousContainer`](https://en.cppreference.com/w/cpp/named_req/ContiguousContainer), and [`ReversibleContainer`](https://en.cppreference.com/w/cpp/named_req/ReversibleContainer) (although this one is not particularly relevant to any of `Motor_Group`'s potential use-cases).

The main benefits of this are:
- compatibility with range-based `for` loops
- compatibility with `std` algorithms
- following C++ standard best practice for how a container-like object should behave

I don't believe that this would be a huge effort, since most of the required types and calls would be redirected to the underlying vector. (i.e. `iterator Motor_Group::begin() { return arr.begin(); }`)  This can certainly be targeted for the PROS 4 release, since it's definitely not urgent.

I'm not sure yet how everything will work with the new PROS 4 design for `Motor_Group` but I'd imagine that the solution used for `Motor_Group::operator[]` in PROS 4 will be transferable to these requirements.

### LINKED ISSUE #507: Motor Group Brake Modes should be OOP like reversed
This was a design oversight and should be fixed ASAP. See the design structure for the other properties for motor groups

### LINKED ISSUE #525: PROS 4.0.1 Motors/Motor Group Design
### Background
The current design for Motors, Motor Groups, and their relation is not well liked and goes against C++ norms. We are reverting back to the original idea since the better maintainability does not out weigh the priority of a good API. 

### Proposal:
Motors and motor groups inherit from a class named `AbstractMotor` like originally planned. This class is not constructable (as it is abstract) and contains virtual methods for functions from motors and motor groups
- The motor's move functions. For motor groups, they tell all the motors in the group to move uniformly
- The motor's set functions. For groups, they set the parameter for each motor uniformly
- The motor's get functions: For groups, they get the value from the first motor
- The motor group's get_all_* functions. For motors, they return a vector of 1.

We are also removing the OOP features from the motors/groups since they caused too many issues. The constant calls to send the parameters caused issues. We're moving the OOP reversed flag to the C api level. 

Looking for any more thoughts and considerations on this.

noam987:
#### Summary:
Reworks the inheritance between motors and motor groups by adding a new BaseMotor virtual class. 

#### Motivation:
#525 

##### References (optional):
Closes #525 
Closes #507 by deciding that this should NOT be intended behavior
Closes #490 as it doesn't make sense in the current architecture. Motor groups are vectors of ports, not of motor objects. We have added indexable methods for all non-movement methods to target any one specific motor in a motor group.

#### Test Plan:
- [ ] It compiles
- [ ] Make a motor and motor group and test out some methods


include/pros/motors.hpp
```diff
@@ -95,77 +95,1065 @@ using Motor_Gear = Motor_Gears;
...
+	 * \code
+	 * void autonomous() {
+	 *   pros::Motor motor (1);
+	 *   motor.move_absolute(100, 100);
+	 *   std::cout << "Motor Target: " << motor.get_target_position();
+	 *   // Prints 100
+	 * }
+	 * \endcode
+	 */
+	double get_target_position(const std::uint8_t index) const;
```
Richard-Stump, line 339: I think this is missing the default index of 0

include/pros/motors.hpp
```diff
@@ -1280,106 +2185,12 @@ class Motor_Group {
...
-	using Motor_Group::get_every_power;
-	using Motor_Group::get_every_raw_position;
-	using Motor_Group::get_every_target_position;
-	using Motor_Group::get_every_target_velocity;
-	using Motor_Group::get_every_temperature;
-	using Motor_Group::get_every_torque;
-	using Motor_Group::get_every_voltage;
-	using Motor_Group::get_every_voltage_limit;
-	using Motor_Group::get_ports;
+	std::int16_t _port;
```
Richard-Stump, line 2193: Is there a reason this is an int16_t instead of a int8_t?

Richard-Stump (commented):
This looks good to me so far. Just a couple things that I noticed, but I couldn't find any functions that were missing nor any spots in the C-API where the port's sign wasn't checked.

djava:
I think I prefer the name `AbstractMotor` to `BaseMotor`, feels less likely to be confusing.  I could see people think its like for their robot base (occasionally used synonym for chassis). other opinions?

Richard-Stump:
This was tested and confirmed to work on real hardware. Indexed setters/getters were tested, vectorized setters/getters were tested, and polymorphism was tested. 

Next we need to decide whether to make the methods in `Motor` and `MotorGroup` virtual or not. 
- Argument for making them virtual: Leaves flexibility for future use either by us or users.
- Argument against making them virtual: We want to encourage users to build wrappers that use our classes, not create subclasses that extend them.

djava:
> We want to encourage users to build wrappers that use our classes, not create subclasses that extend them.
Why? I don't think it should really matter to us how users choose to build functionality on top of PROS.

As far as I'm concerned, the argument is really if you're okay with incurring the runtime cost of virtual dispatch on all motor function calls to allow for the possibility of people extending the motor class. Personally, I probably wouldn't take that cost, considering that extending `AbstractMotor` is already an option? Not totally sure though.

Issue #497, opened 2023-01-25
Add a "is_installed()" method for devices


ifndef-define:
While reading the documentation, I found no dedicated method to check whether a V5 device (motor, distance, radio, etc.) is installed. While this isn't something major since you could just call any method and check for the return of an `errno` of `ENODEV`, I thought it might be worth adding a simple check method like `MOTOR.is_connected()` or `MOTOR.is_installed()`.

Richard-Stump:
I think there's been little reason for such a method since most of the time, teams plug their devices in and the code works because they are there. There is little in the way of connecting/disconnecting devices except for building the robot. However, this would be useful for sanity checks and would be clearer than checking errno.

Additionally, would it be useful to introduce a standalone function that can check if there is a device on a port, and if so return the type of device connected?

I'm marking this as RFC (Request For Comment) for the time being to get some more feedback on this idea.

ifndef-define:
Yea like I said, such a method is there for the sake of it being present. Like a program that warns the driver in case a vital motor disconnects in game or something. I was thinking of having a way to manipulate a 6-motor drive so like should a motor on the right side disconnect, the program can detect that and reduce left side power to maintain a straight tracking. 

> Additionally, would it be useful to introduce a standalone function that can check if there is a device on a port, and if so return the type of device connected?

I don't think this kind of feature is needed. At least I can't think of a way this could be utilized. This would probably be useful if the robots could change device ports automatically.

PR #534, opened 2023-02-23
✨Pros 4/feature/is_installed

ryan-doan:
#### Summary:

Added methods to device classes to check if said device is installed correctly in its port.

#### Motivation:

Quickly check installation status of devices.

##### References (optional):

 Add a "is_installed()" method for devices #497.

#### Test Plan:

- [x] Compile.
- [x] Return true when correct device is installed, true for any device if the variable's class is Device (except class Serial).
- [x] Return false when no device or the incorrect device is installed (except class Serial).
IMPORTANT
- [ ] Right now the assumption is to check serial device against E_DEVICE_GENERIC, but since this class is reserved for the user's custom electronics, this class has not been tested.
- [ ] Update newly merged changes for motor groups.

### DIFF
--- a/include/pros/device.hpp
+++ b/include/pros/device.hpp
@@ -18,6 +18,7 @@
 #ifndef _PROS_DEVICE_HPP_
 #define _PROS_DEVICE_HPP_
 
+#include "apix.h"
 #include "pros/misc.hpp"
 #include "pros/rtos.hpp"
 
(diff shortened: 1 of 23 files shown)


Richard-Stump:
Your PR is based off of develop. It needs to be based off of develop-pros-4.

Issue #536, opened 2023-02-23
In some cases, vision_read_by_size will return an error without setting errno


nickmertin:
#### Expected Behavior:
Anytime an error code is returned from any API function, `errno` should be set to indicate the error which occured.

#### Actual Behavior:
Initially, `vision_read_by_size` fills the `signature` of each element in the output array with `VISION_OBJ_ERR_SIG`. If `object_count` is greater than the number of available objects (as given by `vexDeviceVisionObjectCountGet`), then one or more entries at the end of the array will not be touched by the second `for` loop, leaving them with a value which signifies that an error occurred. Meanwhile, unless an actual error occurs in the underlying VEX API, `errno` will not be set, **leaving it at whatever value it happened to be when the function was called.**

#### Steps to reproduce:
I don't have the hardware to test, but you should be able to reproduce by requesting more objects than exist. Try setting `errno` to something distinctive before calling `vision_read_by_size` and note that it won't have changed afterwards.

#### System information:
Platform: V5
PROS Kernel Version: latest develop branch

nickmertin:
Never mind, please ignore. I just noticed that the function returns the number of objects actually read, which solves this problem.

PR #537, opened 2023-02-24
🐛PROS 3.8: Check for insufficient objects in vision_read_by_size()

R11G:
#### Summary:
Added an if statement to vision_read_by_size() to check that there are enough vision objects for the device to fill object_arr and returns an error otherwise

#### Motivation:
User reported issue

##### References (optional):
#536 

#### Test Plan:
Run method with too large of a object_count value and make sure it sets errno to ERANGE.


nickmertin:
See https://github.com/purduesigbots/pros/issues/536#issuecomment-1442645283 - this is actually not necessary.

nickmertin:
This actually is a regression, since it means that we need to know *exactly* how many objects there are in order to read all of them, which is not possible in general due to race conditions.

PR #539, opened 2023-03-07
✨ PROS 4: Add is_installed to virtual device class

ryan-doan:
#### Summary:

Added methods to device classes to check if said device is installed correctly in its port.
#### Motivation:

Quickly check installation status of devices.

##### References (optional):

 Add a "is_installed()" method for devices #497.

#### Test Plan:

- [x] 1. Compile.
- [x] 2. Return true when correct device is installed, true for any device if the variable's class is Device (except class Serial).
- [x] 3. Return false when no device or the incorrect device is installed (except class Serial).
IMPORTANT
- [x] 4. Right now the assumption is to check serial device against E_DEVICE_GENERIC, but since this class is reserved for the user's custom electronics, this class has not been tested.


src/devices/vdml_distance.cpp
```diff
@@ -11,12 +11,25 @@
...
 
+bool Distance::is_installed() {
+	std::uint8_t port = this->_port - 1;
+    port_mutex_take(port);
+    pros::c::v5_device_e_t deviceType = c::registry_get_plugged_type(port);
+    port_mutex_give(port);
+    if (deviceType == c::E_DEVICE_DISTANCE) {
+        return true;
+    }
+    return false;
```
WillXuCodes, line 30:
```suggestion
    port_mutex_take(_port - 1);
    pros::c::v5_device_e_t deviceType = c::registry_get_plugged_type(port);
    port_mutex_give(_port - 1, deviceType == c::E_DEVICE_DISTANCE);
```

WillXuCodes: do this with every function

WillXuCodes (changes requested):

Richard-Stump:
Implementation looks good, just some minor code cleanups and style changes needed.

include/api.h
```diff
@@ -78,4 +78,4 @@
 #include "pros/vision.hpp"
 #endif
 
-#endif  // _PROS_API_H_
+#endif  // _PROS_API_H_
```
WillXuCodes, line 81: add newline back 
```suggestion
#endif  // _PROS_API_H_

```

include/pros/device.hpp
```diff
@@ -23,6 +23,28 @@
...
+	rotation = 4,
+	imu = 6,
+	distance = 7,
+	radio = 8,
+	vision = 11,
+	adi = 12,
+	optical = 16,
+	gps = 20,
+	serial = 129,
+	generic __attribute__((deprecated("use E_DEVICE_SERIAL instead"))) = serial,
```
WillXuCodes, line 44: this isn't necessary, the only reason this is "deprecated" is because it was old API. Since this is new API we do not need to start with a deprecated member. 

```suggestion
	generic __attribute__((deprecated("use E_DEVICE_SERIAL instead"))) = serial,
```

include/pros/device.hpp
```diff
@@ -39,19 +61,29 @@ class Device {
...
+	*/
+	virtual bool is_installed();
 
 	protected:
 	/**
 	 * Creates a Device object.
 	 *
 	 * \param port The V5 port number from 1-21
-	 *
+	 * \param deviceType The type of the constructed device
```
WillXuCodes, line 78: make this two seperate comments (this is for doxygen's sake).

version
```diff
@@ -1 +1 @@
-4.0.0
+4.0.0
```
WillXuCodes, line 1: add newline back 
```suggestion
4.0.0

```

WillXuCodes (changes requested):
@noam987 @Richard-Stump what is the plan regarding motor groups with this PR? Are we registering those with the device class as motors or are we pretending they aren't?

Richard-Stump:
> @noam987 @Richard-Stump what is the plan regarding motor groups with this PR? Are we registering those with the device class as motors or are we pretending they aren't?

We were going with motor groups not being devices since the idea of a single device doesn't map to multiple. 
We intended that the AbstractMotor class wasn't a device either for this reason.

WillXuCodes:
> > @noam987 @Richard-Stump what is the plan regarding motor groups with this PR? Are we registering those with the device class as motors or are we pretending they aren't?
> 
> We were going with motor groups not being devices since the idea of a single device doesn't map to multiple. We intended that the AbstractMotor class wasn't a device either for this reason.

sounds good, totally forgot lol. my bad

noam987 (approved):
Passed Basic hardware testing

PR #545, opened 2023-03-08
:bug: PROS 4: Fix Motor Behavior for move_absolute and move_relative

noam987:
#### Summary:
Fixes behavior of move absolute and move relative for motors
New behavior: 
motor_move_absolute does not care about sign of the port as it is absolute movement. 
motor_move_relative with a negative port value should move in the reversed direction if the distance is positive.
PROS 3 behavior is still functional if you use the c api motor_set_reversed and only positive port numbers

#### Motivation:
Current code does not behave as expected

#### Test Plan:
Test 1:
- [x] motor_move_absolute(1, 3000, 100);
- [x] After it stops moving motor_get_position(1); returns ~3000
- [x]  motor_move_absolute(-1, 3000, 100); No movement should occur
- [x] motor_move_absolute(-1, 5000, 100)
- [x] After it stops moving motor_get_position(1); returns ~5000
 - [x] motor_move_absolute(1, 5000, 100); No movement should occur
- [x] After it stops moving motor_get_position(1); returns ~5000
Test 2:

- [x] motor_move_relative(1, 3000, 100);
- [x] After it stops moving motor_get_position(1); returns ~3000
- [x]  motor_move_relative(-1, 3000, 100);
- [x] After it stops moving motor_get_position(1); returns ~0
- [x] motor_move_relative(-1, 5000, 100)
- [x] After it stops moving motor_get_position(1); returns ~-5000

### DIFF
--- a/include/pros/motors.h
+++ b/include/pros/motors.h
@@ -195,7 +195,7 @@ int32_t motor_move_absolute(int8_t port, const double position, const int32_t ve
  * }
  * \endcode
  */
-int32_t motor_move_relative(int8_t port, const double position, const int32_t velocity);
+int32_t motor_move_relative(int8_t port, double position, const int32_t velocity);
 
 /**
  * Sets the velocity for the motor.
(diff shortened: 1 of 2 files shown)


noam987:
Hardware tested. It works

noam987:
image

WillXuCodes (approved):
LGTM, sauce

PR #547, opened 2023-03-09
⬆️ PROS 4: Fix the implementation for get_type() in the virtual device class

### LINKED ISSUE #527: PROS 4.0.1: Fix the Device Class with some implementations as a base class.
Currently, the device class has some pure virtual functions that should have implementations as a proper base class with a registry call to get what device is plugged in. The point of this is to allow users to dynamically initialize their devices by detecting what is on plugged in on hardware.

aberiggs:
#### Summary:
Implement the device class as a base class with get_type() no longer being a virtual function. Renamed get_type() to get_plugged_type() for clarity, as get_type() now makes a registry call to get the device type plugged into the port associated with a given Device object.

Removed get_type() from children of the device class.

Implemented changes (based on Ryan Doan's PR) to vdml.h and registry.h to define v5_device_e_t

#### Motivation:
Allows for users to dynamically find out the type of device that is plugged in.

##### References:
Closes #527 

#### Test Plan:

- [x] Connect device to port and initialize a device object for it.
- [x] Call get_plugged_type() on that device.
- [x] Check that device enum is what is expected.
- [x] Plug in different device while still running and check that device enum changes to correctly display the new connected device.


src/devices/vdml_device.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+#include "pros/device.hpp"
+#include "pros/apix.h"
+#include "vdml/vdml.h"
+
+
+namespace pros {
+inline namespace v5 {
+
+	std::uint8_t Device::get_port(void) {
```
Richard-Stump, line 26: Could you fix the indentation here? We don't indent within namespaces.

include/vdml/vdml.h
```diff
@@ -17,6 +17,14 @@
 #include <stdint.h>
 #include "vdml/registry.h"
 
+#ifdef __cplusplus
+#define v5_device_e_t pros::c::v5_device_e_t
```
Richard-Stump, line 21: This conditional define can probably get removed since we went with the `DeviceType` enum class instead of the `v5_device_e_t` enum.

Makefile
```diff
@@ -85,4 +85,4 @@ $(LIBAR): $(call GETALLOBJ,$(EXCLUDE_SRC_FROM_LIB)) $(EXTRA_LIB_DEPS)
 	-$Drm -f $@
 	$(call test_output_2,Creating $@ ,$(AR) rcs $@ $(addprefix $(LIBV5RTS_EXTRACTION_DIR)/, $(LIBV5RTS_OBJECTS)) $(call GETALLOBJ,$(EXCLUDE_SRC_FROM_LIB)),$(DONE_STRING))
 # @echo -n "Stripping non-public symbols "
-# $(call test_output,$D$(OBJCOPY) -S -D -g --strip-unneeded --keep-symbols public_symbols.txt $@,$(DONE_STRING))
+# $(call test_output,$D$(OBJCOPY) -S -D -g --strip-unneeded --keep-symbols public_symbols.txt $@,$(DONE_STRING))
```
WillXuCodes, line 88:
```suggestion
# $(call test_output,$D$(OBJCOPY) -S -D -g --strip-unneeded --keep-symbols public_symbols.txt $@,$(DONE_STRING))

```
Add newline back

src/devices/vdml_device.cpp
```diff
@@ -0,0 +1,44 @@
...
+	DeviceType type = (DeviceType) pros::c::registry_get_plugged_type(_port - 1);
+	
+	return_port(_port - 1, type);
+}
+
+Device::Device(const std::uint8_t port) : _port(port) {}
+
+
+}  // namespace v5
+}  // namespace p
```
WillXuCodes, line 44:
```suggestion
}  // namespace v5
}  // namespace pros

```
Revert

src/devices/vdml_device.cpp
```diff
@@ -0,0 +1,44 @@
+/**
+ * \file devices/vdml_device.cpp
+ *
+ * Base class for all smart devices.
+ *
```
WillXuCodes, line 5: lol be sure to add errno vals here

WillXuCodes (changes requested):
One other request: can we remove the include for rtos.hpp and misc.hpp and move the DeviceType enum class into device.hpp? Thanks.

src/devices/vdml_device.cpp
WillXuCodes: (Oh crap nevermind I am stupid... this was not necessary. I'm insanely sorry don't let me do late night code review ever again )

WillXuCodes (approved):
LGTM

PR #550, opened 2023-03-10
✨ PROS 4: Move device enum out of apix and motor funcs to apix

aberiggs:
#### Summary:
Created C version of device.hpp and device.cpp
- Created get_installed_type() for device, taking a port and returning v5_device_type_e_t (the type that is plugged into that port). 

Moved the enum v5_device_type_e_t from apix.h to device.h

Moved motor_set_reversed() and motor_is_reversed() from motors.h to apix.h

#### Motivation:
We're wanting to use features utilizing v5_device_type_e_t more and are updating how the device class works so it makes sense to move the enum to a C version of the device base class.

#### Test Plan:
To be determined


src/devices/vdml_device.c
```diff
@@ -0,0 +1,29 @@
...
+
+v5_device_e_t get_plugged_type(uint8_t port) {
+    if (!port_mutex_take(port - 1)) {                            
+		errno = EACCES; 
+		return E_DEVICE_UNDEFINED;                                                                           
+	}
+	v5_device_e_t type = registry_get_plugged_type(port - 1);
+	
+	return_port(port - 1, type);
+}
```
WillXuCodes, line 29:
```suggestion
}

```
Add newline back please

include/pros/device.h
```diff
@@ -0,0 +1,62 @@
...
+	E_DEVICE_GPS = 20,
+	E_DEVICE_SERIAL = 129,
+	E_DEVICE_GENERIC __attribute__((deprecated("use E_DEVICE_SERIAL instead"))) = E_DEVICE_SERIAL,
+	E_DEVICE_UNDEFINED = 255
+} v5_device_e_t;
+
+/**
+	 * Gets the type of device on given port.
+	 *
+	 * \return The device type as an enum.
```
WillXuCodes, line 53:
```suggestion
* Gets the type of device on given port.
*
* \return The device type as an enum.
```

fix indentation

src/devices/vdml_motors.c
```diff
@@ -17,6 +17,7 @@
 #include "kapi.h"
 #include "pros/colors.h"
 #include "pros/motors.h"
+#include "pros/apix.h"
```
WillXuCodes, line 20: why is this here? is this necessary?

WillXuCodes (changes requested):

src/devices/vdml_motors.c
aberiggs: Since I was told to move the prototypes of 2 of the functions within motors.h into apix.h, so I thought it made sense to import the header into vdml_motors.c since those 2 functions are now prototyped within apix.h. It appears to compile without, so if you'd rather me remove I can do that.

WillXuCodes: Got it! Srry forgot that was part of it lol

WillXuCodes (changes requested):
Forgot to mention but device.h should be put under api.h

Makefile
noam987, line 1: Can you make sure we get a newline at the end of this file

noam987 (approved):
LGTM except for the newline at the end of the make file missing

Makefile
aberiggs: Oh shoot my bad, I'll do that right now.

WillXuCodes (approved):
LGTM

PR #551, opened 2023-03-11
🐛PROS 4 negative port support in C API

noam987:
#### Summary:
Adjusted implementation of getters and a variety of methods to be more inline with expected behavior

#### Test Plan:
```
 motor_move_relative(-1, 100, 100); // Moves 100 units forward
    while (!((motor_get_position(-1) < 105) && (motor_get_position(-1) > 95))) {
      // Continue running this loop as long as the motor is not within +-5 units of its goal
      delay(2);
    }
 
    motor_move_relative(-1, 100, 100); // Also moves 100 units forward
    while (!((motor_get_position(-1) < 205) && (motor_get_position(-1) > 195))) {
      delay(2);
   }
 }
```

### DIFF
--- a/src/devices/vdml_motors.c
+++ b/src/devices/vdml_motors.c
@@ -90,26 +90,26 @@ int32_t motor_modify_profiled_velocity(int8_t port, int32_t velocity) {
 }
 
 double motor_get_target_position(int8_t port) {
-	port = abs(port);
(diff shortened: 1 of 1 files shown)


noam987:
Hardware tested and works

WillXuCodes (approved):
LGTM

Issue #548, opened 2023-03-10
PROS 4.0.3: Add Overloads to functions that take color enums to also take integer values. 


WillXuCodes:
This is because right now we've accidentally disabled anything outside of our dictated color range.

PR #553, opened 2023-03-13
✨ PROS 4: Add numerical color override for pen/eraser

WillXuCodes:
Summary:
added pros::screen::set_pen and set_eraser with a uint32_t as an input since values outside of pros::Color do not work without it.

Motivation:
Template/porting work is a hassle without custom colors, and it would be nice for our users.

References (optional):
closes https://github.com/purduesigbots/pros/issues/548

Test Plan:
 - [x] [test item](See if this compiles: pros::screen::set_eraser(0x00FF1234);)
 - [x] [test item](See if this compiles: pros::screen::set_pen(0x00FF1234);)

### DIFF
--- a/include/pros/screen.h
+++ b/include/pros/screen.h
@@ -151,7 +151,7 @@ namespace c {
  * }
  * \endcode
  */
-uint32_t screen_set_pen(color_e_t color);
+uint32_t screen_set_pen(uint32_t color);
 
 /**
  * Set the eraser color for erasing and the current background.
@@ -180,7 +180,7 @@ uint32_t screen_set_pen(color_e_t color);
  * }
  * \endcode
  */
-uint32_t screen_set_eraser(color_e_t color);
+uint32_t screen_set_eraser(uint32_t color);
 
 /**
  *  Get the current pen color.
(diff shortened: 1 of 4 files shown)


WillXuCodes:
Was compile tested already.

WillXuCodes:
Ok now this has been compile tested properly with an edge case lol

noam987 (approved):
Can you hardware test or send the code an expected behavior and I can hardware test?
Also what remaining is using pros::color at this point? We should make sure it still has uses based on how many features we removed the color from

WillXuCodes:
> Can you hardware test or send the code an expected behavior and I can hardware test? Also what remaining is using pros::color at this point? We should make sure it still has uses based on how many features we removed the color from

colors.hpp provides a type safe way for library writers to only allow the set colors into their library's API if they so desire. Furthermore, we still need it for the CPP api since you can't exactly pass it into a uint32_t call and have it implicitly convert. 

Here is some code to test:
```cpp
// be sure to disable LLEMU

int opcontrol() {
pros::screen::set_eraser(0x00FF1234);
pros::screen::erase_area(0,0,100,100);
pros::screen::set_pen(0x10FF1234);
pros::screen::erase_area(100,0,200,100);
while(true) {
pros::delay(20);
}
```

}
```

noam987:
Hardware tested and fixed bug in CPP set_eraser calling screen_set_pen for uint

Issue #549, opened 2023-03-10
PROS 4.0.3: Add pros::lcd::is_initialized to weak symbols and make it always return false if weak. 


WillXuCodes:
This is for EZTemplate porting as well, allows for easy detection if LLEMU is installed with LVGL.

PR #555, opened 2023-03-13
:sparkles: PROS 4: Add LCD Initialize Weak Symbol

WillXuCodes:
Summary:
Adds pros::lcd::is_initialized to the weak symbol table for LLEMU stubs. This is a way to help templates who check for this (such as EZTemplate) to see if both LVGL is installed and LLEMU is initialized. To accomplish this, this function will return false if it is still defined as the weak symbol.

Motivation:
Porting work on EZTemplate.

References (optional):
closes https://github.com/purduesigbots/pros/issues/549

Test Plan:
- [ ] Not really much other than seeing if this PR compiles


noam987 (approved):
LGTM

PR #557, opened 2023-03-14
🐛PROS4 Motor::set_reversed_all(false) fix

noam987:
#### Summary:
Motor::set_reversed_all(false) would do nothing, when it was supposed to un reverse a motor. 

#### Test Plan:
Was hardware tested. 
- [x] ```pros::Motor left_mtr(1);

	left_mtr = 100;
	pros::delay(1000);
	left_mtr.set_reversed_all(true);
	left_mtr = 100;
	pros::delay(1000);
	left_mtr.set_reversed_all(false);
	left_mtr = 100;```


WillXuCodes (approved):
Good catch!

PR #560, opened 2023-03-24
🐛 PROS 4: Fix llemu.h and llemu.hpp includes

Richard-Stump:
### Overview: 
This PR is part of the fix for lcd::print() issue on the liblvgl repo: https://github.com/purduesigbots/liblvgl/issues/25

This moves the weak symbols for liblvgl into new `llemu.hpp` and `llemu.h` files. Then, the actual llemu headers in the liblvgl repo are conditionally included if they are found by common.mk. This ensures that when liblvgl is present, the llemu headers are included in `api.h` as they were in PROS 3. When the llemu headers are not included in api.h, the user program only sees the weak symbols (despite the fact strong symbols are present), and will not link to the strong ones.

There are two sperate PRS for the liblvgl repo that implement the necessary changes for liblvgl: one for LVGL 5.3.x and another for LVGL 8.3.x:
* https://github.com/purduesigbots/liblvgl/pull/30
* https://github.com/purduesigbots/liblvgl/pull/31

### Tests:
- [x] Ensure the kernel compiles
- [x] Ensure that the weak symbols work properly


common.mk
```diff
@@ -5,13 +5,23 @@ MFLAGS=-mcpu=cortex-a9 -mfpu=neon-fp16 -mfloat-abi=softfp -Os -g
...
+endif
+
 WARNFLAGS+=-Wno-psabi
 
 SPACE := $() $()
 COMMA := ,
 
 C_STANDARD?=gnu11
-CXX_STANDARD?=gnu++20
+CXX_STANDARD?=gnu++2a
```
WillXuCodes, line 24: revert

WillXuCodes (changes requested):
Does this work with hardware and with the liblvgl template yet?

common.mk
WillXuCodes, line 24:
```suggestion
CXX_STANDARD?=gnu++20
```

WillXuCodes (changes requested):

PR #566, opened 2023-04-09
🐛Removed operator<< for motor groups

noam987:
#### Summary:
removes the << operator from motor groups
#### Motivation:
Calling the << operator on a motor group would cause linking issues
After beta release we can implement it properly

#### Test Plan:
- [] Compiles


WillXuCodes:
LGTM

Issue #500, opened 2023-02-02
BUG: Multiple threads writing to stdout can cause encoding errors


Richard-Stump:
Multiple threads writing to stdout can cause utf-8 encoding errors. This is likely due to stdout not being atomic, or maybe the design of our VFS. A closer look needs to be taken to determine the issue here.

Program to replicate bug:
```cpp
#include <iostream>
#include "main.h"

using namespace pros;

void my_task(void* str) {
    pros::Vision vision(20);
    while (true) {
        std::cout << (char*)str << std::endl;
        std::cout << vision.get_object_count() << std::endl;
        pros::delay(20);
    }
}

void opcontrol() {
    {
        std::string s = "Hello world";
        pros::rtos::Task t(my_task, (void*)(s.c_str()));
    }
    while (true) {
        pros::Motor(11).move(pros::Controller(E_CONTROLLER_MASTER).get_analog(E_CONTROLLER_ANALOG_LEFT_Y));
        pros::Motor(1).move(-pros::Controller(E_CONTROLLER_MASTER).get_analog(E_CONTROLLER_ANALOG_RIGHT_Y));
        std::cout << ("Hello from opcontrol\n") << std::endl;
        pros::Task::delay(20);
    }
}
```

![Image](https://user-images.githubusercontent.com/54607143/216241141-185cec49-b8a4-4f26-8588-7ffb73db350f.png)

WillXuCodes:
Should have been fixed in this PR above ^

PR #570, opened 2023-04-18
🐛Add mutex control for printf and cout

ryan-doan:
#### Summary:

Added mutex control to protect printf and cout from causing utf-8 encoding errors due to race conditions. This is done by preventing context switches while prvWriteBytesToBuffer() and prvReadBytesToBuffer() is executing.

##### References (optional):

 BUG: Multiple threads writing to stdout can cause encoding errors #500 

#### Test Plan:

It is difficult to verify that the changes entirely prevents utf-8 encoding errors since the race condition is difficult to reproduce consistently.

- [x] Compile.
- [ ] Implementation prevents utf-8 encoding from happening.

### DIFF
--- a/src/rtos/stream_buffer.c
+++ b/src/rtos/stream_buffer.c
@@ -537,6 +537,7 @@ TimeOut_t xTimeOut;
 					break;
 				}
 			}
+			taskEXIT_CRITICAL();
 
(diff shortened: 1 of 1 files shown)


noam987 (approved):
LGTM

Issue #569, opened 2023-04-18
PROS 4.0.4: Add Converting Constructors for Each Device Class


WillXuCodes:
#### Expected Behavior:
Issue right now is that you can't cast a motor to a device class, so dynamic instantiation is not possible. 

#### Actual Behavior:
We need a way for users to create motors based on detection, so we should add a constructor to create a motor, imu, and other "vdml" device that takes in a device reference for each device.

PR #571, opened 2023-04-21
Pros4/feature/converting constructor

ryan-doan:
#### Summary:

Add dynamic instantiation by a device constructor that takes another device object

##### References (optional):

 PROS 4.0.4: Add Converting Constructors for Each Device Class #569 

#### Test Plan:

- [x] Compile
- [x] Create a device object with a different subclass which is used to instantiate a device object of another subclass. Is_installed is then called to verify conversion is successful.


noam987 (changes requested):
1. In my opinion these constructors should take in device references, not devices.
 2. This did not pass my simple hardware testing (make a motor, cast it to a device, recast it to a motor then try to spin it), as it did not create a motor object. this is an issue with all the devices

aberiggs:
LGTM

aberiggs (approved):
LGTM

noam987:
I'm going to wait with this until after initial open beta release. We need to think about whether or not some of the devices need to be be explicit constructors or not. This is not critical or a major bug for initial release, more of a QOL feature

noam987:
Stale PR

PR #572, opened 2023-04-27
✨More Motor Constructors

noam987:
#### Summary:
New constructor for motors and motor groups that takes in a AbstractMotor reference

#### Motivation:
This enables better casing back and forth between motors, abstract motors, and motor groups

#### Test Plan:
- [ ] Cast motor to abstract motor to motor
- [ ] cast motor to abstract motor to motor group
- [ ] cast motor group to abstract motor to motor
- [ ] cast motor group to abstract motor to motor group
- [ ] cast motor to motor group
- [ ] cast motor group to motor

### DIFF
--- a/include/pros/motor_group.hpp
+++ b/include/pros/motor_group.hpp
@@ -113,6 +113,35 @@ class MotorGroup : public virtual AbstractMotor {
 	explicit MotorGroup(const std::vector<std::int8_t>& ports, const pros::v5::MotorGears gearset = pros::v5::MotorGears::green,
 	                    const pros::v5::MotorUnits encoder_units = pros::v5::MotorUnits::degrees);
 
+	 /**
+	 * Constructs a new MotorGroup object from an abstract motor.
+	 * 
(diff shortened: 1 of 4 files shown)


djava:
>  cast motor group to motor
wait what? what is the intended behavior here? it turns into just being the first motor in the group or something? it feels like you'd probably rather just prevent that conversion, right?

noam987:
> > cast motor group to motor
> > wait what? what is the intended behavior here? it turns into just being the first motor in the group or something? it feels like you'd probably rather just prevent that conversion, right?

So the idea is that we want a defined behavior for how to create a motor from an abstract motor. We accomplish this by calling the other constructor of motor with the port from abstract_motor.get_port(). If the abstract motor happens to be a motor group, it will return the port of the first motor and make a new object accordingly. Since the constructor takes in an abstract motor, and motor group extends from abstract motor, we want to test that it will behave as expected.

aberiggs (approved):
LGTM

Issue #588, opened 2023-08-08
Can't create a reversed Rotation sensor using the constructor


maxgreen01:
#### Expected Behavior:
Just like motors, users should be able to construct a Rotation sensor object using a signed port number, and the sensor would be reversed if the port is negative. Alternatively, users should be able to pass in a signed port number along with a bool to determine if the sensor is reversed, where the bool overrides the sign of the port number.

Existing pros::Motor equivalent constructors:
```c++
Motor::Motor(const std::int8_t port, const bool reverse) : _port(abs(port)) {
	set_reversed(reverse);
	if (port < 0) 
		set_reversed(true);
}

Motor::Motor(const std::int8_t port) : _port(abs(port)) {
	if (port < 0) 
		set_reversed(true);
}
```
&nbsp;

#### Actual Behavior:
The Rotation constructor that takes an int uses a `uint8_t`, and passing a signed int into this function means the sensor is constructed using a completely different port than intended because of the conversion from `int` to `uint`. When I try to get data from a sensor initialized like this, error values are returned.

Also, in one of my projects, using the constructor that takes a `uint8_t` and a `bool` for reversed immediately causes a Data Abort Exception with the error code 3437384. This happens when the bool is either true or false, but only when the sensor is defined outside of a function scope. I'm not sure if this is caused by something in PROS or by how I have it set up in my project, and I've been able to recreate it in a blank project but with slightly different results.

&nbsp;

#### Steps to reproduce:
Try to construct a Rotation sensor using a negative port number.

Example setup in a default project:
```c++
void opcontrol() {
	pros::Rotation r1(14); //works as expected
	r1.reverse(); //also works as expected
	pros::Rotation r2(-14); //runs but isn't usable - I think tries to use port 242 because of int to uint conversion
	pros::Rotation r3(14, true); //data abort exception in some scenarios
	pros::Rotation r4(14, false); //same as above
	pros::Rotation r5(-14, true); //same as r2, and data abort exception in some scenarios
	pros::Rotation r6(-14, false); //same as above

	while (true) {
		printf("r1:   reversed = %i   pos = %i\n", r1.get_reversed(), r1.get_position()); //prints normally as expected
		printf("r2:   reversed = %i   pos = %i\n", r2.get_reversed(), r2.get_position()); //prints error codes for both
		printf("r3:   reversed = %i   pos = %i\n", r3.get_reversed(), r3.get_position()); //works normally most of the time
[...]

PR #595, opened 2023-08-31
✨Negative port values for Rotation class

R11G:
#### Summary:
Changed unsigned port parameter in Rotation constructor to signed
Allow for negative port values to reverse their positive port

#### Motivation:
Generalize negative port = reverse functionality for more classes

##### References:
#588

#### Test Plan:
Ensure rotation sensor is reversed when a negative port is input, regardless of reverse_flag value.

### DIFF
--- a/src/devices/vdml_rotation.cpp
+++ b/src/devices/vdml_rotation.cpp
@@ -16,8 +16,12 @@
 namespace pros {
 inline namespace v5 {
     
-Rotation::Rotation(const std::uint8_t port, const bool reverse_flag) : Device(port, DeviceType::rotation) {
-	pros::c::rotation_set_reversed(port, reverse_flag);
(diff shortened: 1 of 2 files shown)


src/devices/vdml_rotation.cpp
```diff
@@ -16,8 +16,12 @@
 namespace pros {
 inline namespace v5 {
     
-Rotation::Rotation(const std::uint8_t port, const bool reverse_flag) : Device(port, DeviceType::rotation) {
-	pros::c::rotation_set_reversed(port, reverse_flag);
+Rotation::Rotation(const std::int8_t port, const bool reverse_flag) : Device(port, DeviceType::rotation) {
```
noam987, line 19: This would be instead of the reverse flag.
So you should remove the reverse flag and just use the negative port

noam987 (changes requested):

noam987 (approved):
LGTM. Will post on beta server for hardware testing

PR #599, opened 2023-09-26
Update default index of get_target_position in motor groups

aberiggs:
#### Summary:
Motor Groups method get_target_velocity() does not have a default index of 0. You should be able to provide no arguments to get_target_velocity() and have it return the target velocity of the first motor in the list (motor @ index 0). 

#### Motivation:
Every other similar motor group method has the default index set to 0.

#### Test Plan:
Manually Tested


noam987 (approved):
lgtm

PR #602, opened 2023-09-28
🐛Motor and motor group bug fixes

noam987:
#### Summary:
Fixed bug with motor get_direction and negative ports
migrate emplace_back to push back
fixed some bugs with some configuration setters

#### Motivation:
Reasoning for changing emplace back to push back is that this seems like the correct way to do it when we already have objects of the type that the vectors want. 

#### Test Plan:
manually tested return values for get direction, get_direction_all for both motors and motor groups with positive and negative directions


aberiggs:
LGTM

aberiggs (approved):
LGTM

PR #605, opened 2023-10-02
Fix typos in motors/motor groups docs

aberiggs:
#### Summary:
Updated text in the motors docs to remove typos

#### Motivation:
Correct typos


noam987 (approved):
LGTM

PR #606, opened 2023-10-03
📝Moved function definitions for _all functions in motors to their own …

aberiggs:
…category

#### Summary:
Moved function definitions for _all functions to the bottom of the file into their own category in doxygen

Also made a couple of typo corrections

#### Motivation:
Clean up documentation and move  *_all function definitions to their own category since they're less commonly used on standalone motors.

#### Test Plan:
N/A


noam987 (approved):
LGTM

PR #608, opened 2023-10-12
✨Field Control State Getter

### LINKED ISSUE #471: ✨Add Field Control State/Type Getters
VexCode has the option to detect the difference between a competition switch and field control. This would be helpful for writing debugging programs that still run if the competition switch is being used to test the robot program.

CChheerryyll:
#### Summary:

Added two functions to check if the brain is connected to field control or competition switch.

##### References (optional):

closes #471 

#### Test Plan:

connect the brain to competition switch/field control and call the two functions to see if they return the correct value


djava:
Test-plan wise, there are a few types of field control. Should probably be tested with both the normal older field controller and the new v5 brain-based smart field control

noam987:
> Test-plan wise, there are a few types of field control. Should probably be tested with both the normal older field controller and the new v5 brain-based smart field control

It has been tested on both the smart field control and the old field control switches

noam987 (commented):
The changes you made in controler and misc look great, and I'm exited about this functionality. I'm going to make a quick clean up commit to fix some side effects of the merge conflicts and this will be good to be merged

include/pros/misc.h
```diff
@@ -41,6 +41,7 @@
 #define COMPETITION_DISABLED (1 << 0)
 #define COMPETITION_AUTONOMOUS (1 << 1)
 #define COMPETITION_CONNECTED (1 << 2)
+#define COMPETITION_SYSTEM (1 << 3)
```
WillXuCodes, line 44: since we're making breaking changes anyways, do we want to switch this into an enum instead? If it's a readability thing with bits we could also just add a comment or support both enums and the old macros.

include/pros/misc.h
```diff
@@ -130,6 +131,17 @@ uint8_t competition_get_status(void);
  */
 #define competition_is_autonomous() ((competition_get_status() & COMPETITION_AUTONOMOUS) != 0)
 
+/**
+ * \return True if the V5 Brain is connected to VEXnet Field Controller, false otherwise.
+*/
+#define competition_is_field_control() ((competition_get_status() & COMPETITION_SYSTEM) != 0)
```
WillXuCodes, line 137: ditto on the macro thing... maybe these could be functions?

include/pros/misc.h
```diff
@@ -130,6 +131,17 @@ uint8_t competition_get_status(void);
...
+
+/**
+ * \return True if the V5 Brain is connected to VEXnet Competition Switch, false otherwise.
+*/
+#define competition_is_competition_switch() ((competition_get_status() & COMPETITION_SYSTEM) == 0)
```
WillXuCodes, line 142: we should come up with a better naming scheme for these, I know it's a legacy thing but man competition_is_competition is kind of repetitive.

src/main.cpp
```diff
@@ -58,7 +58,25 @@ void competition_initialize() {}
...
+			if (pros::competition::is_competition_switch()) {
+				printf("[Auton] testing is_competition_switch(): true\n");
+			}
+			else {
+				printf("[Auton] testing is_competition_switch(): false\n");
+			}
+
+			pros::delay(1000);
+	}
+}
```
WillXuCodes, line 79: be sure to remove this before merge.

version
```diff
@@ -1 +1 @@
-4.0.2
\ No newline at end of file
+4.0.5
```
WillXuCodes, line 1: forgot the newline... somebody did somewhere earlier as well. Please add it back

WillXuCodes (changes requested):

noam987 (approved):
LGTM

PR #612, opened 2023-10-23
✨Add list_files function

noam987:
#### Summary:
Adds functionality to list the files on a sd card

#### Motivation:
Frequently requested feature. I will add a new issue to provide a wrapper function for the CPP api to overload it to return a list of strings without the user only needing to give the path

#### Test Plan:

- [x] List the files in the default directory
- [x] List the files in a sub directory

### DIFF
--- a/include/pros/misc.h
+++ b/include/pros/misc.h
@@ -735,13 +735,13 @@ int32_t usd_is_installed(void);
  * \code
  * void opcontrol() {
  * 	char* test = (char*) malloc(128);
- *	pros::c::list_files("/", test, 128);
+ *	pros::c::usd_list_files("/", test, 128);
  *	pros::delay(200);
  *	printf("%s\n", test);
  * }
  * \endcode
 */
-int32_t list_files(const char* path, char* buffer, int32_t len);
+int32_t usd_list_files(const char* path, char* buffer, int32_t len);
 
 /******************************************************************************/
 /**                              Date and Time                               **/
--- a/include/pros/misc.h
+++ b/include/pros/misc.h
@@ -727,7 +727,25 @@ int32_t usd_is_installed(void);
  * Lists the files in a directory specified by the path
  * Puts the list of file names (NOT DIRECTORIES) into the buffer seperated by newlines
  * 
- * \note use a path of "\" to list the files in the main directory
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * 
+ * EIO - Hard error occured in the low level disk I/O layer
+ * EINVAL - file or directory is invalid, or length is invalid
+ * EBUSY - THe physical drinve cannot work
+ * ENOENT - cannot find the path or file
+ * EINVAL - the path name format is invalid
+ * EACCES - Access denied or directory full
+ * EEXIST - Access denied
+ * EROFS - SD card is write protected 
+ * ENXIO - drive number is invalid or not a FAT32 drive
+ * ENOBUFS - drive has no work area
+ * ENFILE - too many open files
+ * 
+ * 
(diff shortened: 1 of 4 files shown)


include/pros/misc.h
```diff
@@ -723,6 +723,26 @@ double battery_get_capacity(void);
  */
 int32_t usd_is_installed(void);
 
+/**
+ * Lists the files in a directory specified by the path
+ * Puts the list of file names (NOT DIRECTORIES) into the buffer seperated by newlines
+ * 
+ * \note use a path of "\" to list the files in the main directory
```
djava, line 730: So how does this interact with the typical `\usd\` path prefix to access the root of the SD card?

include/pros/misc.h
```diff
@@ -723,6 +723,26 @@ double battery_get_capacity(void);
...
+ * 
+ * \return 1 on success or PROS_ERR on failure setting errno
```
djava, line 732: What is a failure?? What do the different errno codes mean?

I assume it can fail because either path doesn't exist, or the buffer was too small for the list, or maybe one of the pointers was null (maybe would just segfault there though...)

Any way to explain how to figure out why it failed? This would actually be very helpful to distinguish failure reasons, which matter here more than in most cases for the API.

include/pros/misc.h
```diff
@@ -723,6 +723,26 @@ double battery_get_capacity(void);
...
+ * 
+ * \b Example
+ * \code
+ * void opcontrol() {
+ * 	char* test = (char*) malloc(128);
+ *	pros::c::list_files("/", test, 128);
+ *	pros::delay(200);
+ *	printf("%s\n", test);
```
djava, line 740: Maybe some indication of what this actually prints?? I have no idea right now what the output format looks like.

src/devices/vdml_usd.cpp
```diff
@@ -21,5 +21,9 @@ std::int32_t is_installed(void) {
 	return usd_is_installed();
 }
 
+int32_t list_files(const char* path, char* buffer, int32_t len) {
+	return list_files(path, buffer, len);
```
djava, line 25: umm this is just infinitely recursing right? I think you need `c::list_files(...)`?

djava (changes requested):

include/pros/misc.h
noam987: I will add more details to the errno section. I ran out of time before meetings and wanted to get this up tonight

include/pros/misc.h
noam987: I can try to add more clarification on what it prints, but all the file names are in the buffer string seperated by newlines. 
For example if you had files A.txt, B.txt and C.txt and a folder D, this would have "A.txt\nB.txt\nC.txt\n"

src/devices/vdml_usd.cpp
noam987: This is not infinitely recursing. there is `using namespace pros::c;` on line 18 of this file. I have verified it works on hardware

include/pros/misc.h
noam987: I have not checked to see how it interacted with \usd\, I can check tomorrow. But a path of "\" listed the files in the root directory of the sd card

aberiggs (approved):
LGTM

djava:
Am I too late to say "hmm maybe this should be called `list_files_raw` or something like that"?

Getting the string isn't particularly useful for.. anything without doing additional processing that honestly pros should really handle for you. Reserving the name `list_files` for this pretty useless base-level functionality seems unfortunate.

PR #613, opened 2023-10-26
:sparkles: ADI Get Port

phinc23:
#### Summary:
Added a get_port() method for all adi sesnors with functionality for adi extenders

#### Motivation:
Requested feature.

#### Test Plan:

- [x] AnalogIn
- [x] AnalogOut
- [x] DigitalIn
- [x] DigitalOut
- [x] Encoder (with extender)
- [x] Ultrasonic
- [x] Gyro
- [x] Potentiometer
- [x] Led

### DIFF
--- a/include/pros/adi.hpp
+++ b/include/pros/adi.hpp
@@ -190,6 +190,26 @@ class Port {
 	 */
 	std::int32_t set_value(std::int32_t value) const;
 
+	/**
+	 * Gets the port of the sensor.
+	 *
+	 * \return 1 if the operation was successful or PROS_ERR if the operation
+	 * failed, setting errno.
+	 * 
+	 * \b Example
+	 * \code
+	 * #define DIGITAL_SENSOR_PORT 1 // 'A'
+	 * 
+	 * void initialize() {
+	 *   pros::adi::AnalogIn sensor (DIGITAL_SENSOR_PORT);
+	 *   
+	 * 	 // Prints the second value from the port tuple (The Adi Port. The first value is the Smart Port)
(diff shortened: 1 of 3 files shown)


include/pros/adi.hpp
```diff
@@ -190,6 +190,26 @@ class Port {
 	 */
 	std::int32_t set_value(std::int32_t value) const;
 
+	/**
+	 * Gets the port of the sensor.
+	 *
+	 * \return 1 if the operation was successful or PROS_ERR if the operation
+	 * failed, setting errno.
```
noam987, line 197: Can we clarify this return a bit more. It doesn't return 1 or pros error it returns a tuple with 3 values instead.
Can we also specify what is in each position of the tuple and state here what the smartport value for the non-expander is

src/devices/vdml_adi.cpp
```diff
@@ -46,6 +46,11 @@ std::int32_t Port::get_value() const {
 	return ext_adi_port_get_value(_smart_port, _adi_port);
 }
 
+ext_adi_port_tuple_t Port::get_port() const {
+	ext_adi_port_pair_t _port_pair = std::make_pair(_adi_port, PROS_ERR_BYTE);
+	return std::make_tuple(+_smart_port, std::get<0>(_port_pair), std::get<1>(_port_pair));
```
noam987, line 51: Why not directly do std::make_tuple(+_smartport, _adi_port, PROS_ERR_byte)

noam987 (changes requested):
Few minor changes. Otherwise looks good

phinc23:
Rewrote: documentation and reduced method redundancy

noam987 (approved):
LGTM

PR #615, opened 2023-10-27
✨Add list_files function #612

Gracelu128:
#### Summary:
Adds functionality to list the files on a sd card.

#### Motivation:
Made wrapper function for the CPP api that returns a vector of strings of file names. Renamed original function as list_files_raw in both CPP api and C api.

#### Test Plan:

- [ ] Get vector of file names in the default directory
- [ ] Get vector of file names in a sub directory

### DIFF
--- a/include/pros/misc.hpp
+++ b/include/pros/misc.hpp
@@ -26,6 +26,7 @@
 
 #include <cstdint>
 #include <string>
+#include <vector>
 
 namespace pros {
 inline namespace v5 {
--- a/include/pros/misc.hpp
+++ b/include/pros/misc.hpp
@@ -526,12 +526,30 @@ namespace usd {
  */
 std::int32_t is_installed(void);
 /**
- * Lists the files in a directory specified by the path
+Lists the files in a directory specified by the path
  * Puts the list of file names (NOT DIRECTORIES) into the buffer seperated by newlines
  * 
- * \note use a path of "\" to list the files in the main directory
+ * This function uses the following values of errno when an error state is
+ * reached:
  * 
- * \return 1 on success or PROS_ERR on failure, setting errno
+ * EIO - Hard error occured in the low level disk I/O layer
+ * EINVAL - file or directory is invalid, or length is invalid
+ * EBUSY - THe physical drinve cannot work
+ * ENOENT - cannot find the path or file
+ * EINVAL - the path name format is invalid
+ * EACCES - Access denied or directory full
+ * EEXIST - Access denied
+ * EROFS - SD card is write protected 
+ * ENXIO - drive number is invalid or not a FAT32 drive
+ * ENOBUFS - drive has no work area
+ * ENFILE - too many open files
+ * 
+ * 
+ *  
+ * \note use a path of "\" to list the files in the main directory NOT "/usd/"
+ *  DO NOT PREPEND YOUR PATHS WITH "/usd/"
+ * 
+ * \return 1 on success or PROS_ERR on failure setting errno
  * 
  * \b Example
  * \code
@@ -539,7 +557,12 @@ std::int32_t is_installed(void);
  * 	char* test = (char*) malloc(128);
  *	pros::usd::list_files("/", test, 128);
  *	pros::delay(200);
- *	printf("%s\n", test);
+ *	printf("%s\n", test); //Prints the file names in the root directory seperated by newlines
+ *  pros::delay(100);
+ *  pros::list_files("/test", test, 128);
+ *	pros::delay(200);
+ *	printf("%s\n", test); //Prints the names of files in the folder named test seperated by newlines
+ *  pros::delay(100);
  * }
  * \endcode
 */
--- a/include/pros/misc.hpp
+++ b/include/pros/misc.hpp
@@ -525,6 +525,25 @@ namespace usd {
  * \endcode
  */
 std::int32_t is_installed(void);
+/**
+ * Lists the files in a directory specified by the path
+ * Puts the list of file names (NOT DIRECTORIES) into the buffer seperated by newlines
+ * 
+ * \note use a path of "\" to list the files in the main directory
+ * 
+ * \return 1 on success or PROS_ERR on failure, setting errno
+ * 
+ * \b Example
+ * \code
+ * void opcontrol() {
+ * 	char* test = (char*) malloc(128);
+ *	pros::usd::list_files("/", test, 128);
+ *	pros::delay(200);
+ *	printf("%s\n", test);
+ * }
+ * \endcode
+*/
+std::int32_t list_files(const char* path, char* buffer, std::int32_t len);
 }  // namespace usd
 
 }  // namespace pros
--- a/include/pros/misc.hpp
+++ b/include/pros/misc.hpp
@@ -526,6 +526,50 @@ namespace usd {
  * \endcode
  */
 std::int32_t is_installed(void);
+
+/**
+Lists the files in a directory specified by the path
+ * Puts the list of file names (NOT DIRECTORIES) into the buffer seperated by newlines
+ * 
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * 
+ * EIO - Hard error occured in the low level disk I/O layer
+ * EINVAL - file or directory is invalid, or length is invalid
+ * EBUSY - THe physical drinve cannot work
+ * ENOENT - cannot find the path or file
+ * EINVAL - the path name format is invalid
+ * EACCES - Access denied or directory full
+ * EEXIST - Access denied
+ * EROFS - SD card is write protected 
+ * ENXIO - drive number is invalid or not a FAT32 drive
+ * ENOBUFS - drive has no work area
+ * ENFILE - too many open files
+ * 
+ * 
+ *  
+ * \note use a path of "\" to list the files in the main directory NOT "/usd/"
+ *  DO NOT PREPEND YOUR PATHS WITH "/usd/"
+ * 
+ * \return 1 on success or PROS_ERR on failure, setting errno
+ * 
+ * \b Example
+ * \code
+ * void opcontrol() {
+ * 	char* test = (char*) malloc(128);
+ *	pros::usd::list_files_raw("/", test, 128);
+ *	pros::delay(200);
+ *	printf("%s\n", test); //Prints the file names in the root directory seperated by newlines
+ *  pros::delay(100);
+ *  pros::list_files_raw("/usd/test", test, 128);
+ *	pros::delay(200);
+ *	printf("%s\n", test); //Prints the names of files in the folder named test seperated by newlines
+ *  pros::delay(100);
+ * }
+ * \endcode
+*/
+std::int32_t list_files_raw(const char* path, char* buffer, int32_t len);
+
 /**
 Lists the files in a directory specified by the path
  * Puts the list of file names (NOT DIRECTORIES) into the buffer seperated by newlines
@@ -557,6 +601,7 @@ Lists the files in a directory specified by the path
  * \code
  * void opcontrol() {
  * 	char* test = (char*) malloc(128);
+ *  // Will return vector containing names of files in root directory
  *	std::vector<std::string> files = list_files("/", test, 128);
  *	pros::delay(200);
  *	// Given vector of std::string file names, print each file name
--- a/include/pros/misc.hpp
+++ b/include/pros/misc.hpp
@@ -549,24 +549,25 @@ Lists the files in a directory specified by the path
  * \note use a path of "\" to list the files in the main directory NOT "/usd/"
  *  DO NOT PREPEND YOUR PATHS WITH "/usd/"
  * 
- * \return 1 on success or PROS_ERR on failure setting errno
+ * \return vector of std::string of file names, if error occurs, returns vector containing
+ * one element, PROS_ERR
  * 
  * \b Example
  * \code
  * void opcontrol() {
  * 	char* test = (char*) malloc(128);
- *	pros::usd::list_files("/", test, 128);
+ *	std::vector<std::string> files = list_files("/", test, 128);
  *	pros::delay(200);
- *	printf("%s\n", test); //Prints the file names in the root directory seperated by newlines
- *  pros::delay(100);
- *  pros::list_files("/test", test, 128);
- *	pros::delay(200);
- *	printf("%s\n", test); //Prints the names of files in the folder named test seperated by newlines
- *  pros::delay(100);
+ *	// Given vector of std::string file names, print each file name
+ *  // Note that if there is an error, the vector will contain one element, PROS_ERR
+ *  // Print each file name
+ *	for (std::string file : files) {
+ *		std::cout << file << std::endl;
+ *	}
  * }
  * \endcode
 */
-std::int32_t list_files(const char* path, char* buffer, std::int32_t len);
+std::vector<std::string> list_files(const char* path, char* buffer, int32_t len);
 }  // namespace usd
 
 }  // namespace pros
--- a/src/devices/vdml_usd.cpp
+++ b/src/devices/vdml_usd.cpp
@@ -21,5 +21,9 @@ std::int32_t is_installed(void) {
 	return usd_is_installed();
 }
 
+int32_t list_files(const char* path, char* buffer, int32_t len) {
+	return list_files(path, buffer, len);
+}
+
 }  // namespace usd
 }  // namespace pros
--- a/src/devices/vdml_usd.cpp
+++ b/src/devices/vdml_usd.cpp
@@ -22,7 +22,7 @@ std::int32_t is_installed(void) {
 }
 
 int32_t list_files(const char* path, char* buffer, int32_t len) {
-	return list_files(path, buffer, len);
+	return usd_list_files(path, buffer, len);
 }
 
 }  // namespace usd
--- a/src/devices/vdml_usd.cpp
+++ b/src/devices/vdml_usd.cpp
@@ -21,10 +21,14 @@ std::int32_t is_installed(void) {
 	return usd_is_installed();
 }
 
+std::int32_t list_files_raw(const char* path, char* buffer, int32_t len) {
+	return usd_list_files_raw(path, buffer, len);
+}
+
 std::vector<std::string> list_files(const char* path, char* buffer, int32_t len) {
 	std::vector<std::string> files = {};
 	// Call the C function
-	int32_t success = usd_list_files(path, buffer, len);
+	int32_t success = usd_list_files_raw(path, buffer, len);
 	// Check if call successful, if PROS_ERR return vector containing PROS_ERR
 	if (success == PROS_ERR) {
 		// push_back PROS_ERR to files vector
--- a/src/devices/vdml_usd.cpp
+++ b/src/devices/vdml_usd.cpp
@@ -21,8 +21,39 @@ std::int32_t is_installed(void) {
 	return usd_is_installed();
 }
 
-int32_t list_files(const char* path, char* buffer, int32_t len) {
-	return usd_list_files(path, buffer, len);
+std::vector<std::string> list_files(const char* path, char* buffer, int32_t len) {
+	std::vector<std::string> files = {};
+	// Call the C function
+	int32_t success = usd_list_files(path, buffer, len);
+	// Check if call successful, if PROS_ERR return vector containing PROS_ERR
+	if (success == PROS_ERR) {
+		// push_back PROS_ERR to files vector
+		files.push_back(std::to_string(success));
+		return files;
+	}
+	// Parse buffer given call successful, split by '/n'
+	// Store char * buffer as std::string in str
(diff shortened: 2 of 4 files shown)


include/pros/misc.hpp
```diff
@@ -549,24 +549,25 @@ Lists the files in a directory specified by the path
...
- *	printf("%s\n", test); //Prints the file names in the root directory seperated by newlines
- *  pros::delay(100);
- *  pros::list_files("/test", test, 128);
- *	pros::delay(200);
- *	printf("%s\n", test); //Prints the names of files in the folder named test seperated by newlines
- *  pros::delay(100);
+ *	// Given vector of std::string file names, print each file name
+ *  // Note that if there is an error, the vector will contain one element, PROS_ERR
+ *  // Print each file name
+ *	for (std::string file : files) {
```
djava, line 611:
```cpp
for (std::string& file : files) {
```
That `&` is very important for performance reasons.

include/pros/misc.hpp
```diff
@@ -549,24 +549,25 @@ Lists the files in a directory specified by the path
...
+ *		std::cout << file << std::endl;
+ *	}
  * }
  * \endcode
 */
```
djava, line 616: Surely the docs should explain the params as well?

src/devices/vdml_usd.cpp
```diff
@@ -21,5 +21,80 @@ std::int32_t is_installed(void) {
 	return usd_is_installed();
 }
 
+std::int32_t list_files_raw(const char* path, char* buffer, int32_t len) {
+	return usd_list_files_raw(path, buffer, len);
+}
+
+std::vector<std::string> list_files(const char* path, char* buffer, int32_t len) {
```
djava, line 28: This function should definitely not require users to pass in their own `buffer` and `len`. You should just allocate a buffer yourself at the top of this function body then free it after it's done being used.

src/devices/vdml_usd.cpp
```diff
@@ -21,5 +21,80 @@ std::int32_t is_installed(void) {
...
+				files.push_back("FILE I/O ERROR");
+				break;
+		}
+		return files;
+	}
+	// Parse buffer given call successful, split by '/n'
+	// Store char * buffer as std::string in str
+	// Store delimiter in std::string, '\n'
+	std::string str(buffer);
+	std::string delimiter = "\n";
```
djava, line 78:
```cpp
char delimeter = '\n';
```

You definitely don't need a whole `std::string` for just `\n`, which is only a single char.

src/devices/vdml_usd.cpp
```diff
@@ -21,5 +21,80 @@ std::int32_t is_installed(void) {
...
+				// If none of the above, will be FILE IO error
+				files.push_back("FILE I/O ERROR");
+				break;
+		}
+		return files;
+	}
+	// Parse buffer given call successful, split by '/n'
+	// Store char * buffer as std::string in str
+	// Store delimiter in std::string, '\n'
+	std::string str(buffer);
```
djava, line 77: There's no reason to allocate a whole additional `std::string` for this, use `std::string_view` over the already-existing buffer instead. You can still use `.find()` the same way.

src/devices/vdml_usd.cpp
```diff
@@ -21,5 +21,80 @@ std::int32_t is_installed(void) {
...
+	// file_name used to store each file name
+	size_t position = 0;
+	std::string file_name;
+
+	// Loop until delimiter '\n' can not be found anymore
+	while ((position = str.find(delimiter)) != std::string::npos) {
+		// file_name is the string from the beginning of str to the first '\n', excluding '\n'
+		file_name = str.substr(0, position);
+		// Add token to files vector
+		files.push_back(file_name);
```
djava, line 90: You should move file_name into the vector instead of copying it.  This will reuse the already allocated heap buffer instead of copying over a new one, resulting in each file name being copied once instead of twice or more.

```cpp
files.emplace_back(std::move(file_name));
```

Alternatively, `file_name` could be a `std::string_view` that you `emplace_back`. Since `files::value_type` is `std::string`, that will result in the file name being copied (for the first time). Actually I think I like this solution better than the move anyway.

src/devices/vdml_usd.cpp
```diff
@@ -21,5 +21,80 @@ std::int32_t is_installed(void) {
...
+		// Erase file_name from str, would be pos + 1 if we wanted to include '\n'
+		str.erase(0, position + 1);
```
djava, line 92: This won't work if you use a `std::string_view` for str (and it's very unnecessarily slow even as a `std::string`). You should either keep track of the index to start at for the next iteration, or maybe `str = str.substr(...)`?

src/devices/vdml_usd.cpp
```diff
@@ -21,5 +21,80 @@ std::int32_t is_installed(void) {
...
+}
+
+std::vector<std::string> list_files(const char* path, char* buffer, int32_t len) {
+	std::vector<std::string> files = {};
+	// Call the C function
+	int32_t success = usd_list_files_raw(path, buffer, len);
+	// Check if call successful, if error return vector containing error state
+	if (success == PROS_ERR) {
+		// Check errno to see which error state occurred
+		// push back error state to files vector as std::string
```
djava, line 35: mmm I'm definitely not convinced that this error handling strategy is a good approach.  Maybe an out-param for the vector is better so you have some other way to return errors? That's more inline with how pros usually does things anyway. (returning errors via return type)

djava (changes requested):

djava:
@Gracelu128 it would be helpful if, when you make changes to address review comments I've made, you could respond to the review comment to let me know (maybe with the commit hash  that you addressed it in as well). Would you be able to go through and do that here so it's easier to dig through this pile of commits?

Thanks!

include/pros/misc.hpp
Gracelu128: Implemented this suggestion in commit on Oct 30, 3d41ca62de29da852509b02fcd8ce85c94a92e60

include/pros/misc.hpp
Gracelu128: Added param in header for funcs in commit 3d41ca62de29da852509b02fcd8ce85c94a92e60 and 4247fb197e02a639c26dd75961c81d81e3e5b30a

src/devices/vdml_usd.cpp
Gracelu128: Implemented this change in commit 4247fb197e02a639c26dd75961c81d81e3e5b30a

src/devices/vdml_usd.cpp
Gracelu128: Implemented this change in commit 3d41ca62de29da852509b02fcd8ce85c94a92e60

src/devices/vdml_usd.cpp
Gracelu128: Implemented this change in commit 3d41ca62de29da852509b02fcd8ce85c94a92e60.

src/devices/vdml_usd.cpp
Gracelu128: Implemented this change in commit 3d41ca62de29da852509b02fcd8ce85c94a92e60. Made file_name a std::string_view.

src/devices/vdml_usd.cpp
Gracelu128: Implemented this change in commit 3d41ca62de29da852509b02fcd8ce85c94a92e60. Altered code to instead have index to keep track of start for next iteration.

src/devices/vdml_usd.cpp
djava: Looks good, can I ask if the number 10000 came from somewhere or was it chosen arbitrarily?

src/devices/vdml_usd.cpp
```diff
@@ -25,10 +25,12 @@ std::int32_t list_files_raw(const char* path, char* buffer, int32_t len) {
...
 }
 
-std::vector<std::string> list_files(const char* path, char* buffer, int32_t len) {
+std::vector<std::string> list_files(const char* path) {
 	std::vector<std::string> files = {};
+	// malloc buffer to store file names
+	char *buffer = (char *) malloc(10000);
 	// Call the C function
-	int32_t success = usd_list_files_raw(path, buffer, len);
+	int32_t success = usd_list_files_raw(path, buffer, 10000);
```
djava: Could you have a `constexpr size_t buffer_size = 10000;` to reuse between these two instances?

src/devices/vdml_usd.cpp
noam987: Came from an estimate of the worst case that is realistic to happen. We needed to chose a number, and decided that 100 files in a single directory with an approximate average length of 100 (really 99 if you include the separators) will be enough for every feasible use case we could come up with.

src/devices/vdml_usd.cpp
```diff
@@ -71,29 +71,30 @@ std::vector<std::string> list_files(const char* path, char* buffer, int32_t len)
...
+	while ((delimiter_pos = str.find('\n', index)) != std::string::npos) {
 		// file_name is the string from the beginning of str to the first '\n', excluding '\n'
-		file_name = str.substr(0, position);
+		file_name = std::string_view(str.data() + index, delimiter_pos - index);
 		// Add token to files vector
-		files.push_back(file_name);
-		// Erase file_name from str, would be pos + 1 if we wanted to include '\n'
-		str.erase(0, position + 1);
-	}
+		files.emplace_back(file_name);
```
djava, line 73: Since it happens via an implicit conversion, it would be nice to have a comment here noting that this is the point where the string containing the file name is copied into its own `std::string`, since everything before this was just `std::string_view`s over the pre-existing `buffer`.

src/devices/vdml_usd.cpp
```diff
@@ -71,29 +71,30 @@ std::vector<std::string> list_files(const char* path, char* buffer, int32_t len)
...
-	std::string delimiter = "\n";
+	std::string_view str(buffer);
 
-	// position to keep track of position in str
+	// delimter_pos is the position of the delimiter '\n'
+	// index of which character to start substr from
 	// file_name used to store each file name
-	size_t position = 0;
-	std::string file_name;
+	size_t delimiter_pos, index = 0;
```
djava, line 65: C++ footgun alert!! `index` is initialized to 0 here, but `delimiter_pos` is left uninitialized (and therefore in an unspecified state). Admittedly, it doesn't matter here since the first use is an assignment, but I generally advise to avoid multiple declaration and always initialize primitive types to avoid even the potential of such mistakes causes issues, since those can be very hard to track down later.

src/devices/vdml_usd.cpp
djava: Very nice! Thanks

src/devices/vdml_usd.cpp
```diff
@@ -29,6 +29,15 @@ std::vector<std::string> list_files(const char* path) {
 	std::vector<std::string> files = {};
 	// malloc buffer to store file names
 	char *buffer = (char *) malloc(10000);
+	if (buffer == NULL) {
+		// try again smaller buffer to see if that works
+		buffer = (char *) malloc(500);
```
djava, line 34: Keeping a `buffer_size` variable is even more pertinent with this addition, wouldn't want to put 10000 entries into a buffer of size 500! It can't be `constexpr` anymore now though, so ignore that part of the above comment.

djava (changes requested):

djava:
Looks a lot better, thanks for the changes! Most of what's left is small or stylistic.

src/devices/vdml_usd.cpp
djava: Ok, fair enough. If you're choosing a number, a power of 2 (or sum of powers of 2) might be nice for memory allocator friendliness :)

(I have no idea what allocation strategy VexOS uses, this may be completely irrelevant)

djava:
Just a thought - right now there's no standard prefix to the error messages, so there's not really a clean way to programmatically check if the function errored. Perhaps they can all start with "ERROR - " or similar, so that one can at least write something like this to check for errors?
```cpp
auto file_list = pros::usd::list_files("/");
if (file_list.front().starts_with("ERROR")) {
    // do error stuff
} else {
    // We know file_list is valid now
}```

I'm not sure how you'd do this as is. `errno` isn't reliable, since it could have been set by a previous function call, and theres no return code from this function that tells us succinctly that we have to check errno, right?

Gosh, thank god for C++20 adding `std::string::starts_with`. Not sure what the clean way to do this otherwise would be.

**Comment by Gracelu128** [comment]
> Looks a lot better, thanks for the changes! Most of what's left is small or stylistic.

I just implemented your suggested stylistic changes in the new commit 490a3ca83be775d3e2c97da582e00099efb6393a.

File: src/devices/vdml_usd.cpp [hunk 13]
```diff
@@ -21,5 +21,77 @@ std::int32_t is_installed(void) {
...
 
+std::int32_t list_files_raw(const char* path, char* buffer, int32_t len) {
+	return usd_list_files_raw(path, buffer, len);
+}
+
+std::vector<std::string> list_files(const char* path) {
+	std::vector<std::string> files = {};
+	// malloc buffer to store file names
+	size_t buffer_size = 10000;
+	char *buffer = (char *) malloc(buffer_size);
```
djava, line 32: You should use the C++ `new char[buffer_size];`instead of C's `malloc` (and later `delete`instead of `free).

It's a bit less scary type-wise and very widely considered better stylistically for C++.

djava: Actually even better would be `std::unique_ptr<char[]> buffer = std::make_shared<char[]>(buffer_size)`and then passing `buffer.get()` to the C function. Then the `delete`happens automatically when you exit the function scope.

djava (changes requested):

WillXuCodes (changes requested):
this review might be slightly outdated... please ignore any outdated comments I just forgot to hit submit 💀

src/devices/vdml_usd.cpp
```diff
@@ -21,5 +21,90 @@ std::int32_t is_installed(void) {
...
+
+	// delimter_pos is the position of the delimiter '\n'
+	// index of which character to start substr from
+	// file_name used to store each file name
+	size_t delimiter_pos = 0;
+	size_t index = 0;
+	std::string_view file_name;
+
+	// Loop until delimiter '\n' can not be found anymore
+	while ((delimiter_pos = str.find('\n', index)) != std::string::npos) {
```
WillXuCodes, line 87: If we really wanted to optimize this even more you could write a sentinel function (or whatever people call it I'm tired atm) that grabs all the delimter indexes, and then another loop that stores all the delimter positions in a data structure such as a vector, and then another loop that iterates through the data structure. This changes the O(N * M) behavior of this to O(N + M) where N is the size of the string and M is the number of delimeters.

src/devices/vdml_usd.cpp
```diff
@@ -43,6 +43,19 @@ std::vector<std::string> list_files(const char* path) {
...
 	}
+
+	// Check path user passed in
+	std::string_view path_sv(path);
+	size_t found = path_sv.find("usd");
+	if (found == 0 || found == 1) {
+		// Deal with when user prepends path with usd
+		// as either "usd/..." or "/usd/..."
+		path_sv.remove_prefix(3);
+	}
```
djava, line 54: How does hardcoding the `3` here deal with both cases? Wouldn't it be `3 + found`?

Also to avoid the magic-number-ness of this 3, I'd probably write this as:
```cpp
constexpr std::string_view usd_prefix {"usd"};
std::string_view path_sv {path};
const size_t usd_prefix_idx = path_sv.find(usd_prefix);
if (found == 0 || found == 1) {
    path_sv.remove_prefix(usd_prefix.length() + usd_prefix_idx);
}
```

src/devices/vdml_usd.cpp
```diff
@@ -43,6 +43,19 @@ std::vector<std::string> list_files(const char* path) {
...
+
+	// set path to path_sv.data()
+	path = path_sv.data();
```
djava, line 57: `path_sv` is already pointing to `*path`. It's a read-only "view" over a string that exists elsewhere in memory, hence the name.

src/devices/vdml_usd.cpp
```diff
@@ -76,7 +89,7 @@ std::vector<std::string> list_files(const char* path) {
 		file_name = std::string_view(str.data() + index, delimiter_pos - index);
 		// This is point where content of the std::string_view file name is copied to its
 		// own std::string and added to the files vector
-		files.emplace_back(file_name);
+		files.emplace_back("/usd" + std::string(path) + "/" + std::string(file_name));
```
djava, line 92: You can't do it like this, this creates 7 different `std::string`s (each of which "may" allocate and pretty likely at least 2 will, wasting a lot of performance).
I don't believe the version of libstdc++ (the standard library) PROS uses supports `std::format` but if so you could write the very nice:
```cpp
files.emplace_back(std::format("/{}/{}", path, file_name));
```
if not i would recommend something more like
```cpp
const size_t path_size = 1 + usd_prefix.length() + path.length() + 1 + file_name.length() + 1; // +1's for the slashes, null terminator 
std::string buf;
buf.reserve(path_size); // Reserve enough space in the internal array that we can snprintf right into it

// The %.*s format specifier allows passing the size of a format arg before the value of the arg, so it lets
// us use the string_view into `str` with the C-style snprintf function even though it does not have the
// null terminator that would usually be required to know when a string ends.
std::snprintf(buf.data(), path_size, "/%s/%.*s", path, static_cast<int>(file_name.length()), file_name.data());

files.push_back(std::move(buf));
```
This retains the property of the old version that it only had to allocate the string once.

djava (changes requested):

src/devices/vdml_usd.cpp
```diff
@@ -21,5 +21,80 @@ std::int32_t is_installed(void) {
 	return usd_is_installed();
 }
 
+std::int32_t list_files_raw(const char* path, char* buffer, int32_t len) {
```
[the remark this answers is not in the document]
djava: std::string_view

src/devices/vdml_usd.cpp
djava: I'm not sure where you got `O(N*M)`.  If I'm reading correctly, you're seeing it as a full loop through all N characters each time we loop? That's not true, though, as `index` changes where the iteration starts each time (starting further along the string on each iteration), and each iteration stops as soon as it locates `'\n'`. Therefore, this is only O(N).

WillXuCodes: > I'm not sure where you got `O(N*M)`. If I'm reading correctly, you're seeing it as a full loop through all N characters each time we loop? That's not true, though, as `index` changes where the iteration starts each time (starting further along the string on each iteration), and each iteration stops as soon as it locates `'\n'`. Therefore, this is only O(N).

ah you're right, I didn't notice the index parameter last night.

src/devices/vdml_usd.cpp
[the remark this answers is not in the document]
WillXuCodes: > std::string_view

I understand that's more C++onic but for the sake of our users I think a std::string would be a little more approachable... saves everyone a google search.

src/devices/vdml_usd.cpp
WillXuCodes: > I don't believe the version of libstdc++ (the standard library) PROS uses supports std::format

Sadly it does not :(

src/devices/vdml_usd.cpp
djava: Also what happens here if they don't include the usd prefix? We just make it work the same as if they did? May be weird, since we also always _return_ paths prefixed with `/usd/`.

My preference here would be to return a "path not found" error, as just assuming they wanted an implicit `/usd/` makes doing anything with a VFS outside of `/usd/` in the future into a backwards compatibility break.

src/devices/vdml_usd.cpp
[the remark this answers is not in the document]
djava: 😖 wasted allocation for every call with a string literal for path , which I assume is the typical case.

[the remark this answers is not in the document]
djava: > make sure all of the std::strings are also references pls*
If you choose to use `std::string` instead of `string_view` do make sure its `const` refs, as mutable refs will cause other issues (won't accept temporaries i.e. string literals that convert to `std::string`)

src/devices/vdml_usd.cpp
djava: @Gracelu128 Ty but
1. I would recommend keeping the comments (or rewriting them in your own words if you want). Always better to explain what you're doing, especially if it may be non-obvious why (and I think readers of this code may not immediately understand what the `%.*s` does, and why its needed. I know I wouldn't know what that format specifier means at first glance)
2. Sorry, I actually made an edit to the code block after you looked at it I think:
> (EDIT: removed the + 1 in path_size calculation for null terminator, as I realized that std::string::reserve implicitly adds 1 for that by itself, and add it back in snprintf)

You could also just subtract 1 from the param to the `reserve` call if you think thats preferable.

PR #618, opened 2023-11-13
🐛Fix Negative Port Rotation Sensor

Cooper7196:
#### Summary:
Fixed the rotation sensor not working when initialized with a negative port number

#### Motivation:
The rotation sensor currently does not work when using a negative port number to reverse it during initialization

#### Test Plan:

- [ ] Verify that this PR fixes the issue

### DIFF
--- a/src/devices/vdml_rotation.cpp
+++ b/src/devices/vdml_rotation.cpp
@@ -15,13 +15,13 @@
 
 namespace pros {
 inline namespace v5 {
-    
-Rotation::Rotation(const std::int8_t port) : Device(port, DeviceType::rotation) {
-    if (port < 0) {
-        pros::c::rotation_set_reversed(abs(port), true);
-    } else {
-        pros::c::rotation_set_reversed(port, false);
-    }
+
+Rotation::Rotation(const std::int8_t port) : Device(abs(port), DeviceType::rotation) {
+	if (port < 0) {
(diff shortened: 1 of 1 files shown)


noam987 (approved):
LGTM but needs to be hardware tested. We don't currently have one in the lab, so I would appreciate it if you can test it and get back to me with the results. Let me know if you need some minimal viable code to test with.

WillXuCodes:
Also please add the 🐛 to the beginning of the pull request name, it just helps us keep track of commit types. Thanks for your contribution!

noam987:
We finally got a rotation sensor. Will test this myself ASAP and merge it in

src/devices/vdml_rotation.cpp
```diff
@@ -77,9 +77,9 @@ std::ostream& operator<<(std::ostream& os, const pros::Rotation& rotation) {
 
 namespace literals {
 const pros::Rotation operator"" _rot(const unsigned long long int r) {
```
noam987, line 79: Now that we support making rotation sensors with negative ports, should this be changed to signed? Probably outside the scope of this pr

noam987 (approved):
LGTM but Will test tomorrow in the lab

src/devices/vdml_rotation.cpp
WillXuCodes: yes

PR #619, opened 2023-11-17
Fflush on write

aberiggs:
#### Summary:
Currently working on an fflush for files on an sd card to be integrated into standard IO; however, this modifies the write function to flush the buffer when writing to a file.

#### Motivation:
No working fflush adjacent function for file IO on the sd card currently.


noam987 (approved):
LGTM

PR #621, opened 2023-12-08
✨Allows for file open to default to the sd card.

noam987:
#### Summary:
Allows for open for the sd card to no longer need paths to start with "/usd/" UNLESS the path begins with "usd", "dev", or "ser".
TO access those folders you still need the usd prefix as before. 
Some examples: 
if there is a file called a.txt in the main folder of the sd card, it can be accessed via a path of "/usd/a.txt" or "a.txt" would work.
However, if there is a file called b.txt in a folder called dev, it must be accessed via the path of "/usd/dev/b.txt"

#### Motivation:
Makes file IO via the SD card way easier for users
#### Test Plan:

- [ ] tested opening a file without prefixed with "/usd/". Tested with both a leading / and without it


aberiggs (approved):
LGTM

Issue #626, opened 2024-01-19
🐛[BUG] - `Imu::is_calibrating` returns `true` if imu is not connected


SizzinSeal:
**Describe the bug**
`Imu::is_calibrating` returns `true` if imu is not connected

**To Reproduce**
1. Instantiate an IMU on a port where it is not connected
2. call `Imu::is_calibrating()`

**Expected behavior**
It should return `false`

**Screenshots**

**Desktop (please complete the following information):**
 - PROS 3.8.0

WillXuCodes:
Closed by #628

WillXuCodes:
Not closing this yet due to #628 not being merged yet.

liam-teale:
Reproduced in kernel version 3.8.3

PR #628, opened 2024-02-03
🐛Fixed Imu::is_calibrating function for PROS 3 #626

Gracelu128:
#### Summary:
Fixed Imu::is_calibrating function to return correct values.

#### Motivation:
Users can use this function to properly check if their connected imu is calibrating or not, and also get the proper return value of false if they call this function on an imu that is not connected.

#### Test Plan:

- [x] Tested function on imu that is not connected
- [x] Tested function on connected imu but not calibrating
- [x] Tested function on connected imu that is calibrating

### DIFF
--- a/include/pros/imu.h
+++ b/include/pros/imu.h
@@ -29,8 +29,8 @@ namespace c {
 #endif
 
 typedef enum imu_status_e {
-	E_IMU_STATUS_READY = 0,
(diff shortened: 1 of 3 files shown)


WillXuCodes:
Closes: #626

noam987 (changes requested):
Revert main.cpp and then this looks good

PR #629, opened 2024-02-09
🐛Fixed Imu::is_calibrating function for PROS 4 #626

Gracelu128:
#### Summary:
Fixed Imu::is_calibrating function to return correct values.

#### Motivation:
Users can use this function to properly check if their connected imu is calibrating or not, and also get the proper return value of false if they call this function on an imu that is not connected.

#### Test Plan:

- [x] Tested function on imu that is not connected
- [x] Tested function on connected imu but not calibrating
- [x] Tested function on connected imu that is calibrating


noam987 (approved):
LGTM

src/devices/vdml_imu.cpp
```diff
@@ -65,7 +65,7 @@ pros::ImuStatus Imu::get_status() const {
 }
 
 bool Imu::is_calibrating() const {
-	return (int)get_status() & (int)(pros::ImuStatus::calibrating);
+	return (int)get_status() == (int)(pros::ImuStatus::calibrating);
```
cgjeffries: This introduces a bug where the condition will evaluate to false if the IMU is in any position other than upright, as the orientation bits from `get_status` will no longer exactly match the value of the `calibrating` enum. See https://github.com/purduesigbots/pros/pull/643

phinc23 (approved):
LGTM

Issue #625, opened 2024-01-16
🐛[BUG] -  An issue with the get_rotation function of the GPS module


OWmess:
The code I wrote is as follows：
```cpp
imuRaw = degToRad(gps.get_rotation());
        pros::lcd::print(3, "imuRaw:%f\n",imuRaw);
        pros::lcd::print(4, "gpsRaw:%lf\n",gps.get_rotation());
        pros::lcd::print(5, "gpsHeading:%lf\n",gps.get_heading());
        pros::lcd::print(6, "errno:%d\n",errno);
        pros::lcd::print(7,"yaw:%lf\n",gps.get_status().yaw);
```
Then I found that the gpsRaw（gps.get_rotation()） data printed on the screen is always 0.000,but the other data is normal.

Looking forward to a reply.

OWmess:
The version of PROS is 3.8.0

noam987:
Closed In #630

PR #630, opened 2024-02-09
✨ Add to GPS API

phinc23:
#### Summary:
Added Gps features

get_x_position()
get_y_position()
get_pitch()
get_roll()
get_yaw()

*Fixed*

get_rotation() // Now returns the angle unbounded (past +/-360) no long returns intial set rotation

#### Motivation:
More clear api methods and calls for gps utility 

##### References (optional):
[link](https://github.com/purduesigbots/pros/issues/625)

#### Test Plan:

- [x] Test X and Y position
- [x] Test pitch
- [x] Test roll
- [x] Test yaw

- [x] Test get_rotation


src/devices/vdml_gps.c
```diff
@@ -101,7 +146,7 @@ double gps_get_heading_raw(uint8_t port) {
 
 double gps_get_rotation(uint8_t port) {
 	claim_port_f(port - 1, E_DEVICE_GPS);
-	double rtv = vexDeviceGpsRotationGet(device->device_info);
+	double rtv = vexDeviceGpsHeadingGet(device->device_info);
```
WillXuCodes, line 149: I don't exactly approve of this change, it gives users less options and also doesn't fit the VEX API loose rule of having total rotation as simply "rotation" whereas heading and angle are used as the 360 constrained rotation.

version
```diff
@@ -1 +1 @@
-3.8.0
+3.7.4-dirty.19.1e7513d
```
WillXuCodes, line 1: be sure to revert

include/pros/gps.hpp
```diff
@@ -232,6 +232,16 @@ class Gps {
 	 */
 	virtual std::int32_t set_rotation(double target) const;
 
+	virtual double get_x_position() const;
```
WillXuCodes, line 235: be sure to doxygenize

include/pros/gps.h
```diff
@@ -188,6 +188,16 @@ double gps_get_error(uint8_t port);
  */
 gps_status_s_t gps_get_status(uint8_t port);
 
+double gps_get_x_position(uint8_t port);
+
+double gps_get_y_position(uint8_t port);
+
+double gps_get_pitch(uint8_t port);
+
```
WillXuCodes, line 196: be sure to doxygenize

include/api.h
```diff
@@ -40,9 +40,9 @@
 #endif /* __cplusplus */
 
 #define PROS_VERSION_MAJOR 3
-#define PROS_VERSION_MINOR 8
-#define PROS_VERSION_PATCH 0
-#define PROS_VERSION_STRING "3.8.0"
+#define PROS_VERSION_MINOR 7
+#define PROS_VERSION_PATCH 4
+#define PROS_VERSION_STRING "3.7.4-dirty.19.1e7513d"
```
WillXuCodes, line 45: be sure to revert

WillXuCodes (changes requested):
Biggest objection is the rotation change. All else is good.

src/devices/vdml_gps.c
phinc23: That was the intention of the get_rotation the call we were using before returning the initial set rotation. If a user set rotation to 90 degrees then gps_get_rotation would only return that value and if the user didn't set it it would return 0. The previous behavior was more get_intial_rotation. The new method returns the live unbounded rotation unconstrained

WillXuCodes: Interesting, we may want to contact VEX about this actually. This seems to be more of a bug than an intended feature.

src/main.cpp
```diff
@@ -90,4 +90,4 @@ void opcontrol() {
 
 		pros::delay(20);
 	}
-}
+}
```
WillXuCodes, line 93: remember the newline

version
```diff
@@ -1 +1 @@
-3.8.0
+3.8.0
```
WillXuCodes, line 1: ditto on newline

WillXuCodes (changes requested):

noam987 (approved):
LGTM, but we need to change the DOCS site to reflect these new methods.

WillXuCodes (dismissed):

noam987 (dismissed):
LGTM

phinc23:
Changed the degree ranges of pitch roll and yaw to correct ranges. Docs changes only.

noam987 (approved):
LGTM

PR #632, opened 2024-02-16
🐛Motor move operator fix

phinc23:
#### Summary:
Fixed an issue where users could not use the "=" operator to set a motor's voltage/speed 

#### Motivation:
Received an "Ambiguous overload for operator" error when attempting to set a motor's voltage with = operator. 

Ex. `my_motor = 127;  // Result in error`

#### Test Plan:

- [x] Test motor = [-127, 127]
- [x] Test with other methods
- [x] Test motor group operator

### DIFF
--- a/include/pros/motors.hpp
+++ b/include/pros/motors.hpp
@@ -71,7 +71,7 @@ class Motor : public AbstractMotor, public Device {
  	 * \endcode
 	 * 
 	 */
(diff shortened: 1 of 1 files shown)


noam987:
Can you test if this is also the case for motor groups as well and fix it if so?

noam987 (approved):
LGTM for now. Might get reverted by a future PR to address the root cause

noam987:
Feature is being removed in PROS 4 with PR #652

Issue #631, opened 2024-02-14
🐛[BUG] - Addressible LEDs are broken.


Tropix126:
**Describe the bug**
The `ADILed` device setters appear to be broken in a few ways.
- `ext_adi_led_set_pixel` returns the [wrong port mutex](https://github.com/purduesigbots/pros/blob/1e7513d4f110d2eac625b6300dbbb8c086ab6c0c/src/devices/vdml_ext_adi.c#L493) on L493. It essentially tries to zero-index a port that was already converted to zero-index, since `get_ports` already transforms the smart port outvalue. As a result, it tries to release a mutex on the port *below* what the LEDs are actually plugged into. In many cases this can permanently deadlock the user program (from past testing).
- All other setters have a strange bug where it takes calling them twice, with a delay between calls to properly update the hardware state.  The minimum delays between two setter calls seems to be around 70mS.

**To Reproduce**
This does not work
```cpp
void initialize() {
  led.set_all(0xFFFFFF);
}```

This does not work either
```cpp
void initialize() {
  led.set_all(0xFFFFFF);
  led.set_all(0xFFFFFF);
}```

This does
```cpp
void initialize() {
  led.set_all(0xFFFFFF);
  pros::delay(70);
  led.set_all(0xFFFFFF);
}```

This is likely an SDK bug. Sylib gets around this by updating LEDs in a daemon loop rather than just setting once.

**Expected behavior**
LEDs should update the first time the setter is called.

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Windows
 - Version 11

**Additional context**
None

PR #633, opened 2024-02-23
🐛ADI mutex fix

R11G:
#### Summary:
Fixes extraneous zero indexing in ext_adi_led_set_pixel and validate_type calls.

##### References:
#631 

#### Test Plan:

ADI LEDs should be able to be set the first time without deadlocking and without needing delays.
Validate type upon finding a type error should release the correct ADI port mutex.


src/devices/vdml_ext_adi.c
```diff
@@ -290,7 +290,7 @@ int32_t ext_adi_encoder_get(ext_adi_encoder_t enc) {
 	get_ports(enc, smart_port, adi_port);
 	transform_adi_port(adi_port);
 	claim_port_i(smart_port, E_DEVICE_ADI);
```
noam987: why do you claim the smart_port mutex instead of smart_port -1 mutex like the previous functions do?

noam987 (approved):
LGTM

PR #635, opened 2024-03-08
🐛Fix motor voltage limit port mutex returning

Gavin-Niederman:
#### Summary:

This fixes an issue where a port mutex was not returned when calling `motor_get_voltage_limit`.  

#### Motivation:

Calling `motor_get_voltage_limit` on a port that has a motor plugged in would break any subsequent motor functions on that port as well as the serial terminal.

#### Test Plan:

- [ ] Plug in a motor and call `motor_get_voltage_limit` on that port. Then check if printing to the serial terminal and calling any more motor methods on that port works.


max-niederman (approved):

noam987 (approved):
LGTM

PR #637, opened 2024-03-12
🐛LVGL display update task at equal priority to user tasks

noam987:
#### Summary:
Changes the delay of the LVGL background daemon to be 8ms from 2 ms
Updates the priority of the LVGL background daemon to be equal priority to user tasks 

#### Motivation:
Currently the lvgl display daemon runs at a much lower priority than user tasks. Since PROS does not do priority upgrading when a task is starved, the lvgl daemon doesn't run if users don't have delays in their code. 
This PR upgrades the LVGL display daemon to equal priority to user tasks. In order to prevent LVGL from hogging the resources, the delay in the lvgl display daemon is upped from 2 ms to 8 ms. This still should give LVGL more resources when users don't have delays in their code, while slightly reducing the amount of cpu time LVGL uses when users use delays properly. LVGL should still have more than enough resources as the brain's refresh rate corresponds to a delay of 16.6ms in the display daemon.

#### Test Plan:

- [x] Screen is kept up to date when a user task with delays is running. 
- [ ] Screen is kept up to date when a user task without delays is running.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,8 +51,8 @@ EXTRA_LIB_DEPS=$(INCDIR)/api.h $(PATCHED_SDK)
 -include ./common.mk
 
 .PHONY: $(INCDIR)/api.h
-# $(INCDIR)/api.h: version.py
-# 	$(VV)python version.py
+$(INCDIR)/api.h: version.py
+	$(VV)python version.py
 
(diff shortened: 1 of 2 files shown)


noam987:
Closing this for now as it didn't pass hardware testing. (Although it's behavior is arguably better than the current develop behavior)

PR #642, opened 2024-03-16
🐛Fix gps comments

Rocky14683:
#### Summary:
Fix gps comments

#### Motivation:
截圖 2024-03-16 下午5 18 53

截圖 2024-03-16 下午5 20 38

<br>
Add periods to separate two sentences shown in PROS 4 doc. And, change "an" GPS to "a" GPS .


phinc23 (approved):
Good catch. LGTM :rock: :rock:

PR #643, opened 2024-03-27
🐛Imu Reset bugfix and IMU get orientation

noam987:
#### Summary:

Fixes the issues with the IMU reset never returning
Adds functionality to get the Orientation of the IMU

#### Motivation:

Currently on pros 3.8.2 IMU can only calibrate in the Z_DOWN orientation. The reset function is an infinite loop otherwise. 
Adds functionality to get the Physical Orientation of the IMU. 

#### Test Plan:

- [x] IMU can calibrate in all orientations
- [x] IMU can return the correct orientation from each orientation

### DIFF
--- a/include/pros/imu.h
+++ b/include/pros/imu.h
@@ -545,7 +545,7 @@ int32_t imu_set_roll(uint8_t port, double target);
 int32_t imu_set_yaw(uint8_t port, double target);
 
 /**
- * Returns the orientation of the IMU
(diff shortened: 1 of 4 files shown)


Rocky14683 (dismissed):
LGTM

Rocky14683:
Suggestion: Change get_orientation to get_physical_orientation

Rocky14683 (approved):
LGTM

PR #650, opened 2024-04-16
Modify PROS 4 GPS API

CChheerryyll:
#### Summary:

This update modifies the GPS sensor API for PROS 4.
Modified list of functions:
- initialize_full()
- set_offset()
- get_offset()
- set_position()
- set_data_rate()
- get_error()
- get_position_and_orientation()
- get_position()
- get_position_x()
- get_position_y()
- get_orientation()
- get_pitch()
- get_roll()
- get_yaw()
- get_heading()
- get_heading_raw()
- get_gyro_rate()
- get_gyro_rate_x()
- get_gyro_rate_y()
- get_gyro_rate_z()
- get_accel()
- get_accel_x()
- get_accel_y()
- get_accel_z()

Removed functions:
- get_rotation()
- set_rotation()
- tare_rotation()

#### Motivation:

The purpose of some functions are unclear (e.g. get_rotation), and the name of some functions are not straightforward enough (e.g. get_status)

#### Test Plan:

- [ ] call the functions individually and print their return values


src/devices/vdml_gps.c
```diff
@@ -167,3 +227,27 @@ gps_accel_s_t gps_get_accel(uint8_t port) {
...
+	return_port(port - 1, rtv);
+}
+
+double gps_get_accel_z(uint8_t port) {
+    claim_port_f(port - 1, E_DEVICE_GPS);
+	V5_DeviceGpsRaw data;
+	vexDeviceGpsRawAccelGet(device->device_info, &data);
+	double rtv = data.z;
+	return_port(port - 1, rtv);
+}
```
WillXuCodes: newline at end of file is missing.

version
```diff
@@ -1 +1 @@
-4.0.6
\ No newline at end of file
+4.0.4-dirty.28.4e2730a
```
WillXuCodes: need newline and need to revert this.

src/devices/vdml_gps.cpp
```diff
@@ -57,34 +81,45 @@ double Gps::get_heading_raw() const {
 	return pros::c::gps_get_heading_raw(_port);
 }
 
-double Gps::get_rotation() const {
-	return pros::c::gps_get_rotation(_port);
```
WillXuCodes, line 61: I understand that rotation isn't working as we thought, but is it possible to have some way to set the heading so it can at least be tared/zero'd by a hard reset? I believe that is still a valid use case.

include/api.h
```diff
@@ -41,8 +41,8 @@
 
 #define PROS_VERSION_MAJOR 4
 #define PROS_VERSION_MINOR 0
-#define PROS_VERSION_PATCH 6
-#define PROS_VERSION_STRING "4.0.6"
+#define PROS_VERSION_PATCH 4
+#define PROS_VERSION_STRING "4.0.4-dirty.28.4e2730a"
```
WillXuCodes: revert

src/devices/vdml_gps.c
```diff
@@ -110,34 +110,70 @@ gps_position_s_t gps_get_position(uint8_t port) {
...
+
+double gps_get_heading_raw(uint8_t port) {
+	claim_port_f(port - 1, E_DEVICE_GPS);
+	double rtv = vexDeviceGpsHeadingGet(device->device_info);
+	return_port(port - 1, rtv);
 }
 
 gps_gyro_s_t gps_get_gyro_rate(uint8_t port) {
```
WillXuCodes, line 179: I know this is completely my fault but can we see if typecasting V5_DeviceGPSRaw to the return type works, I know what I did here was super scuffed and I'd appreciate if we just casted it instead of doing the extraneous copy.

src/devices/vdml_gps.c
```diff
@@ -81,7 +81,7 @@ double gps_get_error(uint8_t port) {
 	return_port(port - 1, rtv);
 }
 
-gps_status_s_t gps_get_status(uint8_t port) {
+gps_status_s_t gps_get_position_and_orientation(uint8_t port) {
 	gps_status_s_t rtv = GPS_STATUS_ERR_INIT;
 	if (!claim_port_try(port - 1, E_DEVICE_GPS)) {
 		return rtv;
```
WillXuCodes, line 87: Another alternative is typecasting the input parameter (data) from gps_status_s_t or whatever it is to the VEX API struct.

WillXuCodes: Outside of the scope of this PR but the IMU API may have the same issue.

WillXuCodes (changes requested):

src/devices/vdml_gps.cpp
CChheerryyll: I think set_position() is able to do that since it takes heading as a parameter? I'll test the function again to see if it works that way.

CChheerryyll: set_position() is able to reset the heading through set_position(get_position_x(), get_position_y(), 0)

src/devices/vdml_gps.c
CChheerryyll: Can I have more context on this? vexDeviceGpsHeadingGet() already returns a double value, I don't see a need for typecasting.

CChheerryyll: Can I have more context on this? vexDeviceGpsHeadingGet() already returns a double value, I don't see a need for typecasting.

src/devices/vdml_gps.c
CChheerryyll: This function currently converts the VEX API struct (V5_DeviceGpsAttitude) into gps_status_s_t, and returns the gps_status_s_t. If we want to cast gps_status_s_t into V5_DeviceGpsAttitude, should the return type be just V5_DeviceGpsAttitude then?

noam987 (approved):
Really like the docs changes. LGTM

PR #651, opened 2024-04-22
✨ Print stack trace to brain screen on data/prefetch abort

ion098:
#### Summary:
This PR prints the stack addresses to the brain screen during a data abort or prefetch abort exception.

#### Motivation:
This change makes it easier to debug data abort/prefetch abort exceptions if the user is not connected to the terminal.

#### Test Plan:
- [x] Check that it compiles
- [x] Trigger some data abort exceptions and check that the output is correct
- [x] Trigger some prefetch abort exceptions and check that the output is correct


src/system/unwind.c
```diff
@@ -96,9 +96,18 @@ _Unwind_Ptr __gnu_Unwind_Find_exidx(_Unwind_Ptr pc, int* nrec) {
 	return (_Unwind_Ptr)&__exidx_start;
 }
 
+struct {
```
noam987: The code style for this struct is wack. Could you please give the struct a typedef (such as trace_t)

src/system/unwind.c
```diff
@@ -155,6 +167,21 @@ void report_data_abort(uint32_t _sp) {
 	fprintf(stderr, "\t%p\n", (void*)vrs.core.r[R_PC]);
 	__gnu_Unwind_Backtrace(trace_fn, NULL, &vrs);
```
noam987: It would also make more sense for the trace to not be a global variable. Instead you can create the struct locally here and pass the pointer to the struct instead of NULL here. and then you can use it as d in the trace_fn.

src/system/unwind.c
```diff
@@ -96,9 +96,18 @@ _Unwind_Ptr __gnu_Unwind_Find_exidx(_Unwind_Ptr pc, int* nrec) {
 	return (_Unwind_Ptr)&__exidx_start;
 }
 
+struct {
+	uint32_t[12] pcs;
```
noam987: Wouldn't it make more sense for this to either be a pointer to an int array which is dynamically allocated, or as a linked list so that it can dynamically adjust the size based on the size of the trace?

src/system/unwind.c
```diff
@@ -155,6 +167,21 @@ void report_data_abort(uint32_t _sp) {
 	fprintf(stderr, "\t%p\n", (void*)vrs.core.r[R_PC]);
 	__gnu_Unwind_Backtrace(trace_fn, NULL, &vrs);
 	fputs("END OF TRACE\n", stderr);
+	
+	for(size_t i = 0; i < trace.size / 4; i++) {
+		vexDisplayString(brain_line_no++, "%p %p %p %p", (void*)trace.pcs[4*i], (void*)trace.pcs[4*i+1], (void*)trace.pcs[4*i+2], (void*)trace.pcs[4*i+3]);
```
noam987: This display string should probably have more formatting than just 4 raw pointers in a line. There's currently not even any indication to the user that it's a stack trace.

src/system/unwind.c
```diff
@@ -96,9 +96,18 @@ _Unwind_Ptr __gnu_Unwind_Find_exidx(_Unwind_Ptr pc, int* nrec) {
...
+	size_t size;
+} trace;
+
 _Unwind_Reason_Code trace_fn(_Unwind_Context* unwind_ctx, void* d) {
 	uint32_t pc = _Unwind_GetIP(unwind_ctx);
 	fprintf(stderr, "\t%p\n", (void*)pc);
+	if (trace.size < sizeof(trace.pcs) / sizeof(trace.pcs[0]))
+		trace.pcs[trace.size++] = pc;
+	else
+		; // TODO: handle this
```
noam987: Can you please add curly brackets to the if and else cases. If you want suggestions for how to do the else case please reach out, I'm happy to help brainstorm the desired behavior here. I personally think that trying to get dynamically sized trace will be more useful or if you are to hard code in the cap for the length of the array then you shouldn't even need a else case.

noam987 (dismissed):
Idea behind this PR looks good but there are a few major issues, both stylistically and content wise. I would like the trace to not be global and instead passed into the trace function via __gnu_unwind_Backtrace (source for that can be found here if you want to look at it for your reference: [https://github.com/gcc-mirror/gcc/blob/master/libgcc/unwind-arm-common.inc](https://github.com/gcc-mirror/gcc/blob/master/libgcc/unwind-arm-common.inc)). Feel free to ask follow up questions here or ping me on discord in our beta server or in a help post on VTOW.

src/system/unwind.c
ion098: I wanted to include the word "STACKTRACE: " before the pointers, but just eyeballing it based on pictures of the current data abort display, I wasn't sure if the text would be too wide to fit on the screen.

src/system/unwind.c
ion098: I used a fixed sized array because I was worried about running out of memory, which seems like something that would be bad in this scenario. I'm not sure how likely this is though, if it's highly unlikely then I can change this to use dynamic allocation.

noam987:
Your code also does not compile. Here is a list of the compile errors generated by the pipeline: 
```
src/system/unwind.c:100:10: error: expected identifier or '(' before '[' token
  100 |  uint32_t[12] pcs;
      |          ^
src/system/unwind.c: In function 'trace_fn':
src/system/unwind.c:107:31: error: 'struct <anonymous>' has no member named 'pcs'
  107 |  if (trace.size < sizeof(trace.pcs) / sizeof(trace.pcs[0]))
      |                               ^
src/system/unwind.c:107:51: error: 'struct <anonymous>' has no member named 'pcs'
  107 |  if (trace.size < sizeof(trace.pcs) / sizeof(trace.pcs[0]))
      |                                                   ^
src/system/unwind.c:108:8: error: 'struct <anonymous>' has no member named 'pcs'
  108 |   trace.pcs[trace.size++] = pc;
      |        ^
src/system/unwind.c: In function 'report_data_abort':
src/system/unwind.c:172:64: error: 'struct <anonymous>' has no member named 'pcs'
  172 |   vexDisplayString(brain_line_no++, "%p %p %p %p", (void*)trace.pcs[4*i], (void*)trace.pcs[4*i+1], (void*)trace.pcs[4*i+2], (void*)trace.pcs[4*i+3]);
      |                                                                ^
src/system/unwind.c:172:87: error: 'struct <anonymous>' has no member named 'pcs'
  172 |   vexDisplayString(brain_line_no++, "%p %p %p %p", (void*)trace.pcs[4*i], (void*)trace.pcs[4*i+1], (void*)trace.pcs[4*i+2], (void*)trace.pcs[4*i+3]);
      |                                                                                       ^
src/system/unwind.c:172:112: error: 'struct <anonymous>' has no member named 'pcs'
  172 |   vexDisplayString(brain_line_no++, "%p %p %p %p", (void*)trace.pcs[4*i], (void*)trace.pcs[4*i+1], (void*)trace.pcs[4*i+2], (void*)trace.pcs[4*i+3]);
      |                                                                                                                ^
src/system/unwind.c:172:137: error: 'struct <anonymous>' has no member named 'pcs'
  172 |   vexDisplayString(brain_line_no++, "%p %p %p %p", (void*)trace.pcs[4*i], (void*)trace.pcs[4*i+1], (void*)trace.pcs[4*i+2], (void*)trace.pcs[4*i+3]);
      |                                                                                                                                         ^
src/system/unwind.c:176:62: error: 'struct <anonymous>' has no member named 'pcs'
  176 |    vexDisplayString(brain_line_no++, "%p %p %p", (void*)trace.pcs[trace.size-2], (void*)trace.pcs[trace.size-1], (void*)trace.pcs[trace.size]);
      |                                                              ^
src/system/unwind.c:176:94: error: 'struct <anonymous>' has no member named 'pcs'
  176 |    vexDisplayString(brain_line_no++, "%p %p %p", (void*)trace.pcs[trace.size-2], (void*)trace.pcs[trace.size-1], (void*)trace.pcs[trace.size]);
      |                                                                                              ^
src/system/unwind.c:176:126: error: 'struct <anonymous>' has no member named 'pcs'
  176 |    vexDisplayString(brain_line_no++, "%p %p %p", (void*)trace.pcs[trace.size-2], (void*)trace.pcs[trace.size-1], (void*)trace.pcs[trace.size]);
      |                                                                                                                              ^
src/system/unwind.c:179:59: error: 'struct <anonymous>' has no member named 'pcs'
  179 |    vexDisplayString(brain_line_no++, "%p %p", (void*)trace.pcs[trace.size-1], (void*)trace.pcs[trace.size]);
      |                                                           ^
src/system/unwind.c:179:91: error: 'struct <anonymous>' has no member named 'pcs'
  179 |    vexDisplayString(brain_line_no++, "%p %p", (void*)trace.pcs[trace.size-1], (void*)trace.pcs[trace.size]);
      |                                                                                           ^
src/system/unwind.c:182:56: error: 'struct <anonymous>' has no member named 'pcs'
  182 |    vexDisplayString(brain_line_no++, "%p", (void*)trace.pcs[trace.size]);
      |                                                        ^
make: *** [common.mk:273: bin/system/unwind.c.o] Error 1

```

src/system/unwind.c
ion098: Also, is dynamic memory allocation still guaranteed to work after a data abort exception?

src/system/unwind.c
ion098: Implemented this

src/system/unwind.c
ion098: Added curly braces, also added an else case

src/system/unwind.c
ion098: After discussing on vtow I'm keeping this as a fixed sized array for now, I increased the array size to 20.

src/system/unwind.c
ion098: Added "TRACE: " before the pointers to indicate that its a stacktrace.

ion098:
@noam987 Do you think that the registers and the heap used and stack remaining should also be printed to the brain screen? Is there enough space to print those? If not, should they be written to a log file on the SD card, if present?

ion098:
Also, does the stacktrace contain the PC? If so, then that should save a line on the brain screen.

noam987:
Looks much better. Waiting to test this in the lab before I approve it

src/system/rtos_hooks.c
```diff
@@ -29,11 +29,15 @@ void FIQInterrupt() {
 // Replacement for DataAbortInterrupt
 void DataAbortInterrupt() {
 	taskDISABLE_INTERRUPTS();
+	taskENTER_CRITICAL();
 
 	register int sp;
 	asm("add %0,sp,#8\n" : "=r"(sp));
-	extern void report_data_abort(uint32_t);
-	report_data_abort(sp);
+	for(uint8_t i = 1; i <= 21; i++) {
```
AndrewLuGit, line 36: This for loop should be moved to after the call to `report_fatal_error`. Currently, it messes up the program counter when printing the stack trace.

AndrewLuGit (changes requested):

src/system/rtos_hooks.c
```diff
@@ -29,11 +29,15 @@ void FIQInterrupt() {
 // Replacement for DataAbortInterrupt
 void DataAbortInterrupt() {
 	taskDISABLE_INTERRUPTS();
+	taskENTER_CRITICAL();
```
AndrewLuGit: Please remove this, using `taskENTER_CRITICAL()` here prevents the data abort from printing to serial for mysterious reasons. Also remove it from `PrefetchAbortInterrupt()`.

src/system/rtos_hooks.c
ion098: Fixed in https://github.com/purduesigbots/pros/pull/651/commits/b8f302634494d455f5c0c74918423112030dcb65.

src/system/rtos_hooks.c
ion098: Fixed in https://github.com/purduesigbots/pros/pull/651/commits/b8f302634494d455f5c0c74918423112030dcb65.

src/system/unwind.c
```diff
@@ -132,34 +146,76 @@ void p2vrs_from_data_abort(_uw* sp, struct phase2_vrs* vrs) {
...
+	fputs("REGISTERS AT ERROR\n", stderr);
 	print_phase2_vrs(&vrs);
 
 	fputs("BEGIN STACK TRACE\n", stderr);
 	fprintf(stderr, "\t%p\n", (void*)vrs.core.r[R_PC]);
-	__gnu_Unwind_Backtrace(trace_fn, NULL, &vrs);
+	struct trace_t trace = {{0}, 0, false};
+	__gnu_Unwind_Backtrace(trace_fn, &trace, &vrs);
 	fputs("END OF TRACE\n", stderr);
+	/*
```
AndrewLuGit: Everything works now, just get rid of this commented out code

ion098: Done in https://github.com/purduesigbots/pros/pull/651/commits/c7d518be077a9387974f9a99ab316baf8d1086cd.

AndrewLuGit (approved):
LGTM

PR #652, opened 2024-04-26
Pros4/feature/get devices vector

aberiggs:
#### Summary:

- Gave critical device classes the get_all_devices() static class functions which returns a vector of that class' objects (e.g. Motor, Imu, etc.)
- Removed equals operator for motor movement assignment
- Removed explicit from device constructors
- Created static version of get_plugged_type() that takes in a port number

#### Motivation:

- The function get_all_devices() can be used in the automatic setup of devices for users.
- get_plugged_type() having a static overload allows for the checking of device types before creating devices

#### Test Plan:
Did lab testing for the new functions (but didn't get a chance to check for all devices)

### DIFF
--- a/src/devices/vdml_device.cpp
+++ b/src/devices/vdml_device.cpp
@@ -42,6 +42,48 @@ pros::DeviceType Device::get_plugged_type() const {
 	return_port(_port - 1, type);
 }
 
+template <typename T>
+std::vector<T> Device::get_all_devices() {
+	DeviceType device_type {};
+
+	if(std::is_same<T,Motor>::value)
+		device_type = DeviceType::motor;
+	else if (std::is_same<T,Rotation>::value)
+		device_type = DeviceType::motor;
+	else if (std::is_same<T,IMU>::value)
+		device_type = DeviceType::imu;
+	else if (std::is_same<T,Distance>::value)
+		device_type = DeviceType::distance;
+	else if (std::is_same<T,Vision>::value)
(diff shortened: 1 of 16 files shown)


src/devices/vdml_device.cpp
```diff
@@ -28,18 +28,40 @@ bool Device::is_installed() {
...
+	DeviceType type = (DeviceType) pros::c::registry_get_plugged_type(port - 1);
 	
-	return_port(_port - 1, type);
+	return_port(port - 1, type);
+}
+
+std::vector<Device> Device::get_all_devices(pros::DeviceType device_type) {
+	std::vector<Device> device_list {};
+
+	for (std::uint8_t curr_port = 0; curr_port < 21; ++curr_port) {
```
noam987, line 52: Why not have curr_port be std::int8_t and then remove the cast on line 60?

noam987 (approved):
Main Comments I have is to make sure we get docs done for these functions before full release of PROS 4. Based on the results for testing yesterday I feel good to release this into testing.

src/devices/vdml_device.cpp
aberiggs: The cast is to a uint8_t on line 60 because of type promotion to an int when I did a uint8_t value + 1. The Device constructor expects a uint8_t.

noam987: Ok that makes sense then

PR #653, opened 2024-05-02
Adds static getters for some devices

noam987:
#### Summary:

Creates a get_imu, get_gps, and get_vision functions so users don't need to know their ports at compile time

#### Motivation:

Makes it possible for users to make programs that aren't reliant on their IMU ports so that if a port dies, no code needs to be changed, just plug into a new port

#### Test Plan:

- [ ] Compiles
- [ ] 1 IMU plugged in always returns a device for that port
- [ ] 2 IMUs plugged in returns a device for the lower port number when called odd times, and higher even times.
- [ ] Does not crash when trying to get a device that isn't on the robot
- [ ] Other devices behave similarly


phinc23 (approved):
LGTM

PR #656, opened 2024-05-04
🐛Remove = operator overload from motor groups

noam987:
#### Summary:

Removes the = overload from motor groups as motors don't have them either
#### Test Plan:

- [x] compiles
- [x] Doesn't compile if you use the = overload


phinc23 (approved):
Looks good

PR #677, opened 2024-06-07
✨ Add RecursiveMutex class & make Mutex/RecursiveMutex ctors constexpr

ion098:
#### Summary:
Adds a `RecursiveMutex` class, and make the constructors for Mutex/RecursiveMutex `constexpr`.

#### Motivation:
There is currently no C++ API for recursive mutexes, which prevents using `std::lock_guard` with recursive mutexes. Making the constructors `constexpr` allows mutex variables to be declared as `constinit`.

#### Test Plan:

- [x] Check that it compiles.
- [x] Check that the mutex only allows one thread to lock it.
- [x] Check that the same thread can lock the `RecursiveMutex` multiple times.

##### Test Code:

```cpp
constinit pros::RecursiveMutex rec_mut{};
constinit pros::Mutex mut{};

void thread_1() {
{
    std::lock_guard lock{rec_mut};
    assert(rec_mut.try_lock());
    rec_mut.unlock();
    for (int j = 0; j != 10; j++) {
        printf("Thread 1: j is %i", j);
}
}
{
    std::lock_guard lock{mut};
    assert(!mut.try_lock());
    for (int j = 0; j != 10; j++) {
        printf("Thread 1: j is %i", j);
    }
}
}

void thread_2() {
{
    std::lock_guard lock{rec_mut};
    assert(rec_mut.try_lock());
    rec_mut.unlock();
    for (int j = 0; j != 10; j++) {
        printf("Thread 2: j is %i", j);
        pros::delay(10);
}
}
{
    std::lock_guard lock{mut};
    assert(!mut.try_lock());
    for (int j = 0; j != 10; j++) {
        printf("Thread 2: j is %i", j);
        pros::delay(10);
    }
}
}
```


AndrewLuGit:
@ion098 there's issues with including `rtos/task.h` and `pros/apix.h` in `pros/rtos.hpp` that result in build errors. For `rtos/task.h`, the issue is that the task functions (e.g. `task_notify`) are defined in both `rtos/task.h` and `pros/rtos.h`. For `pros/apix.h`, the issue is there is a circular dependency: `pros/apix.h` includes `pros/serial.hpp`, which includes `pros/device.hpp`, but `pros/device.hpp` includes `pros/rtos.hpp` (which now includes `pros/apix.h`).
To fix the circular include, I think the best option might be to just move the `mutex_recursive` functions from `pros/apix.h` to `pros/rtos.h`.

include/pros/rtos.hpp
```diff
@@ -22,8 +22,10 @@
 #ifndef _PROS_RTOS_HPP_
 #define _PROS_RTOS_HPP_
 
+#include "rtos/task.h"
```
AndrewLuGit: This cannot be included in `pros/rtos.hpp`, since the rtos headers aren't added to the kernel template, so this will result in an include file not found error.

src/rtos/rtos.cpp
```diff
@@ -14,9 +14,12 @@
 
 #include "pros/rtos.hpp"
 
+#include <atomic>
 #include <system_error>
 
 #include "kapi.h"
+#include "pros/rtos.h"
```
AndrewLuGit: Replace this with `#include rtos/task.h`

AndrewLuGit (changes requested):
Tested and the recursive mutex works, just change the includes as suggested. There is still some redefinitions when compiling `rtos.cpp`, but it's ok for now (I'll fix it in a later PR)

src/rtos/rtos.cpp
ion098: Addressed in https://github.com/purduesigbots/pros/pull/677/commits/26f77c86587f6ac3e6fb0991f71622199ac848f1.

include/pros/rtos.hpp
ion098: Addressed in https://github.com/purduesigbots/pros/pull/677/commits/26f77c86587f6ac3e6fb0991f71622199ac848f1.

AndrewLuGit (approved):
:+1:

PR #683, opened 2024-06-28
fix: :bug: Increase max text length for controller text setting

### LINKED ISSUE #679: 🐛[BUG] - Controller text cuts off too early
### Describe the bug
When printing to the controller, the text cuts off earlier than it should.  

### To Reproduce
In a new project, run this code twice, switching which `master.set_test()` is ran.  
```cpp
void opcontrol() {
  pros::Controller master(pros::E_CONTROLLER_MASTER);

  while (true) {
    // master.set_text(2, 0, "0.0          0.0");  // both of these print to controller, but the second one isn't all the way to the right 
    master.set_text(2, 0, "0.0             0.0");  // with an additional 3 spaces, only the first 0.0 will print to controller
    
    pros::delay(50);
  }
}
```

### Expected Behavior
The text cutoff should be at the edge of the screen instead of somewhere in the middle.  

### Screenshots
The cutoff being early can be seen in the image below.  
image

### Desktop (please complete the following information):
 - Windows 10
 - Kernel 4.1.0
 - lvgl 8.3.8

ion098:
#### Summary:
Changes max length of controller text to 31 characters.

#### Motivation:
VEXCode allows users to set controller text with a maximum length of 31 characters, but PROS caps controller text at 15 characters.

##### References (optional):
Close #679

#### Test Plan:
- [x] Check it compiles
- [x] Check that controller text setting works with strings up to 31 characters long
- [x] Check that this doesn't break controller text clearing

### DIFF
--- a/src/devices/controller.c
+++ b/src/devices/controller.c
@@ -19,6 +19,7 @@
 #include "vdml/vdml.h"
 
 #define CONTROLLER_MAX_COLS 15
+#define CONTROLLER_MAX_CHARS 31
 
 // From enum in misc.h
 #define NUM_BUTTONS 12
@@ -107,7 +108,7 @@ int32_t controller_set_text(controller_id_e_t id, uint8_t line, uint8_t col, con
 	else
 		col++;
 
-	char* buf = strndup(str, CONTROLLER_MAX_COLS + 1);
+	char* buf = strndup(str, CONTROLLER_MAX_CHARS + 1);
 
 	uint32_t rtn_val = vexControllerTextSet(id, line, col, buf);
 	free(buf);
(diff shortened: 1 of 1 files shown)


AndrewLuGit (dismissed):
LGTM

liucougar:
this would be nice to have

src/devices/controller.c
```diff
@@ -19,6 +19,7 @@
 #include "vdml/vdml.h"
 
 #define CONTROLLER_MAX_COLS 15
+#define CONTROLLER_MAX_CHARS 31
```
WillXuCodes: A couple of nits: 
- Might as well get rid of CONTROLLER_MAX_COLS
- Parenthesize and unsign the define

Other than that this looks good, thanks for the contribution :) 

```suggestion
#define CONTROLLER_MAX_CHARS ( 31U )
```

WillXuCodes (changes requested):
Slight change requested

src/devices/controller.c
ion098: Addressed in https://github.com/purduesigbots/pros/pull/683/commits/f907d62932d81042b081eac8641ee435cb3e5e18. (`CONTROLLER_MAX_COLS` still needs to be there since the controller only supports 20 columns, but can print 31 characters, due to the font being proportional).

src/devices/controller.c
```diff
@@ -18,7 +18,8 @@
 #include "v5_api.h"
 #include "vdml/vdml.h"
 
-#define CONTROLLER_MAX_COLS 15
+#define CONTROLLER_MAX_COLS ( 20U )
```
AndrewLuGit, line 21: Why was this changed from 15 to 20?

ion098: Because VEXCode supports 20 columns:
image

AndrewLuGit (approved):

PR #685, opened 2024-07-05
feat: :sparkles: Make controller ctor constexpr

ion098:
#### Summary:
Changes the `pros::Controller` constructor to be `constexpr`.

#### Motivation:
Allows `pros::Controller` instances to be declared `constinit`/`constexpr`.

#### Test Plan:

- [x] Check it compiles
- [ ] Check that `Controller` instances can be declared as `constinit`/`constexpr`.

### DIFF
--- a/include/pros/misc.hpp
+++ b/include/pros/misc.hpp
@@ -45,7 +45,7 @@ class Controller {
 	 * 			  The ID of the controller (e.g. the master or partner controller).
 	 * 			  Must be one of CONTROLLER_MASTER or CONTROLLER_PARTNER
 	 */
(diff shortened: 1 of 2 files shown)


djava:
While you're doing this, what other ctors should this be applied to? I imagine this isn't the only one.

ion098:
> While you're doing this, what other ctors should this be applied to? I imagine this isn't the only one.

The issue is that some device ctors can't be made constexpr becuase they touch global state (changing this would require a breaking change to the PROS API which I don't think is gonna happen now), it feels a bit weird to have some devices with constexpr ctors and others without.

PR #686, opened 2024-07-05
:sparkles: Add rotation constructor with reversed flag

ion098:
#### Summary:
Add a rotation sensor constructor that takes a reversed flag.

#### Motivation:
Some people prefer using a reversed flag instead of negative port numbers.

#### Test Plan:

- [ ] Check it compiles
- [ ] Check that both constructors work as expected

### DIFF
--- a/include/pros/rotation.hpp
+++ b/include/pros/rotation.hpp
@@ -55,6 +55,27 @@ class Rotation : public Device {
 	 */
 	Rotation(const std::int8_t port);
 
+	/**
+	 * Constructs a new Rotation Sensor object
+	 *
+	 * ENXIO - The given value is not within the range of V5 ports |1-21|.
+	 * ENODEV - The port cannot be configured as a Rotation Sensor
+	 *
+	 * \param port
+	 *        The V5 port number from 1 to 21.
+	 * \param reversed
+	 *        Whether or not the rotation sensor is reversed.
+	 *
+	 * 	\b Example
+	 * \code
+	 * void opcontrol() {
+	 * 	 pros::Rotation rotation_sensor(1, false); //Creates a Rotation Sensor on port 1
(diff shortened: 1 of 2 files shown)


src/devices/vdml_rotation.cpp
```diff
@@ -16,12 +16,10 @@
...
 
-Rotation::Rotation(const std::int8_t port) : Device(abs(port), DeviceType::rotation) {
-	if (port < 0) {
-		pros::c::rotation_set_reversed(abs(port), true);
-	} else {
-		pros::c::rotation_set_reversed(port, false);
-	}
+Rotation::Rotation(const std::int8_t port) : Rotation(abs(port), port < 0) {}
+
+Rotation::Rotation(const std::uint8_t port, const bool reversed): Device(port, DeviceType::rotation) {
```
BennyBot, line 21:
```suggestion
Rotation::Rotation(const std::int8_t port, const bool reversed): Device(abs(port), DeviceType::rotation) {
```
in the case the user provides a negative number and specifies the reverse flag, the reverse flag should be used.

include/pros/rotation.hpp
```diff
@@ -55,6 +55,27 @@ class Rotation : public Device {
...
+	 *
+	 * 	\b Example
+	 * \code
+	 * void opcontrol() {
+	 * 	 pros::Rotation rotation_sensor(1, false); //Creates a Rotation Sensor on port 1
+	 *   pros::Rotation reversed_rotation_sensor(2, true); //Creates a reversed Rotation Sensor on port 2
+	 * }
+	 * \endcode
+	 */
+	Rotation(const std::uint8_t port, const bool reversed);
```
BennyBot, line 77:
```suggestion
	Rotation(const std::int8_t port, const bool reversed);
```

BennyBot (changes requested):

PR #689, opened 2024-07-16
feat: :technologist: Check device literal port numbers at compile time

ion098:
#### Summary:
Checks port numbers for device literals at compile time rather than at runtime. (Thanks to @djava for giving me suggestions on refactoring my mess of templates into actually readable code)

#### Motivation:
If something can be checked at compile time, it should be checked at compile time.

#### Test Plan:
- [x] check it compiles
- [ ] check that port literals work as expected
- [ ] check that out of range port numbers give a compile time error that is readable

#### Note:
This is an ABI breaking change and therefore should be a minor version bump.

### DIFF
--- a/include/pros/detail.hpp
+++ b/include/pros/detail.hpp
@@ -13,6 +13,8 @@
  * file You can obtain one at http://mozilla.org/MPL/2.0/.
  *
  */
+#ifndef _PROS_DETAIL_HPP_
+#define _PROS_DETAIL_HPP_
 
 #include <array>
 #include <charconv>
@@ -22,30 +24,9 @@ namespace pros::detail {
 template <char ...Chars>
 consteval uint8_t parse_port() {
     const std::array chars{Chars...};
-    const char* str = chars.begin();
-    int base = 10;
     uint8_t port_num{0};
-    if (chars.size() > 2 && chars[0] == '0') {
-        str += 2;
-        // from_chars doesn't seem to handle hex/binary/octal prefixes,
-        // so we handle them ourselves
-        switch (chars[1]) {
-            case 'X': case 'x':
-                // hex literal
-                base = 16;
-                break;
-            case 'B': case 'b':
-                // binary literal
-                base = 2;
-                break;
-            default:
-                // octal literal
-                str -= 1;
-                base = 8;
-        }
-    }
-    auto result = std::from_chars(str, chars.end(), port_num, base);
-    if (result.ptr != chars.end()) {
+    auto result = std::from_chars(chars.begin(), chars.end(), port_num, 0);
+    if (result.ptr != chars.end() || result.ec != std::errc{}) {
         return 0;
     }
     return port_num;
@@ -56,4 +37,23 @@ consteval bool is_valid_port() {
     auto port_num = parse_port<Chars...>();
     return port_num >= 1 && port_num <= 21;
 }
+
+// Tests:
+static_assert(is_valid_port<'1'>() == true, "Kernel test failed");
+static_assert(is_valid_port<'0'>() == false, "Kernel test failed");
+// 0x16 = 22
+static_assert(is_valid_port<'0','x','1','6'>() == false, "Kernel test failed");
+// 0x15 = 21
+static_assert(is_valid_port<'0','x','1','5'>() == true, "Kernel test failed");
+// 025 = 21
+static_assert(is_valid_port<'0','2','5'>() == true, "Kernel test failed");
+// 0b1 = 1
+static_assert(is_valid_port<'0','b','1'>() == true, "Kernel test failed");
+// 0x1 = 1
+static_assert(is_valid_port<'0','x','1'>() == true, "Kernel test failed");
+static_assert(is_valid_port<'-','1'>() == false, "Kernel test failed");
+static_assert(is_valid_port<'2','2'>() == false, "Kernel test failed");
+
 }
+
+#endif // _PROS_DETAIL_HPP_
--- a/include/pros/detail.hpp
+++ b/include/pros/detail.hpp
@@ -1,0 +1,59 @@
+/**
+ * \file pros/detail.hpp
+ *
+ * Contains PROS internal functions used in port literal functions.
+ *
+ * This file should not be modified by users, since it gets replaced whenever
+ * a kernel upgrade occurs.
+ *
+ * Copyright (c) 2017-2023 Purdue University ACM SIGBots.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License v. 2.0. If a copy of the MPL was not distributed with this
+ * file You can obtain one at http://mozilla.org/MPL/2.0/.
+ *
+ */
+
+#include <array>
+#include <charconv>
+#include <cstdint>
+
+namespace pros::detail {
+template <char ...Chars>
+consteval uint8_t parse_port() {
+    const std::array chars{Chars...};
+    const char* str = chars.begin();
+    int base = 10;
+    uint8_t port_num{0};
+    if (chars.size() > 2 && chars[0] == '0') {
+        str += 2;
+        // from_chars doesn't seem to handle hex/binary/octal prefixes,
+        // so we handle them ourselves
+        switch (chars[1]) {
+            case 'X': case 'x':
+                // hex literal
+                base = 16;
+                break;
+            case 'B': case 'b':
+                // binary literal
+                base = 2;
+                break;
+            default:
+                // octal literal
+                str -= 1;
+                base = 8;
+        }
+    }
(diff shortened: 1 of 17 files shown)


djava:
Please fix this MR to only include relevant commits so its reviewable.

include/pros/detail.hpp
```diff
@@ -0,0 +1,59 @@
...
+template <char ...Chars>
+consteval uint8_t parse_port() {
+    const std::array chars{Chars...};
+    const char* str = chars.begin();
+    int base = 10;
+    uint8_t port_num{0};
+    if (chars.size() > 2 && chars[0] == '0') {
+        str += 2;
+        // from_chars doesn't seem to handle hex/binary/octal prefixes,
+        // so we handle them ourselves
```
djava, line 33: This sucks lmao, good catch though

include/pros/detail.hpp
```diff
@@ -0,0 +1,59 @@
...
+        switch (chars[1]) {
+            case 'X': case 'x':
+                // hex literal
+                base = 16;
```
djava, line 37: Does `std::from_chars` handle both uppercase and lowercase hex literals?

include/pros/detail.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <charconv>
+#include <cstdint>
+
+namespace pros::detail {
+template <char ...Chars>
+consteval uint8_t parse_port() {
+    const std::array chars{Chars...};
+    const char* str = chars.begin();
+    int base = 10;
+    uint8_t port_num{0};
```
djava, line 29: i'd move this declaration to right above the `std::from_chars` call because the type of this argument is actually vital to what the function does (continually blown away by how much the `std::from_chars` API sucks!)

include/pros/detail.hpp
```diff
@@ -0,0 +1,59 @@
...
+            default:
+                // octal literal
+                str -= 1;
+                base = 8;
+        }
+    }
+    auto result = std::from_chars(str, chars.end(), port_num, base);
+    if (result.ptr != chars.end()) {
+        return 0;
+    }
```
djava: Not sure if this is enough error checking on stupid from_chars - the idiom i've seen elsewhere is `result.ec != std::errc{}`. Some research (https://stackoverflow.com/questions/72831184/bulletproof-use-of-from-chars) says you should be checking both?  (WOW FROM_CHARS API SUCKS!!!!)

include/pros/distance.hpp
```diff
@@ -239,7 +240,12 @@ namespace literals {
  * }
  * \endcode
  */
-const pros::Distance operator"" _dist(const unsigned long long int d);
+template <char... Cs>
+const pros::Distance operator"" _dist() {
+	static_assert(pros::detail::is_valid_port<Cs...>(), "Port is out of range!");
```
djava: Was originally unsure if this is better than requires-clause, but it probably is because it allows giving a message (and benefits of requires-clause, ie SFINAE, arent really useful here).

The message you have is not great though, because a failed `from_chars` will give the same error. I imagine that this happens if you try to give it a non-integer? I would probably prefer "invalid port number" or something

djava: Same goes for other declarations

include/pros/motors.hpp
```diff
@@ -2419,7 +2426,12 @@ const pros::Motor operator"" _mtr(const unsigned long long int m);
  * }
  * \endcode
  */
-const pros::Motor operator"" _rmtr(const unsigned long long int m);
+template <char... Cs>
+const pros::Motor operator"" _rmtr() {
```
djava, line 2430: I wonder if theres a single pros project in the whole world that uses `_rmtr`

include/pros/detail.hpp
```diff
@@ -0,0 +1,59 @@
...
+    return port_num;
+}
+
+template <char ...Chars>
+consteval bool is_valid_port() {
+    auto port_num = parse_port<Chars...>();
+    return port_num >= 1 && port_num <= 21;
+}
```
djava, line 60: Add some test-cases in static_asserts?

include/pros/rotation.hpp
```diff
@@ -369,6 +370,23 @@ class Rotation : public Device {
...
+	 * \code
+	 * using namespace pros::literals;
+	 *
+	 * void opcontrol() {
+	 * 	pros::Rotation rotation_sensor = -1_rot;
+	 *  printf("Reversed: %d \n", rotation_sensor.get_reversed());
+	 * }
+	 * \endcode
+	 */
+	Rotation operator-() && { return Rotation(-this->get_port()); }
```
djava, line 388: why should this exist? can you not create negatives with the other user-defined literal operator? If you can, then this probably shouldnt exist

include/pros/detail.hpp
djava, line 33: Do we need to handle literals of the form `-0xABCD`?

djava: this is supported for normal integer literals so im thinkin probably yes.

include/pros/rotation.hpp
djava: Oh I see that its rvalue-this-qualified. still a little weird

ion098: No, not for rotation sensors. Both `10_rrot` and `10_rotr` seem a bit confusing compared to `-10_rot` (negative port numbers also match how the ctors work)

djava: If this is necessary for rotation, its surely necessary for motors too?

djava: hopefully its not necessary for either though lol

include/pros/detail.hpp
ion098: Well a negative number isn't meaningful for port numbers anyway, so I don't think that needs special handling

include/pros/detail.hpp
ion098: Done in https://github.com/purduesigbots/pros/pull/689/commits/bea9c443195fbecb738a4154f2048603fa4dd178.

include/pros/detail.hpp
ion098: Done in https://github.com/purduesigbots/pros/pull/689/commits/bea9c443195fbecb738a4154f2048603fa4dd178.

include/pros/motors.hpp
tropicaaal: found one https://github.com/EastRobotics/2616C/blob/bc6b76e1a5fe70ac314ab24600e2ef1da815df4a/src/opcontrol.cpp#L21

include/pros/detail.hpp
ion098: Turns out it does, I failed to read the docs lol. Fixed in https://github.com/purduesigbots/pros/pull/689/commits/bea9c443195fbecb738a4154f2048603fa4dd178.

djava: oh thank god

include/pros/rotation.hpp
djava: im just confused why this doesnt work with the user-defined literal directly.

ion098: Because `-10_rot` is `-(operator""_rot<'1','0'>())`, not `operator""_rot<'-','1','0'>()`

ion098: Motors have _rmtr so its not needed for motors, although adding this to motors might be a good idea for consistency

djava: oh i see. yeah that sucks lmao. i think this should be added to motors as well if it has to exist.

ion098:
@Rocky14683 can you rerun the pipeline please?

include/pros/detail.hpp
ion098: Turns out I failed to read the docs doubly lol: from_chars does indeed NOT auto detect base womp womp

djava: lol? so what, we need to add back in the `0b` and `0x` detection? that sucks bruh

djava: > Well a negative number isn't meaningful for port numbers anyway, so I don't think that needs special handling

This isn't really true, it at least sort of means "This port, with the rotation of the corresponding device reversed if there is any". However, seems like user-defined literals aren't able to deal with negatives (wtf bruh) so whatever.

ion098:
@Rocky14683 It should compile now, can you rerun the pipeline please?

include/pros/detail.hpp
ion098: Yeah that's true, but it doesn't matter anyway:

> `-10_rot` is `-(operator""_rot<'1','0'>())`, not `operator""_rot<'-','1','0'>()`

PR #697, opened 2024-09-19
Added optical set/get integration time

Gracelu128:
#### Summary:
Added optical set/get integration time

#### Motivation:
Requests to add this to pros 4 as well(in pros 3).

##### References (optional):
Refer to #481 

#### Test Plan:
- [ ] Test valid, invalid value for set integration
- [ ] Get integration before setting it to value
- [ ] Get integration after setting value


Rocky14683 (approved):
Tested, this works.

PR #699, opened 2024-09-22
:sparkles: Add set/get integration time methods to optical sensor

SprintKeyz:
#### Summary:
Add the ability to get and set the integration time of the optical sensor in PROS 4

#### Motivation:
Before this change, PROS 4 did not have a way to get or set the integration time of the optical sensor.

#### Test Plan:
- [x] Get optical sensor integration time and confirm it matches default
- [ ] Set optical integration time to a new value
- [ ] Get optical sensor integration time and confirm it matches new value

### DIFF
--- a/include/pros/optical.h
+++ b/include/pros/optical.h
@@ -460,6 +460,39 @@ int32_t optical_enable_gesture(uint8_t port);
  */
 int32_t optical_disable_gesture(uint8_t port);
 
+/**
+ * Get integration time (update rate) of the optical sensor in milliseconds, with
+ * minimum time being 
+ *
+ * This function uses the following values of errno when an error state is
+ * reached:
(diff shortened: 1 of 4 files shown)


SprintKeyz:
The reason the other two boxes are not checked here is because in my testing, the VEX SDK will always return the same value of 103 when calling `vexDeviceOpticalIntegrationTimeGet` regardless of what the user sets it to. I am therefore unable to confirm whether this value is correctly updated, but the code in this PR should not be causing this issue (nor is this an issue the PROS team would be able to fix). 

The solution is likely to do some more testing to confirm this can be reproduced and then make a vex forum post and hope this gets solved.

liam-teale:
Can confirm, but my results were a bit stranger. Originally I was able to set it but now I can't

PR #703, opened 2024-10-07
⚡Use `-mfloat-abi=hard`, Update `libm.a` and `libc.a`

liam-teale:
### Summary:
- Update `libm.a` and `libc.a` to the optimized versions found in [arm-none-eabi v13.3Rel1](https://developer.arm.com/downloads/-/arm-gnu-toolchain-downloads)
- Switch `-mfloat-abi` from `softfp` to `hard`

### Motivation:
Teams are starting to use more advanced algorithms for guidance, navigation, and control. Some of these algorithms are computationally expensive, such as a particle filter, path planning with object avoidance, and parsing LiDAR data. Having additional CPU performance overhead will enable teams to experiment more freely and be at lower risk of experiencing strange behavior due to the scheduler not having enough time to execute tasks.

The `libm.a` used currently can be replaced with a more up-to-date version of libm, which offers a slight performance increase of approximately **500%-1000%**.
Additionally, `libm.a` is 169 KB smaller, and `libc.a` is 217 KB smaller.

### Benchmarks:
[Benchmark Methodology](https://github.com/SizzinSeal/libm-benchmark-hard/blob/main/src/main.cpp)
[Google Sheets](https://docs.google.com/spreadsheets/d/1kipf_qSo_dR5Y96VMBYFXJpHGfX1XmrR7F6MwnM8IFw/edit?gid=0#gid=0)
softfp vs hard vs soft

### Implementation details
PROS would normally fail to link with libv5rts if it was compiled with `-mfloat-abi=hard` due to ABI incompatibilities. However, by marking all functions in libv5rts with `__attribute__((pcs("aapcs"))`, the compiler automatically handles these differences. libv5rts also needs to be stripped of `.ARM.attributes` so the linker does not throw an error due to handled ABI incompatibilities.

A python script automatically patches libv5rts. The implementation of this script does not reveal any information about libv5rts, and robust error handling prevents information being leaked when an exception is thrown. This script has been fully integrated with the kernel makefile. The SDK will automatically be patched when needed, and is removed when the `clean` target is run. The only change to `common.mk` was making the `clean` target use a double colon so we could implement the patched sdk removal functionality without adding additional logic to common.mk.

### libm and libc
libm.a and libc.a are the hard versions bundled with arm-none-eabi v.13.3 rel1, the latest version at the time this PR was opened. The specific version used: `arm-none-eabi-v13-3-rel1/lib/thumb/v7-a+simd/hard/`. The version of newlib is v4.4.0

### Toolchain version
This PR does not compile without the latest version of arm-none-eabi-gcc, as breaking changes were introduced to newlib. This means that the arm-none-eabi distributed with the PROS-CLI needs to be updated to v13.3rel1 before this PR can make it into a new release.

Do note that PROS user projects and templates for older versions of PROS will still compile and link properly with a new version of the toolchain.

### Breaking Changes
Any template compiled for PROS 4.1.0 (or older), will need to be recompiled to link properly. Even though no source code has changed, the break in ABI compatibility does necessitate a major version increment of PROS.

However, PROS has not made major version increments for breaking ABI in the past. Granted, templates were not in use at the same scale as today, but given the limited number of templates in use today it is feasible to update them all rapidly. @SizzinSeal has volunteered to open PRs to any and all PROS 4 templates to support a new version of PROS.

### Test Plan:
- [x] compiles on Windows
- [x] compiles on MacOS
- [x] compiles on Linux
- [x] the SDK patching script works
- [x] the SDK patching script does not reveal information about the SDK upon an error
- [x] the patched SDK is deleted by the clean target
- [x] the build script does not interact with the SDK at all when not trying to compile PROS 
- [x] user code compiles
- [x] functions that interact with SDK functions which return or take floats as parameters still work
- [x] libm functions that return or take floats as parameters still work
- [x] libc functions that return or take floats as parameters still work

### Credits
 - [Vexide team](https://github.com/vexide) for the idea
 - @Rocky14683 for figuring out the right version of libm and libc to use, as well as various changes to the makefile and common.mk to get gcc to play nice
 - @ion098 for creating the python script that patches libv5rts
 - @SizzinSeal for cleaning up the python script and integrating it with make, as well as benchmarks
 - @AndrewLuGit for feedback and support
 - @cgjeffries for help with decompiling and keeping me from going insane
 - @Tropix126 for help with decompiling and tricking me into making this PR

### tl;dr
clark


djava:
The vexide blog post states that they switched their platform target to use `eabihf` to attain these improvements. Would pros need this as well? This PR doesn't include that, and I imagine it would need to be reserved for a minor kernel version upgrade due to the ABI break.

djava:
Also, please repeat the benchmarks with pros for validation. Vexide's benchmarks are not sufficient to justify this, as their platform isn't even based on PROS.

liam-teale:
> The vexide blog post states that they switched their platform target to use `eabihf` to attain these improvements. Would pros need this as well?

Vexide uses `eabihf` for their latest release, but they used `softfp` in development versions. They include the results of a benchmark done on PROS. The data is as follows:

| current PROS libm.a | newlib libm.a +softfp | newlib libm.a +hard |
| -------------------- | ---------------------- | -------------------- |
| 383.284 ms             | 3.699ms                     | 1.659ms                 |

As you can see, the performance improvement matches that which I've listed in the description of the PR. See the `VFP registers` section of this PR for additional information

liam-teale:
> Also, please repeat the benchmarks with pros for validation. Vexide's benchmarks are not sufficient to justify this, as their platform isn't even based on PROS.

This is planned. I can't compile PROS, seeing as libv5rts.a is only distributed to vendors. I've opened this PR so that the PROS team can build this PR, and then I can perform benchmarks.

AndrewLuGit:
The kernel will not compile with the new `libc` due to changes in newlib. I was planning to update `libc` and `libm` alongside updating the toolchain.

liam-teale:
That's odd. It compiled fine with the vexcode sdk (obviously can't test it with the sdk but its useful to see if it will compile). Can you share the error messages?

liam-teale:
it seems to have compiled without any problems on azure

AndrewLuGit:
Interesting. On my computer I get this:
```
/home/alu/.config/Code/User/globalStorage/sigbots.pros/install/pros-toolchain-linux/bin/../lib/gcc/arm-none-eabi/10.3.1/../../../../arm-none-eabi/bin/ld: ./bin/libpros.a(vfs.c.o): in function `vfs_initialize':
/home/alu/Documents/VEX/pros/src/system/dev/vfs.c:60: undefined reference to `_global_impure_ptr'
collect2: error: ld returned 1 exit status
make: *** [common.mk:233: bin/monolith.elf] Error 1
```
Ah, it's because when building the template it doesn't link with libc.

liam-teale:
Could you share the artifact built by the azure pipeline? Then I can benchmark and test this PR

AndrewLuGit:
[template.zip](https://github.com/user-attachments/files/17285905/template.zip)

Rocky14683:
@SizzinSeal Can u also test the template in the depot I sent u earlier?

liam-teale:
Yes rocky, I can. I'll post results here in 1-2 days, i don't have regular access to hardware

djava:
> Ah, it's because when building the template it doesn't link with libc.

@AndrewLuGit for clarity, is this a problem, or not?

AndrewLuGit:
> @AndrewLuGit for clarity, is this a problem, or not?

No, the `libc.a` that Liam is using is different from the version that I was testing with.

liam-teale:
Changed to draft for now, worked with @ion098 to get PROS to build with `-mfloat-abi=hard` by patching libv5rts.a with a python script

liam-teale:
This PR now uses `-mfloat-abi=hard`

@Rocky14683 ready for another round of review

AndrewLuGit:
[kernel@4.0.4-pr703.6513.zip](https://github.com/user-attachments/files/17388532/kernel%404.0.4-pr703.6513.zip)

common.mk
AndrewLuGit, line 1: If possible, could you move the rules for patching the SDK from `common.mk` to `Makefile`? Ideally, `common.mk` includes only common stuff needed for both compiling the kernel and compiling user projects.

Rocky14683: I think it needs to be in Makefile because common.mk is the one we release with kernel. And the makefile is specifically for kernel so i think its fine(makefile_template is the one user got).

liam-teale: i think this might be possible, I'll take a look

AndrewLuGit:
[kernel@4.0.4-pr703.6516.zip](https://github.com/user-attachments/files/17388820/kernel%404.0.4-pr703.6516.zip)

common.mk
liam-teale: this has been done

Rocky14683:
[kernel@4.0.4-pr703.6517.zip](https://github.com/user-attachments/files/17389244/kernel%404.0.4-pr703.6517.zip)

liam-teale:
I'm having some trouble with the benchmarks. While I can confirm that this version of libm is most certainly faster than the current version used by PROS, I can't get good benchmarks comparing softfp and hard versions of the latest libm. I'll be working on this over the next couple days, but some help would be greatly appreciated 🙏

liam-teale:
What I can do is test this version of libm on vexide. Benchmarking is much easier in rust thanks to `std::black_box`

liam-teale:
OK, benchmarks have been completed, as well as hardware tests. Pending another round of reviews, i think this should be good to go

ABUCKY0:
Will the switch to an updated toolchain break compatibility with older PROS 3 projects?

AndrewLuGit (dismissed):
looks good :+1:

Rocky14683:
I will test a few things tonight.

ion098:
@SizzinSeal Did you test whether libraries compiled for earlier kernel versions will either work properly when linked to the new PROS kernel, or throw an error at compile/link time, rather than failing inexplicably at runtime?

AndrewLuGit:
@ion098 libraries compiled for the soft ABI will fail to link when hot/cold linking is enabled. It doesn't fail when hot/cold linking is disabled, though...

liam-teale:
> Will the switch to an updated toolchain break compatibility with older PROS 3 projects?

No

> Did you test whether libraries compiled for earlier kernel versions will either work properly

Yes, and they all break. They will not link due to ABI incompatibilities. However, recompiling for the latest version of PROS will fix it without any changes to the source code.

Rocky14683 (dismissed):
LGTM(maybe merge it in the next pros meeting)

liam-teale:
Accidentally pushed local changes, and reverted them. However, this means your review is stale @Rocky14683

Rocky14683 (approved):
LGTM AGAIN.

PR #704, opened 2024-10-12
Gcc13 toolchain

### LINKED ISSUE #705: ✨[FEATURE REQUEST] C++ Module Support
I would like to use the Modules feature from C++20 in my project [cppreference](https://en.cppreference.com/w/cpp/language/modules).

When I currently compile I get this error:
`warning: keyword 'export' not implemented, and will be ignored`
I have tried adding this compiler flag:
`-fmodules-ts`
But the flag is not recognized.

I am not sure if this is a bug or an unimplemented feature because I know support for modules took a while for compilers to add. Thank you for your hard work in maintaining pros and keeping it free and open source.

Seems to be related to #687

Rocky14683:
#### Summary:
Update gcc toolchain to version 13. And have module support.

#### Motivation:
#703 #705

#### Test Plan:

- [ ] test item


ion098:
Why GCC 13 and not GCC 14?

Rocky14683:
@ion098 GCC ARM 13.3.Rel1 is the newest ARM toolchain based on GCC 13.3.

Rocky14683:
This was merged with #703 .

PR #706, opened 2024-10-22
✨ Link Time Optimization (LTO)

### LINKED ISSUE #660: ✨[FEATURE REQUEST] LTO
**Overview**
[Linker Time Optimization (LTO)](https://gcc.gnu.org/onlinedocs/gccint/LTO-Overview.html) performs optimizations during linking, as the name suggests. This allows for more aggressive optimizations, especially around binary size, and especially in larger projects with many files.

**Is your feature request related to a problem? Please describe.**
The PROS cold package is huge. This makes wireless upload times very long when updating the cold package, and downloading new kernel versions slower, a big problem with slow internet. It also makes online storage a concern.

**OPTIONAL Describe any solutions you've considered**
GCC has a `-flto` flag to enable linker-time optimization

**Additional context**
This requires modification of the `common.mk` file, it's not related to the PROS API.

liam-teale:
#### Summary:
Use Link Time Optimization (LTO) when building templates. Closes #660 

#### Motivation:
Significantly reduced binary size. This means:
- less storage used on V5
- faster upload times
- faster template download times
- less RAM used on V5

this applies to both the PROS kernel, and all PROS templates

#### Implementation
If `IS_LIBRARY` is equal to `1`, then `-flto` is added to `GCCFLAGS`. This is done so LTO is not used for user projects, where LTO would have minimal effect yet build time is noticeably increased.

Since LTO can break when compiling code that does not follow strict aliasing rules, some code in PROS had to be modified. While one could just use `-fno-strict-aliasing` and `-Wno-lto-type-mismatch`, the benefits of LTO would be reduced, so instead these issues have been fixed.

#### Binary Size
(pending build by PROS)

#### Build times
(pending build by PROS)

#### Test Plan:
- [X] compiles
- [ ] user code compiles
- [X] linker does not use excessive resources (<50mb RAM)
- [ ] hot/cold linking works
- [ ] stack traces work
- [ ] tasks work
- [ ] linker scripts still work

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -3,7 +3,7 @@ DEVICE=VEX EDR V5
 
 MFLAGS=-mcpu=cortex-a9 -mfpu=neon-fp16 -mfloat-abi=softfp -Os -g
 CPPFLAGS=-D_POSIX_THREADS -D_UNIX98_THREAD_MUTEX_ATTRIBUTES -D_POSIX_TIMERS -D_POSIX_MONOTONIC_CLOCK
-GCCFLAGS=-ffunction-sections -fdata-sections -fdiagnostics-color -funwind-tables -flto -fno-fat-lto-objects
+GCCFLAGS=-ffunction-sections -fdata-sections -fdiagnostics-color -funwind-tables -fno-strict-aliasing -flto
 
 # Check if the llemu files in libvgl exist. If they do, define macros that the
 # llemu headers in the kernel repo can use to conditionally include the libvgl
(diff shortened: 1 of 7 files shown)


Rocky14683:
[download link](https://artprodcus3.artifacts.visualstudio.com/A268c8aad-3bb0-47d2-9a57-cf06a843d2e8/a421d041-2579-4bf8-a0aa-d7a0307dade3/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3B1cmR1ZS1hY20tc2lnYm90cy9wcm9qZWN0SWQvYTQyMWQwNDEtMjU3OS00YmY4LWEwYWEtZDdhMDMwN2RhZGUzL2J1aWxkSWQvNjUyMi9hcnRpZmFjdE5hbWUvdGVtcGxhdGU1/content?format=zip)

PR #707, opened 2024-10-23
✨ Added Motor Group Append/Erase Functions

Gracelu128:
#### Summary:
Added append function that takes in port number and erase function that takes in reference to motor group.

#### Motivation:
The append function previously required a reference to the motor group, and the erase function required a port number. This update allows the append function to accept a port number, and the erase function to use a reference to the motor group.

#### Test Plan:

- [x] Try appending port via port number to motor group
- [x] Try erasing motors in motor group via reference to motor group

### DIFF
--- a/include/pros/motor_group.hpp
+++ b/include/pros/motor_group.hpp
@@ -2357,6 +2357,15 @@ class MotorGroup : public virtual AbstractMotor {
 	 */
 	void append(AbstractMotor&);
 
+
+	/**
+	 * Appends all the motors on the port to this motor group
+	 *
+	 * \param port The port to append to the motor group
+	 *
+	 */
+	void append(std::int8_t port);
+
 	/**
 	 * Removes the all motors on the port (regardless of reversal) from the motor group
 	 *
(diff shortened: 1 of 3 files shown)


include/pros/motor_group.hpp
```diff
@@ -2357,6 +2357,15 @@ class MotorGroup : public virtual AbstractMotor {
...
 
+
+	/**
+	 * Appends all the motors on the port to this motor group
+	 *
+	 * \param port The port to append to the motor group
+	 *
+	 */
+	void append(std::int8_t port);
+
```
WillXuCodes, line 2368: Needs some doc updates on rejection conditions (such as if the same absolute value port is found in the motor group)

src/devices/vdml_motorgroup.cpp
```diff
@@ -676,5 +676,32 @@ void MotorGroup::erase_port(std::int8_t port) {
...
+    }
+}
+
+void MotorGroup::erase_port(AbstractMotor& other) {
+	(*this) -= other;
+}
+
+
 }  // namespace v5
 }  // namespace pros
```
WillXuCodes, line 707: Add newline before merge (not part of changes in this PR but missing in general)

WillXuCodes (changes requested):

PR #708, opened 2024-10-24
refactor: :zap: Reduce code binary size

ion098:
#### Summary:
This PR reduces the memory footprint of PROS projects; it does this by reducing code size through the usage of Thumb mode, as well as removing unneeded linker sections.

#### Motivation:
Reducing memory footprint means smaller upload times.

#### Test Plan:
- [x] Check it compiles
- [x] Check that the kernel runs fine


Rocky14683:
https://artprodcus3.artifacts.visualstudio.com/A268c8aad-3bb0-47d2-9a57-cf06a843d2e8/a421d041-2579-4bf8-a0aa-d7a0307dade3/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3B1cmR1ZS1hY20tc2lnYm90cy9wcm9qZWN0SWQvYTQyMWQwNDEtMjU3OS00YmY4LWEwYWEtZDdhMDMwN2RhZGUzL2J1aWxkSWQvNjUyNy9hcnRpZmFjdE5hbWUvdGVtcGxhdGU1/content?format=zip

djava:
Explanation for these changes please? I'm unclear on why you're making like, any of them.

firmware/v5-common.ld
```diff
@@ -228,7 +226,7 @@ _SDA2_BASE_ = __sdata2_start + ((__sbss2_end - __sdata2_start) / 2 );
    _heap_end = .;
    HeapLimit = .;
 } > HEAP
-
+/*
```
djava: Surely commenting out this section without an explanation is not the best way to achieve this?

src/system/unwind.c
```diff
@@ -114,6 +114,7 @@ _Unwind_Reason_Code trace_fn(_Unwind_Context* unwind_ctx, void* d) {
 /******************************************************************************/
 // recover registers from a data abort. Specific knowledge about callee stack
 // used from FreeRTOS_DataAbortHandler and it calling DataAbortInterrupt
+__attribute((target("arm")))
```
djava: Since when can you call `__attribute` without the trailing `__`?

common.mk
```diff
@@ -1,7 +1,7 @@
 ARCHTUPLE=arm-none-eabi-
 DEVICE=VEX EDR V5
 
-MFLAGS=-mcpu=cortex-a9 -mfpu=neon-fp16 -mfloat-abi=hard -Os -g
+MFLAGS=-mcpu=cortex-a9 -mfpu=neon-fp16 -mfloat-abi=hard -Os -g -mthumb-interwork -mthumb
```
liam-teale:
```suggestion
MFLAGS=-mcpu=cortex-a9 -mfpu=neon-fp16 -mfloat-abi=hard -Os -g -mthumb
```

GCC docs explicitly state that it is not necessary to use `-mthumb-interwork` ([source](https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html#index-mthumb-interwork))

> Generate code that supports calling between the ARM and Thumb instruction sets. Without this option, **on pre-v5 architectures**, the two instruction sets cannot be reliably used inside one program.

The cortex A9 uses the arm v7 architecture, so, according to these docs, there is no reason to use `-mthumb-interwork`. Not using `-mthumb-interwork` will decrease the binary size.

src/system/unwind.c
```diff
@@ -114,6 +114,7 @@ _Unwind_Reason_Code trace_fn(_Unwind_Context* unwind_ctx, void* d) {
 /******************************************************************************/
 // recover registers from a data abort. Specific knowledge about callee stack
 // used from FreeRTOS_DataAbortHandler and it calling DataAbortInterrupt
+__attribute__((target("arm")))
```
liam-teale, line 134: Probably best to document why this attribute is necessary

Rocky14683:
https://artprodcus3.artifacts.visualstudio.com/A268c8aad-3bb0-47d2-9a57-cf06a843d2e8/a421d041-2579-4bf8-a0aa-d7a0307dade3/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3B1cmR1ZS1hY20tc2lnYm90cy9wcm9qZWN0SWQvYTQyMWQwNDEtMjU3OS00YmY4LWEwYWEtZDdhMDMwN2RhZGUzL2J1aWxkSWQvNjU0My9hcnRpZmFjdE5hbWUvdGVtcGxhdGU1/content?format=zip

liam-teale:
tested, works

firmware/v5-common.ld
ion098: I've added an explanation in https://github.com/purduesigbots/pros/pull/708/commits/8b9cfae4726ca3191efb34155bf8d3d13538a26c.

src/system/unwind.c
ion098: Thanks for catching this, fixed in https://github.com/purduesigbots/pros/pull/708/commits/8b9cfae4726ca3191efb34155bf8d3d13538a26c.

common.mk
ion098: Not necessarily, the docs don't say that on post-v5 architectures it will always work without `-mthumb-interwork`, just that on pre-v5 architectures it will always fail. It depends on how the SDK was compiled, although I guess we could just test if removing the flag works.

src/system/unwind.c
ion098: Done in https://github.com/purduesigbots/pros/pull/708/commits/e2fcbd135d52d0829681ea946de8dd66e3739697.

common.mk
ion098: Trying this in https://github.com/purduesigbots/pros/pull/708/commits/ac69470a22436d2ed3b4d7ea576035fd902f3ee7.

Rocky14683:
[Download link](https://artprodcus3.artifacts.visualstudio.com/A268c8aad-3bb0-47d2-9a57-cf06a843d2e8/a421d041-2579-4bf8-a0aa-d7a0307dade3/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3B1cmR1ZS1hY20tc2lnYm90cy9wcm9qZWN0SWQvYTQyMWQwNDEtMjU3OS00YmY4LWEwYWEtZDdhMDMwN2RhZGUzL2J1aWxkSWQvNjU0NS9hcnRpZmFjdE5hbWUvdGVtcGxhdGU1/content?format=zip)

liam-teale:
this works, tests successful

liam-teale:
Here is the size difference of the cold package when using `-mthumb-interwork` and not:

with `-mthumb-interwork` 1,326,332 bytes
without `-mthumb-interwork` 1,326,332 bytes

They are exactly the same size. This indicates that `-mthumb-interwork` has no effect on the size of the cold package.
This is supported by the [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html):
> `-mthumb-interwork`
> Generate code that supports calling between the ARM and Thumb instruction sets. Without this option, on pre-v5 architectures, the two instruction sets cannot be reliably used inside one program. The default is -mno-thumb-interwork, since slightly larger code is generated when -mthumb-interwork is specified. In AAPCS configurations this option is meaningless.

There are 2 points in the text that support this:
> In AAPCS configurations this option is meaningless

PROS uses an AAPCS configuration

>Without this option, on pre-v5 architectures, the two instruction sets cannot be reliably used inside one program

The Cortex A9 in the brain uses the "arm-v7" architecture, which is not a "pre-v5" architecture, implying that this option is not needed.

Since it has been shown PROS works without issue without the `-mthumb-interwork` flag, and is supported by GCC documentation and testing, the `-mthumb-interwork` flag should not be added back.

Rocky14683:
[Download link](https://artprodcus3.artifacts.visualstudio.com/A268c8aad-3bb0-47d2-9a57-cf06a843d2e8/a421d041-2579-4bf8-a0aa-d7a0307dade3/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3B1cmR1ZS1hY20tc2lnYm90cy9wcm9qZWN0SWQvYTQyMWQwNDEtMjU3OS00YmY4LWEwYWEtZDdhMDMwN2RhZGUzL2J1aWxkSWQvNjU0Ny9hcnRpZmFjdE5hbWUvdGVtcGxhdGU1/content?format=zip)

liam-teale:
@ion098 is this ready to be merged?

ion098:
@SizzinSeal Yes, this is ready to be merged now.

liam-teale:
@Rocky14683 the download link is invalid

ion098:
@Rocky14683 would you be able to rerun the pipeline? The pipeline seemed to not download the toolchain properly for some reason.

AndrewLuGit:
This PR saves reduces the size of the cold package by 34,240 bytes when compared to what is currently on the develop-pros-4 branch. In terms of upload time, this saves about 2-3 seconds when uploading the cold package wirelessly.

AndrewLuGit (approved):

PR #711, opened 2024-10-26
:bug: use `pros` instead of `prosv5` in Makefile

liam-teale:
#### Summary:
Use `pros` instead of `prosv5` in the makefile

#### Motivation:
currently, if `make template` is run, it will fail because `prosv5` is not a valid command anymore.

#### Test Plan:
- [X] `make template` runs successfully


WillXuCodes (approved):
Get this in immediately. Hotfix release ASAP if possible

liam-teale:
@WillXuCodes this does not require a hotfix. This is only a problem when compiling the pros kernel, as its specific to its makefile. You can compile regular templates just fine

WillXuCodes:
Ah gotcha, did not look at the change as closely as I should have last night, thanks lol

PR #712, opened 2024-10-26
🚨 fix motorgroup compiler warnings

liam-teale:
#### Summary:
fix motorgroup compiler warnings

#### Motivation:
Even though these specific warnings are harmless, it clutters the compiler output and may confuse new PROS developers.

#### Test Plan:
- [X] no warnings

### DIFF
--- a/src/devices/vdml_motorgroup.cpp
+++ b/src/devices/vdml_motorgroup.cpp
@@ -562,7 +562,7 @@ std::int32_t MotorGroup::set_gearing(const motor_gearset_e_t gearset, const std:
 }
 std::int32_t MotorGroup::set_gearing(std::vector<motor_gearset_e_t> gearsets) const {
 	empty_MotorGroup_check(PROS_ERR);
-	for (size_t i = 0; i < gearsets.size(); i++) {
+	for (std::size_t i = 0; i < gearsets.size(); i++) {
 		this->set_gearing(gearsets[i], _ports[i]);
 	}
 	if (gearsets.size() != _ports.size()) {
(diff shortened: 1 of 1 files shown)


src/devices/vdml_motorgroup.cpp
```diff
@@ -561,7 +562,7 @@ std::int32_t MotorGroup::set_gearing(const motor_gearset_e_t gearset, const std:
 }
 std::int32_t MotorGroup::set_gearing(std::vector<motor_gearset_e_t> gearsets) const {
 	empty_MotorGroup_check(PROS_ERR);
-	for (int i = 0; i < gearsets.size(); i++) {
+	for (uint32_t i = 0; i < gearsets.size(); i++) {
```
ion098:
```suggestion
	for (size_t i = 0; i < gearsets.size(); i++) {
```
This is a bit pedantic but `size_t` is the proper type for indices; if you're going to change it, might as well make it the right type.

src/devices/vdml_motorgroup.cpp
```diff
@@ -572,7 +573,7 @@ std::int32_t MotorGroup::set_gearing(std::vector<motor_gearset_e_t> gearsets) co
 
 std::int32_t MotorGroup::set_gearing(std::vector<MotorGears> gearsets) const {
 	empty_MotorGroup_check(PROS_ERR);
-	for (int i = 0; i < gearsets.size(); i++) {
+	for (uint32_t i = 0; i < gearsets.size(); i++) {
```
ion098:
```suggestion
	for (size_t i = 0; i < gearsets.size(); i++) {
```

src/devices/vdml_motorgroup.cpp
djava: `std::size_t` for consistent style - looks like we're using `std::` prefixed integer types in this file (as per the function signature).

Should add `#include <cstddef>` in that case instead of `<cstdint>`, though seems like we should have `<cstdint>` too anyway..

AndrewLuGit (approved):

PR #714, opened 2024-10-26
✨ Link Time Optimization (LTO)

liam-teale:
#### Summary:
Use Link Time Optimization (LTO) when building templates. Closes #660 

#### Motivation:
Significantly reduced binary size. This means:
- less storage used on V5
- faster upload times
- faster template download times
- less RAM used on V5

this applies to both the PROS kernel, and all PROS templates

#### Implementation
If `IS_LIBRARY` is equal to `1`, then `-flto` is added to `GCCFLAGS`. This is done so LTO is not used for user projects, where LTO would have minimal effect yet build time is noticeably increased.

Since LTO can break when compiling code that does not follow strict aliasing rules, some code in PROS had to be modified. While one could just use `-fno-strict-aliasing` and `-Wno-lto-type-mismatch`, the benefits of LTO would be reduced, so instead these issues have been fixed.

#### Binary Size
(pending build by PROS)

#### Build times
(pending build by PROS)

#### Test Plan:
- [ ] compiles
- [ ] user code compiles
- [ ] linker does not use excessive resources (<50mb RAM)
- [ ] hot/cold linking works
- [ ] stack traces work
- [ ] tasks work
- [ ] linker scripts still work

PR #715, opened 2024-11-08
:construction: Use unity build

ion098:


liam-teale:
would this be faster than just passing `-j`to make?

ion098:
@SizzinSeal 
> would this be faster than just passing `-j`to make?

They do different things, the point of this PR is to increase optimization by combining all the source files into one, thus increasing optimization (similar to LTO).

djava:
I can no find no real sources online claiming that your given reason for this change is effective compared to LTO. Why?

ion098:
@djava 
> I can no find no real sources online claiming that your given reason for this change is effective compared to LTO. Why?

LTO would be the ideal solution, but there are a bunch of [linker errors](https://github.com/purduesigbots/pros/pull/714#issuecomment-2439775817) that look nontrivial to fix (although the errors here make unity builds also seem less trivial than I was hoping).

djava:
Yeah, if you're getting mismatches across linked files, then putting them all into one file won't make them less mismatched.

ion098:
@djava 
The errors here vs from LTO are very different and (most likely) stem from different root causes.

Issue #725, opened 2024-11-26
✨[FEATURE REQUEST] get_digital_new_release


miles352:
There's a method to get if a controller button was just pressed, I suggest a method to see if a button was just released.

edwrdq:
this should be closed considering the PRs been merged for 9 months

PR #726, opened 2024-11-28
Implemented get_digital_new_release

miles352:
#### Summary:
I added the function Controller::get_digital_new_release

#### Motivation:
I would like to be able to use this method in my pros project and I think it makes sense because there is already get_digital_new_press.

##### References (optional):
PR made for issue #725

### DIFF
--- a/include/pros/misc.h
+++ b/include/pros/misc.h
@@ -482,6 +482,45 @@ int32_t controller_get_digital(controller_id_e_t id, controller_digital_e_t butt
 int32_t controller_get_digital_new_press(controller_id_e_t id, controller_digital_e_t button);
 
 /**
+ * Returns a falling-edge case for a controller button press.
+ *
+ * This function is not thread-safe.
+ * Multiple tasks polling a single button may return different results under
+ * the same circumstances, so only one task should call this function for any
+ * given button. E.g., Task A calls this function for buttons 1 and 2.
+ * Task B may call this function for button 3, but should not for buttons
+ * 1 or 2. A typical use-case for this function is to call inside opcontrol
+ * to detect new button releases, and not in any other tasks.
+ *
+ * This function uses the following values of errno when an error state is
+ * reached:
+ * EACCES - Another resource is currently trying to access the controller
+ * port.
+ *
+ * \param button
+ * 			  The button to read. Must be one of
+ *        DIGITAL_{RIGHT,DOWN,LEFT,UP,A,B,Y,X,R1,R2,L1,L2}
+ *
+ * \return 1 if the button on the controller is not pressed and had been
+ * pressed the last time this function was called, 0 otherwise.
+ *
+ * \b Example
+ * \code
+ * void opcontrol() {
+ * 	pros::Controller master(pros::E_CONTROLLER_MASTER);
+ * 	while (true) {
+ *   if (master.get_digital_new_release(pros::E_CONTROLLER_DIGITAL_A)) {
+ *     // Toggle pneumatics or other similar actions
+ *   }
+ *
+ *   delay(2);
+ *   }
+ * }
+ * \endcode
+ */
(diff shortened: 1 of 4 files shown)


Rocky14683:
[download link](https://artprodcus3.artifacts.visualstudio.com/A268c8aad-3bb0-47d2-9a57-cf06a843d2e8/a421d041-2579-4bf8-a0aa-d7a0307dade3/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3B1cmR1ZS1hY20tc2lnYm90cy9wcm9qZWN0SWQvYTQyMWQwNDEtMjU3OS00YmY4LWEwYWEtZDdhMDMwN2RhZGUzL2J1aWxkSWQvNjYwMC9hcnRpZmFjdE5hbWUvdGVtcGxhdGU1/content?format=zip)

AndrewLuGit:
@miles352 I see you mirrored the logic in `get_digital_new_press`. The issue with this is that you cannot use both `get_digital_new_press` and `get_digital_new_release`, because if one of them calls `set_button_pressed`, it'll mess up the other one's call to `get_button_pressed`.
Example:
You call `get_digital_new_press`, and `get_digital_new_release` afterwards. When the button goes from being pressed to not being pressed, then `get_digital_new_press` will call `set_button_pressed(port, button_num, false)`. Then, in `get_digital_new_release`, `get_button_pressed(port, button_num)` will return false, and so  `get_digital_new_release` will return false, even though the button was just released.
Since you were the one who created the issue, if this behavior is okay with you, I'll go ahead and approve the PR. Otherwise, we'll need to implement some more complicated logic.

miles352:
@AndrewLuGit That's a good point, I didn't realize that before. I was going to use both in my project so that issue won't be okay. I don't fully understand how the internal button states are stored but it seems like it wouldn't make sense to create separate  button press storages for the two functions, so I'm not sure the best way to do this.

djava:
> it wouldn't make sense to create separate button press storages for the two functions

why not?

miles352:
> why not?

How would that work? I guess you would modify the controller_data struct, do you think something like this would work?
```cpp
typedef struct controller_data {
    bool button_pressed_for_pressed[NUM_BUTTONS];
    bool button_pressed_for_released[NUM_BUTTONS];
} controller_data_s_t;

miles352:
I went ahead and did it that way, it's probably not the best solution but I think the controller_data struct is only used in those two functions so it's probably okay for now.

djava:
I'm not inherently opposed to this, assuming it works? It's only like, 12 bytes of static memory used if thats what you're worried about. We could consider doing this as a bitfield to save memory I guess but whatever.

Rocky14683:
[download link](https://artprodcus3.artifacts.visualstudio.com/A268c8aad-3bb0-47d2-9a57-cf06a843d2e8/a421d041-2579-4bf8-a0aa-d7a0307dade3/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3B1cmR1ZS1hY20tc2lnYm90cy9wcm9qZWN0SWQvYTQyMWQwNDEtMjU3OS00YmY4LWEwYWEtZDdhMDMwN2RhZGUzL2J1aWxkSWQvNjYyMC9hcnRpZmFjdE5hbWUvdGVtcGxhdGU1/content?format=zip)

thiccaxe:
What is the status on this, has it been tested? Does it work?

It works, except for a small bug: new release will always return true as program begins. Is this intended behavior?

aghangurde:
@miles352 can you confirm that this is intended behavior:

new_release will be always be true on the first call to the function on program startup.

miles352:
@aghangurde No, this is not intended behavior.

aghangurde:
If you initialize the release array in the struct to true (or the equivalent by inverting all the logic) this issue will be fixed.

AndrewLuGit (approved):

PR #728, opened 2024-12-01
🐛 Fix display when building a monolith

liam-teale:
#### Summary:
Fix display when a project is built as a monolith

#### Motivation:
When you build a PROS project as a monolith, the display is unusable. This is because the weakly declared functions in `libpros.a` do not get overridden with the strongly declared functions in `liblvgl.a`, but only if building a monolith.

#### How does this fix work?
Weakly defined symbols are now only declared if LVGL is not present.
Except for one:
the `pros_initialize` function calls the weakly declared `display_initialize()` function. One solution would be to call this function after `pros_initialize`, in liblvgl.a. But alas, this won't work (see https://github.com/purduesigbots/pros/pull/144/#issuecomment-496901942). Instead, we initialize pros partially, then `display_initialize` is called as a constructor, then pros is initialized fully.

#### The underlying issue

The underlying issue is that weakly defined symbols do not get overridden, but only if building a monolith. Despite my best efforts, I was not able to identify the cause. It was not the order in which the libraries were linked, or the objects getting optimized out, since in my tests I eliminated both cases and the weakly declared functions still weren't overridden.

This PR does not fix this problem, it just treats the symptoms with a workaround.

#### Why does it work if using hot/cold linking?

Because all objects in every library are included in the cold package. When using a monolith, only the required objects are included. We could do the same when building a monolith, but this would cause the binary size to grow in size by an order of magnitude. Hence, this PR introduces a different workaround.

#### Incompatibilities

This PR requires a new version of liblvgl to be created. Current versions of liblvgl are not compatible with this version of pros. However, the new version of liblvgl will not be compatible with current versions of PROS. The new version of LVGL and the new version of PROS must be released simultaneously, and PROS would require, at minimum, a minor version increment (NOT a patch).

liblvgl pull request: https://github.com/purduesigbots/liblvgl/pull/51

See the related PR to liblvgl: 

#### Test Plan (pending on PROS team sharing a template zip):

This snippet works as expected when user code is built as
-[ ] hot/cold packages
-[ ] monolith

```c++
#include "main.h"

void initialize() {
    pros::lcd::initialize();
}
```

### DIFF
--- a/include/pros/llemu.h
+++ b/include/pros/llemu.h
@@ -15,7 +15,7 @@
 /******************************************************************************/
 #ifdef _PROS_INCLUDE_LIBLVGL_LLEMU_H
 #include "liblvgl/llemu.h"
-#endif
+#else
 
 #ifdef __cplusplus
 extern "C" {
@@ -50,5 +50,7 @@ bool __attribute__((weak)) lcd_print(int16_t line, const char* fmt, ...)  {
 } // namespace pros
 } // extern "C"
 #endif//__cplusplus
+
+#endif // _PROS_INCLUDE_LIBLVGL_LLEMU_H
 
 #endif // _PROS_LLEMU_H_
--- a/include/pros/llemu.h
+++ b/include/pros/llemu.h
@@ -16,7 +16,7 @@
 /******************************************************************************/
 #ifdef _PROS_INCLUDE_LIBLVGL_LLEMU_H
 #include "liblvgl/llemu.h"
-#else
+#else // prevent lcd_print being resolved to definition below when liblvgl is present
 
 #ifdef __cplusplus
 extern "C" {
--- a/include/pros/llemu.h
+++ b/include/pros/llemu.h
@@ -52,6 +52,6 @@ bool __attribute__((weak)) lcd_print(__attribute__((unused)) int16_t line, __att
 } // extern "C"
 #endif//__cplusplus
 
-#endif // _PROS_INCLUDE_LIBLVGL_LLEMU_H
+#endif // _PROS_INCLUDE_LIBLVGL_LLEMU_H not present
 
 #endif // _PROS_LLEMU_H_
--- a/src/system/startup.c
+++ b/src/system/startup.c
@@ -22,7 +22,6 @@ extern void rtos_initialize();
 extern void vfs_initialize();
 extern void system_daemon_initialize();
 extern void graphical_context_daemon_initialize(void);
-extern __attribute__((weak)) void display_initialize(void) {}
 extern void rtos_sched_start();
 extern void vdml_initialize();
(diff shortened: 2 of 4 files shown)


include/pros/llemu.h
```diff
@@ -15,7 +15,7 @@
 /******************************************************************************/
 #ifdef _PROS_INCLUDE_LIBLVGL_LLEMU_H
 #include "liblvgl/llemu.h"
-#endif
+#else
```
djava: add a comment explaining this fix

include/pros/llemu.h
```diff
@@ -51,4 +51,6 @@ bool __attribute__((weak)) lcd_print(int16_t line, const char* fmt, ...)  {
 } // extern "C"
 #endif//__cplusplus
 
+#endif // _PROS_INCLUDE_LIBLVGL_LLEMU_H
```
djava:
```suggestion
#endif // _PROS_INCLUDE_LIBLVGL_LLEMU_H not present
```

src/system/startup.c
```diff
@@ -33,21 +32,24 @@ extern void invoke_install_hot_table();
 // gives the compiler instructions on the priority of the constructor,
 // from 0-~65k. The first 0-100 priorities are reserved for language
 // implementation.
-__attribute__((constructor(101))) static void pros_init(void) {
+__attribute__((constructor(101))) static void pros_init_a(void) {
```
djava: maybe `_1` and `_2` suffix instead of `_a` and `_b`? I was initially wondering what `_a` stood for.

Rocky14683:
https://artprodcus3.artifacts.visualstudio.com/A268c8aad-3bb0-47d2-9a57-cf06a843d2e8/a421d041-2579-4bf8-a0aa-d7a0307dade3/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3B1cmR1ZS1hY20tc2lnYm90cy9wcm9qZWN0SWQvYTQyMWQwNDEtMjU3OS00YmY4LWEwYWEtZDdhMDMwN2RhZGUzL2J1aWxkSWQvNjYwMy9hcnRpZmFjdE5hbWUvdGVtcGxhdGU1/content?format=zip

src/system/startup.c
WillXuCodes: a more verbose `stage_1` and `stage_2` would do the trick too

liam-teale:
Suggestions implemented, merge conflicts fixed. Ready for another round of reviews

Rocky14683:
It's still giving me data abort

liam-teale:
are you using the right version of LVGL?

liam-teale:
Closing, I'm not happy with this PR. It should solve the underlying issue, not work around it.

Issue #731, opened 2024-12-20
🐛[BUG] - Cannot set pistons through adi expander


ssejrog:
#### Expected Behavior:

Setting pistons to extend/retract through the ADI expander with a `DigitalOut` or `Pneumatics` doesn't work. 

#### Actual Behavior:

The piston should move.

#### Steps to reproduce:

In the code below, L1 will not move the piston if it's an `adi::Pneuamatics` or `adi::DigitalOut`.  This code works if you plug the solenoid into the brain. 
```cpp
// pros::adi::Pneumatics claw({4, 'A'}, false);
pros::adi::DigitalOut claw({4, 'A'});
ez::Piston claw2('A');
void opcontrol() {
  while (true) {
    if (master.get_digital(DIGITAL_L1)) {
      claw.set_value(true);
      // claw.extend();
    } else {
      claw.set_value(false);
      // claw.retract();
    }

    if (master.get_digital(DIGITAL_R1)) {
      claw2.set(true);
    } else {
      claw2.set(false);
    }

    pros::delay(10);
  }
}
```

Without changing any hardware, I confirmed that this works as expected with VEXcode.
```cpp
int main() {
  // Initializing Robot Configuration. DO NOT REMOVE!
  vexcodeInit();
  while (true) {
    if (Controller1.ButtonR1.pressing()) {
      piston.set(true);
    } else {
      piston.set(false);
    }
    wait(10, msec);
  }
}
```

#### System information:
Platform: 
PROS Kernel Version: 4.1.1

PR #732, opened 2024-12-21
Fix digital out for adi expander

Rocky14683:
#### Summary:
Pneumatics on adi expander not working

##### References (optional):
#731 

#### Test Plan:

- [x] Toggle pneumatics that is connected through adi expander


ssejrog:
Tested and works 👀

AndrewLuGit (approved):
I trust Jess

PR #734, opened 2024-12-24
:sparkles: Allow calling user provided hooks in system daemon

ion098:
#### Summary:
Allows calling user provided functions inside of the PROS system daemon.

#### Motivation:
Allows user code to process sensor data right after `vexBackgroundProcessing` is called.

#### Test Plan:
- [ ] check it compiles
- [ ] check that hooks are added and called properly

### DIFF
--- a/include/pros/apix.h
+++ b/include/pros/apix.h
@@ -52,6 +52,9 @@ extern "C" {
 
 typedef void* queue_t;
 typedef void* sem_t;
+typedef void (*generic_fn_t)(void);
+
+void add_daemon_hook(generic_fn_t hook);
 
 /**
  * Unblocks a task in the Blocked state (e.g. waiting for a delay, on a
--- a/src/system/system_daemon.c
+++ b/src/system/system_daemon.c
@@ -12,6 +12,7 @@
  */
 
 #include "kapi.h"
+#include "common/linkedlist.h"
 #include "system/optimizers.h"
 #include "system/user_functions.h"
 #include "v5_api.h"
@@ -37,6 +38,9 @@ static void _competition_initialize_task(void* ign);
 static void _initialize_task(void* ign);
 static void _system_daemon_task(void* ign);
 
+static linked_list_s_t* hook_list = NULL;
+static mutex_t mut = 0;
+
 enum state_task { E_OPCONTROL_TASK = 0, E_AUTON_TASK, E_DISABLED_TASK, E_COMP_INIT_TASK };
 
 char task_names[4][32] = {"User Operator Control (PROS)", "User Autonomous (PROS)", "User Disabled (PROS)",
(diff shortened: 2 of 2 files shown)


djava:
Can we please get a more detailed motivation for this?

src/system/system_daemon.c
```diff
@@ -45,13 +49,27 @@ task_fn_t task_fns[4] = {_opcontrol_task, _autonomous_task, _disabled_task, _com
 
 extern void ser_output_flush(void);
 
+void add_daemon_hook(generic_fn_t hook) {
+	// TODO: make thread safe
+	if (unlikely(hook_list == NULL)) {
```
djava, line 54: Not that it matters because this whole function is like, the coldest of paths, but is this even unlikely? It happens at least once per program that uses this function, and it seems at least semi-likely that most programs that use this would only have one.

Not that it matters, but you did add the `unlikely` so it seems worth questioning.

src/system/system_daemon.c
```diff
@@ -45,13 +49,27 @@ task_fn_t task_fns[4] = {_opcontrol_task, _autonomous_task, _disabled_task, _com
...
 static inline void do_background_operations() {
 	port_mutex_take_all();
 	ser_output_flush();
 	rtos_suspend_all();
 	vexBackgroundProcessing();
 	rtos_resume_all();
+	linked_list_foreach(hook_list, &call_hook, NULL);
 	vdml_background_processing();
 	port_mutex_give_all();
 }
```
djava, line 75: I'm not sold on this order - I frankly don't know what vdml_background_processing is, but it sounds like something that should happen first.

Also, any user code should definitely not be called while all port mutexes are locked, because thats obviously going to deadlock, as every device API function requires locking a port mutex.

I think the `linked_list_foreach` should be at the end, after `port_mutex_give_all()`.

include/pros/apix.h
```diff
@@ -52,6 +52,9 @@ extern "C" {
 
 typedef void* queue_t;
 typedef void* sem_t;
+typedef void (*generic_fn_t)(void);
+
+void add_daemon_hook(generic_fn_t hook);
```
djava, line 57: Doxygen, obviously.

Also, I'm unsold on the names of both `generic_fn_t` and `add_daemon_hook`.

djava: Probably `data_collection_hook_fn_t` and `add_data_processing_hook()`?

liam-teale:
There is a much better way to achieve the desired behavior which does not make it possible to accidentally crash the PROS system daemon (credit to @meisZWFLZ for this idea).

The PROS system daemon will notify tasks in a global linked list after mutexes have been released. If these tasks have a high enough priority, they will be scheduled immediately after the PROS system daemon ticks.

PR #735, opened 2024-12-27
fix liblvgl with monolith

Rocky14683:
#### Summary:
fix liblvgl with monolith 

also delete make template overload(That's just a small bug)

#### Motivation:
#728 

#### Test Plan:

- [x] Hot/cold
- [x] monolith
- [ ] More tests, just in case I'm stupid

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ CREATE_TEMPLATE_ARGS+=--user "src/main.{cpp,c,cc}" --user "include/main.{hpp,h,h
 CREATE_TEMPLATE_ARGS+=--target v5
 CREATE_TEMPLATE_ARGS+=--output bin/monolith.bin --cold_output bin/cold.package.bin --hot_output bin/hot.package.bin --cold_addr 58720256 --hot_addr 125829120
 
-template:: patch_sdk_headers clean-template library
+template: patch_sdk_headers clean-template library
 	$(VV)mkdir -p $(TEMPLATE_DIR)
 	@echo "Moving template files to $(TEMPLATE_DIR)"
 	$Dif [ $(shell uname -s) == "Darwin" ]; then \
--- a/common.mk
+++ b/common.mk
@@ -43,8 +43,8 @@ else
     FILTERED_LIBRARIES=$(LIBRARIES)
 endif
 
-LNK_FLAGS=--gc-sections --start-group $(strip $(LIBRARIES)) -lgcc -lstdc++ --end-group -T$(FWDIR)/v5-common.ld
-LNK_FLAGS_MONOLITH=--gc-sections --start-group $(strip $(LVGL_LIB_FLAGS) $(FILTERED_LIBRARIES)) -lgcc -lstdc++ --end-group -T$(FWDIR)/v5-common.ld
+LNK_FLAGS=--gc-sections --start-group $(strip $(LIBRARIES)) -lgcc -lstdc++ --end-group -T$(FWDIR)/v5-common.ld --no-warn-rwx-segments
+LNK_FLAGS_MONOLITH=--gc-sections --start-group $(strip $(LVGL_LIB_FLAGS) $(FILTERED_LIBRARIES)) -lgcc -lstdc++ --end-group -T$(FWDIR)/v5-common.ld --no-warn-rwx-segments
 
 
 ASMFLAGS=$(MFLAGS) $(WARNFLAGS)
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -75,7 +75,7 @@ CREATE_TEMPLATE_ARGS+=--user "src/main.{cpp,c,cc}" --user "include/main.{hpp,h,h
 CREATE_TEMPLATE_ARGS+=--target v5
 CREATE_TEMPLATE_ARGS+=--output bin/monolith.bin --cold_output bin/cold.package.bin --hot_output bin/hot.package.bin --cold_addr 58720256 --hot_addr 125829120
 
-template:: patch_sdk_headers clean-template library
```
djava, line 78: are you sure this was a mistake? It has a semantically different meaning now.

common.mk
```diff
@@ -34,7 +34,18 @@ LIBRARIES+=$(wildcard $(FWDIR)/*.a)
...
+ifneq (,$(wildcard ./firmware/liblvgl.a))
+    LVGL_LIB_FLAGS=--whole-archive ./firmware/liblvgl.a --no-whole-archive
+    FILTERED_LIBRARIES=$(filter-out ./firmware/liblvgl.a, $(LIBRARIES))
+else
+    LVGL_LIB_FLAGS=
+    FILTERED_LIBRARIES=$(LIBRARIES)
+endif
+
 LNK_FLAGS=--gc-sections --start-group $(strip $(LIBRARIES)) -lgcc -lstdc++ --end-group -T$(FWDIR)/v5-common.ld --no-warn-rwx-segments
+LNK_FLAGS_MONOLITH=--gc-sections --start-group $(strip $(LVGL_LIB_FLAGS) $(FILTERED_LIBRARIES)) -lgcc -lstdc++ --end-group -T$(FWDIR)/v5-common.ld --no-warn-rwx-segments
```
djava: Can we figure out a better way to do this than duplication? Also should `LNK_FLAGS` be `LNK_FLAGS_HOTCOLD` or something?

djava (commented):
I'm confused about what is happening here, can you explain what these changes actually are?

And should this really go here instead of in liblvgl?

Makefile
Rocky14683: yeh, its just a kernel thing. If it overload it will duplicate a libpros.a in the project root directory.

Rocky14683:
Yes it should be in kernel because it's just linking issue.

Rocky14683:
I make it ```--whole-archive``` so that every symbol in liblvgl are included and are strong ref.

djava:
Could I have an explanation of what the linking issue is and how this solves it? And also probably set it as the top-level PR description,

Rocky14683:
The linking process didn't work initially because the linker was discarding unused symbols from liblvgl.a. Adding ```--whole-archive``` works because it forces the linker to include all symbols from liblvgl.a, even if the linker doesn't detect any direct references to those symbols.

Makefile
djava: So the `template` rule in common.mk is no longer used then, since it's being overridden in user project `Makefile`? Can it be removed, or is it used to build the kernel? If so, should it be in the kernel's `Makefile`?

djava:
Ok, can you add that explanation as a comment in the makefile?

Makefile
Rocky14683: It's only being used by users. There's no template rule in the distributed makefile, only in common.mk.

Rocky14683:
https://artprodcus3.artifacts.visualstudio.com/A268c8aad-3bb0-47d2-9a57-cf06a843d2e8/a421d041-2579-4bf8-a0aa-d7a0307dade3/_apis/artifact/cGlwZWxpbmVhcnRpZmFjdDovL3B1cmR1ZS1hY20tc2lnYm90cy9wcm9qZWN0SWQvYTQyMWQwNDEtMjU3OS00YmY4LWEwYWEtZDdhMDMwN2RhZGUzL2J1aWxkSWQvNjY2NS9hcnRpZmFjdE5hbWUvdGVtcGxhdGU1/content?format=zip

liam-teale:
So, if I'm reading this correctly, the entirety of liblvgl is put in the monolith? The binary size is going to balloon.

Additionally, what if one generic template overrides a weakly declared symbol in another template? Is that going to work? The changes in this PR is specific to liblvgl

Makefile
djava: Ok, so the kernel build uses this `template` rule here, and user project builds use the one in `common.mk`? Then shouldn't the one in `common.mk` be moved to `template-Makefile`? If it's not used by both, it shouldnt be in `common.mk`, right?

Rocky14683:
1. I didn't think about the binary size
2. It should still work

liam-teale:
PR #728 also fixes the problem, without modifications to `common.mk` which is specific to LVGL, and without causing the binary size to explode. What's the motivation for this PR?

PR #737, opened 2025-01-09
feat: :sparkles: Get the state of the controller's power button

ion098:
#### Summary:
Allows accessing the state of the power button on the controller.

#### Motivation:
It can be useful occasionally to be able to use the controller power button as an input.

#### Test Plan:

- [x] Check it compiles
- [x] Check that `controller_get_digital` works with the power button
- [x] Check that `controller_get_digital_new_press` works with the power button


thiccaxe:
#### Test Plan:

- [x] Check it compiles
- [x] Check that `controller_get_digital` works with the power button
- [x] Check that `controller_get_digital_new_press` works with the power button

AndrewLuGit (approved):
LGTM

Issue #736, opened 2024-12-30
🐛[BUG] - Negative port reversed motors do not work with is_installed. 


theguy20phi:
#### Expected Behavior:
The is_installed motor method should return true whenever it is reversed by constructing it with a negative port value. 

#### Actual Behavior:
It returns false. 

#### Steps to reproduce:
![motor_works_and_is_reversed_but_apparently_not_installed](https://github.com/user-attachments/assets/70bacf6d-3090-4e14-af28-d48790a273cd)

#### System information:
Platform: V5
PROS Kernel Version: 4.1.1

#### Additional Information
I think it has to do with the constructor called for the Device parent class taking an unsigned integer for its port whereas the Motor class takes a signed integer. I say this because whenever the motor is reversed by set_reversed it works as expected.
![also_works](https://github.com/user-attachments/assets/2e96cb19-82c3-47d2-9936-8fcb868a8931)

thiccaxe:
probably can be solved like this
```cpp
Motor::Motor(const std::int8_t port, const pros::v5::MotorGears gearset, const pros::v5::MotorUnits encoder_units)
    : Device(abs(port), DeviceType::motor), _port(port) {...}
```

PR #738, opened 2025-01-09
build: :technologist: Add more warning flags to kernel

ion098:
#### Summary:
Adds more warning flags to PROS kernel.

#### Motivation:
These warning flags would have caught issues like #736 at compile time.

#### Test Plan:
- [ ] Check that kernel compiles

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ EXCLUDE_SRCDIRS+=$(SRCDIR)/tests
 C_STANDARD=gnu2x
 CXX_STANDARD=gnu++23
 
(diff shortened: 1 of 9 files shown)


AndrewLuGit:
I don't think adding `-Werror` is a good idea.

ion098:
re @AndrewLuGit 

> I don't think adding -Werror is a good idea.

I would disagree with you, save for the fact that I don't feel like fixing 525 errors at the moment. That being said, some of these warnings should probably be fixed at a later time.

PR #741, opened 2025-01-15
fix: :bug: Fix is_installed when using negative port reversed motors

### LINKED ISSUE #736: 🐛[BUG] - Negative port reversed motors do not work with is_installed.

ion098:
#### Summary:
Fixed an issue where `is_installed` would not work with `Motor`s initialized with a negative port number.

#### Motivation:
Self explanatory.

##### References (optional):
Closes #736 

#### Test Plan:
- [x] Check that `is_installed` works with positive motor ports
- [x] Check that `is_installed` works with negative motor ports

### DIFF
--- a/src/devices/vdml_motors.cpp
+++ b/src/devices/vdml_motors.cpp
@@ -13,6 +13,7 @@
 #include "kapi.h"
 #include "pros/motors.hpp"
 #include "vdml/vdml.h"
+#include <cmath>
 
 namespace pros {
 inline namespace v5 {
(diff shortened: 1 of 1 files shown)


thiccaxe:
#### Test Plan:
- [x] Check that `is_installed` works with positive motor ports
- [x] Check that `is_installed` works with negative motor ports

AndrewLuGit (approved):
LGTM

PR #753, opened 2025-02-23
:truck: General Cleanup

liam-teale:
#### Summary:
The root directory was filled with single files. This has been fixed

#### Motivation:
It was very messy

#### Test Plan:
- [X] compiles


liam-teale:
oops wrong repo my bad

PR #754, opened 2025-02-24
📝 more accurate README

liam-teale:


liam-teale:
wrong repo again. I'll disconenct ZestCode from the fork network so this does not happen again

PR #759, opened 2025-03-19
add serial.h/.hpp to api.h

Rocky14683:
For some reason serial.h and serial.hpp was not included in api.h. 
Shout out to Renzhi for finding this mistake.


AndrewLuGit:
serial is included as part of `apix.h`, as typical users do not need to use serial.

PR #760, opened 2025-03-23
fix adi expander docs... lmao

Rocky14683:
#### Summary:
For some reason the adi docs shows `pros::adi::DigitalOut sensor ( {{ EXT_ADI_SMART_PORT , ADI_PORT }});` and that create a std::initializer_list of pairs, so uhh yep.


AndrewLuGit:
fixed in 4.1.2, will be merged with develop eventually

PR #762, opened 2025-03-27
fix compiler warnings for kernel

Rocky14683:
#### Summary:
During kernel compilation, there are a lot of warnings generated, most of which are intentional and do not require attention. To improve compilation speed and reduce distractions, these warnings are suppressed in this pr.

The warning for liblvgl not installed will still be shown, but it will only show once for kernel builds.

#### Motivation:
Those warnings are annoying.


AndrewLuGit (approved):
LGTM

PR #764, opened 2025-03-27
Ai vision 

Rocky14683:
#### Summary:
They force me to do documentations 😢 

#### Test Plan:

- [ ] hardware test


src/devices/vdml_ai_vision.cpp
```diff
@@ -80,8 +80,17 @@ AIVision::Object AIVision::get_object(uint32_t object_index) {
...
 
-uint32_t AIVision::get_class_name(int32_t id, uint8_t* class_name) {
-	return c::aivision_get_class_name(this->_port, id, class_name);
+uint32_t AIVision::get_class_name(int32_t id, char* class_name) {
+	return c::aivision_get_class_name(this->_port, id, reinterpret_cast<uint8_t*>(class_name));
+}
+
+std::optional<std::string> AIVision::get_class_name(int32_t id) {
+	char class_name[21];
+	uint32_t result = c::aivision_get_class_name(this->_port, id, reinterpret_cast<uint8_t*>(class_name));
```
ion098: `char` and `uint8_t` are the same types with the arm ABI, so there should be no need for `reinterpret_cast` here.

Rocky14683: In assembly uint8_t* is unsigned char*, int8_t is signed char*, and char* is just char*. Also theres no implicit conversion for pointer, I do need reinterpret_cast here. Oh and reinterpret_cast is done during compile-time.

include/pros/ai_vision.h
```diff
@@ -68,6 +68,7 @@ typedef enum aivision_mode_type {
 	E_AIVISION_MODE_TAGS = (1 << 0),
 	E_AIVISION_MODE_COLORS = (1 << 1),
 	E_AIVISION_MODE_OBJECTS = (1 << 2),
+	E_AIVISION_MODE_COLOR_MERGE = (1 << 4),
 	E_AIVISION_MODE_ALL = (1 << 0) | (1 << 1) | (1 << 2),
```
aghangurde, line 75: update to include 1<<4?

AndrewLuGit: color merge enables merging of adjacent color detections

src/devices/vdml_ai_vision.cpp
AndrewLuGit: changed it to a C-style cast instead

AndrewLuGit (approved):

PR #765, opened 2025-04-03
swap optical get/set integration time docs

AndrewLuGit:
#### Summary:

Swaps the docs so that they match the functions.
#### Motivation:

These were swapped for some reason


Rocky14683 (approved):
LGTM

PR #774, opened 2025-09-08
ci: :construction_worker: check that examples in header files compile

ion098:
#### Summary:
This PR adds a Python script to be ran during CI that isolates all the example code snippets in the `include/pros` directory and checks that they compile.
~~(Note: this script was partially written by Copilot)~~ The current code doesn't contain any part of the original code which contained code from Copilot
 
#### Motivation:
Prevents issues like #760.

#### Test Plan:
- [ ] CI passes

### DIFF
--- a/check_example_code.py
+++ b/check_example_code.py
@@ -1,139 +1,119 @@
+import asyncio
 import os
 import re
 import subprocess
 import sys
-import time
-import traceback
+import tempfile
 
 include_directory = "include/pros"
-is_error = False
+precompiled_include_directory = ""
+error_count = 0
+gcc_semaphore = asyncio.Semaphore(value=os.cpu_count())
 
 def print_and_exit(message):
-    print(f"Failed to check example code in pros headers:\n{message}")
+    print(f"Failed to check example code in pros headers:\n{message}", file=sys.stderr)
     sys.exit(1)
 
-max_subprocesses = os.cpu_count()
-current_subprocesses = 0
-subprocess_list = []
+def precompile_header():
+    flags = [
+        "-x", "c++-header",
+        "include/main.h",
+        "-std=c++23",
+        "-I", "include",
+        "-D", "_PROS_KERNEL_SUPPRESS_LLEMU_WARNING",
+        "-Wfatal-errors",
+        "-o", os.path.join(precompiled_include_directory, "main.h.gch")
+    ]
+    subprocess.run(["arm-none-eabi-gcc", *flags], check=True)
 
-def compile_code(code_text, filename, is_cpp):
+async def compile_code(code_text, filename, is_cpp):
     """
     Compiles the given code, discarding the output file.
     
     Args:
         code_text (str): The C/C++ source code to compile.
-        filename (str): The name of the file.
         is_cpp (bool): Whether the code is C++ or not.
+        
+    Returns:
+        success (bool): Whether the compilation succeeded or not.
     """
-    global max_subprocesses, current_subprocesses, subprocess_list
-    while current_subprocesses == max_subprocesses:
-        for i in range(len(subprocess_list)):
-            proc, (code_snippet, filename) = subprocess_list[i]
-            if isinstance(proc, subprocess.Popen):
-                retcode = proc.poll()
-                if retcode is not None:
-                    current_subprocesses -= 1
-                    success = (retcode == 0)
-                    if not success:
-                        global is_error
-                        is_error = True
-                    stdout, stderr = proc.stdout.read(), proc.stderr.read()
-                    subprocess_list[i] = (success, stdout, stderr), (code_snippet, filename)
-        time.sleep(1)
-    while len(subprocess_list) > 0 and not isinstance(subprocess_list[0][0], subprocess.Popen):
-        (success, stdout, stderr), (code_snippet, filename) = subprocess_list.pop(0)
-        if not success:
-            print(f"=== example code from {filename} failed to compile: ===")
-            print(code_snippet)
-            print("=== compiler output below: ===")
-            print(stderr)
     flags = [
-        "arm-none-eabi-gcc",
         "-x", "c++" if is_cpp else "c",
         "-std=c++23" if is_cpp else "-std=c2x",
+        # Only need precompiled headers for C++
+        *(("-I", precompiled_include_directory) if is_cpp else ()),
         "-I", "include",
+        # Prevent spurious warnings
         "-D", "_PROS_KERNEL_SUPPRESS_LLEMU_WARNING",
+        # Stop compilation on first error
         "-Wfatal-errors",
-        "-",                                         # Read input from stdin
-        "-S",                                        # Generate assembly (avoids linker errors)
-        "-o", os.devnull                             # Discard the output file
+        # Read input from stdin
+        "-",
+        # Generate assembly (avoids linker errors)
+        "-S",
+        # Discard the output file
+        "-o", os.devnull
     ]
-    process = subprocess.Popen(
-        flags,
-        stdin=subprocess.PIPE,
-        stdout=subprocess.PIPE,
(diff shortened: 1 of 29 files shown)


ion098:
@Rocky14683 Can you please run the pipeline?

ion098:
@Rocky14683 can you please rerun pipeline?

ion098:
@BennyBot can you please rerun the pipeline?

check_example_code.py
```diff
@@ -0,0 +1,139 @@
...
+#    process = subprocess.Popen(
+#        flags,
+#        stdin=subprocess.PIPE,
+#        stdout=subprocess.PIPE,
+#        stderr=subprocess.PIPE
+#    )
+#    stdout, stderr = process.communicate(input=code_text.encode("utf-8"))
+#    success = (process.returncode == 0)
+#
+#    return success, stdout.decode(), stderr.decode()
```
djava: Presumably intended for removal? Make sure you do

check_example_code.py
```diff
@@ -0,0 +1,139 @@
...
+
+def print_and_exit(message):
+    print(f"Failed to check example code in pros headers:\n{message}")
+    sys.exit(1)
+
+max_subprocesses = os.cpu_count()
+current_subprocesses = 0
+subprocess_list = []
+
+def compile_code(code_text, filename, is_cpp):
```
djava: I do not like this subprocess handling at all. be so honest rn did an LLM write this

djava: Surely python has a premade thread-pool type construct somewhere? Can't this be done with asyncio?

ion098: No, its human written; I'll switch to asyncio though --- that should make this a lot cleaner

ion098: Switched to using `asyncio` in https://github.com/purduesigbots/pros/pull/774/commits/2d9a547423024ed154159fdff2c8af448a240921.

check_example_code.py
ion098: Done in https://github.com/purduesigbots/pros/pull/774/commits/2d9a547423024ed154159fdff2c8af448a240921.

ion098:
@BennyBot can you please rerun the pipeline?

check_example_code.py
```diff
@@ -1,139 +1,119 @@
...
-max_subprocesses = os.cpu_count()
-current_subprocesses = 0
-subprocess_list = []
+def precompile_header():
+    flags = [
+        "-x", "c++-header",
+        "include/main.h",
+        "-std=c++23",
+        "-I", "include",
+        "-D", "_PROS_KERNEL_SUPPRESS_LLEMU_WARNING",
```
djava, line 23: huge nitpick (like, dont bother changing this) but for the record its UB to have any identifier (including macros) that start with underscore then a capital letter lol.

djava: actually idek if you wrote that macro or not but 🤷 just having fun being an asshole by mentioning it

check_example_code.py
```diff
@@ -1,139 +1,119 @@
...
+    subprocess.run(["arm-none-eabi-gcc", *flags], check=True)
 
-def compile_code(code_text, filename, is_cpp):
+async def compile_code(code_text, filename, is_cpp):
     """
     Compiles the given code, discarding the output file.
     
     Args:
         code_text (str): The C/C++ source code to compile.
-        filename (str): The name of the file.
```
djava: was this remove a mistake?

check_example_code.py
```diff
@@ -1,139 +1,119 @@
...
         "-D", "_PROS_KERNEL_SUPPRESS_LLEMU_WARNING",
+        # Stop compilation on first error
         "-Wfatal-errors",
-        "-",                                         # Read input from stdin
-        "-S",                                        # Generate assembly (avoids linker errors)
-        "-o", os.devnull                             # Discard the output file
+        # Read input from stdin
+        "-",
+        # Generate assembly (avoids linker errors)
+        "-S",
```
djava, line 53: does the more standard -c not work? if not then probably give a longer comment to explain why youre doing -S because it took me a second for that to click

check_example_code.py
```diff
@@ -1,139 +1,119 @@
...
+    async with gcc_semaphore:
+        process = await asyncio.create_subprocess_exec(
+            "arm-none-eabi-gcc",
+            *flags,
+            stdin=asyncio.subprocess.PIPE,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.PIPE
+        )
+        stdout, stderr = await process.communicate(input=code_text.encode("utf-8"))
+        success = (process.returncode == 0)
```
djava, line 66: interesting - where do you await the compilation being done? im not super familiar with this but im confused. does process.communicate() imply blocking to wait for a response? is there a timeout?

djava (commented):
this asyncio approach is much cleaner, nice job.

Last request besides what comments i left was just - type hints?

check_example_code.py
ion098: Yeah, that shouldnt have been removed

check_example_code.py
ion098: -c would probably work, but assembling the files seems like an unnecessary step.

check_example_code.py
ion098: Yes, communicate sends input via stdin and then waits for the process to finish running (no timeout though)

check_example_code.py
djava: Ok, maybe just a better comment - something like `Just generate assembly, the further compilation steps are not necessary for this CI check` maybe

check_example_code.py
djava: ok maybe a comment that says that because its not obvious imo

check_example_code.py
```diff
@@ -0,0 +1,119 @@
...
+        is_cpp (bool): Whether the code is C++ or not.
+        
+    Returns:
+        success (bool): Whether the compilation succeeded or not.
+    """
+    flags = [
+        "-x", "c++" if is_cpp else "c",
+        "-std=c++23" if is_cpp else "-std=c2x",
+        # Only need precompiled headers for C++
+        *(("-I", precompiled_include_directory) if is_cpp else ()),
```
djava, line 44: weird syntax trick lol maybe just do it after the list initialization

meisZWFLZ:
Some silly notes I had while skimming the code:
1. Perhaps you could use the compiler flags from the makefile? Likely by adding a make rule to extract the flags
  a. This could be applied to other configuration too, like the include directory
2. Idk how azure pipelines work, but the examples check can be done in parallel to the other checks
3. Maybe you're already doing this, but you could compile each snippet in parallel
4. A make rule could be added to run the script
5. Being able to specify a specific header to test would be useful for development

1 seems like it would be a lot of work, but everything else seems reasonable and justifiable

ion098:
re @meisZWFLZ 

> 1. Perhaps you could use the compiler flags from the makefile? Likely by adding a make rule to extract the flags
>    a. This could be applied to other configuration too, like the include directory

That's probably a good idea in the long run, I didn't implement it because Makefile scares me a little.

> 2. Idk how azure pipelines work, but the examples check can be done in parallel to the other checks
> 3. Maybe you're already doing this, but you could compile each snippet in parallel

The code compiles everything in parallel using as many cores as possible with asyncio, so I don't think doing both checks in parallel would offer much speedup.

> 4. A make rule could be added to run the script
> 5. Being able to specify a specific header to test would be useful for development

That seems reasonable, but the script needs a few changes for development use---caching the example code for example, since the script always recompiles every example code snippet right now, which works for ci but makes it painful to repeatedly run during development.

meisZWFLZ:
Do the other checks still run if the example check fails? I would think that would be ideal.

I kinda wonder whether this should be written in the makefile rather than a python script. Might make it a bit easier to implement caching with make rules.

Issue #783, opened 2026-01-24
BSS should not be used as the stack


lewisfm:
PROS uses `.bss` as its stack rather than having a dedicated section. As a result, variables in the last ~400 bytes of `.bss` will be corrupted.

Here is a repro:

```cpp
// Expected output:
// 00000000
// 00000000
// 00000000
// 00000000
// ...
// Actual output:
// 0002041371083224
// 2106731362106730
// 0240248001055
// 211293000310
// 000281361343135
// 2371283180176112328
// ...

#include "main.h"
#include <cstdio>

extern "C" {
__attribute__((section(".gnu.linkonce.b.end"))) uint8_t buffer[128];
}
void initialize() {
  for (int i = 0; i < 128; i += 8) {
    printf("%hhu%hhu%hhu%hhu%hhu%hhu%hhu%hhu\n", buffer[i], buffer[i + 1],
           buffer[i + 2], buffer[i + 3], buffer[i + 4],
           buffer[i + 5], buffer[i + 6], buffer[i + 7]);
  }
}
```

There's nothing special about ".gnu.linkonce.b.end" except that it is being mapped to nearby the end of `.bss` in the PROS linker script. If I hadn't included it, `buffer` might have been placed too far from the end of `.bss` to be corrupted.

```ld
/* Excerpt from the PROS linker script. */
.bss (NOLOAD) : {
   __bss_start = .;
   *(.bss)
   *(.bss.*)
   *(.gnu.linkonce.b.*)
   *(COMMON)
   __bss_end = .;
} > RAM
```

You should consider explicitly setting a stack in the linker script. Maybe the space in-between the cold package's data and the heap could be used.

## Why is this happening?

`vexMain`, called by `vexStartup` (the entry point), sets the stack pointer to `__stack`. PROS doesn't have a symbol called `__stack`, so it gets automatically created in the special `COMMON` section seen in the excerpt above. So, by the time `main` gets called, the stack is pointing (roughly) to the end of `.bss`.

lewisfm:
This bug is giving me some trouble porting [v5gdb](https://github.com/vexide/v5gdb) to PROS, since it expects globals in .bss to be zeroed.

Rocky14683:
#708 might be the issue

PR #785, opened 2026-02-04
Add a `.stack` segment in the linker.

speedskater1610:
#### Summary:
This PR adds a segment named `.stack` that reserves a memory space at  startup. This segment allocates a contiguous block of uninitialized memory in the RAM region, divided into dedicated stack areas. When a program that has a significant amount of globals BSS will overflow and cause UB for the program.  This happens because RTOS hasn't created a stack yet so the program just uses BSS as the stack. Having a dedicated stack for these will solve this issue. 

#### Motivation:

I was bored, and have to commonly help teammates debug and being able to use [v5gdb](https://github.com/vexide/v5gdb) would be helpful.

##### References (optional):

**This PR is related to #783**

#### Test Plan:

- [x] Test with #783's repro on [vexide's v5 qemu](https://github.com/vexide/vex-v5-qemu).
- [ ] Test #783's repro on a physical v5 brain.


firmware/v5-common.ld
```diff
@@ -209,7 +209,28 @@ _SDA_BASE_ = __sdata_start + ((__sbss_end - __sdata_start) / 2 );
...
+   . += _IRQ_STACK_SIZE;
+   __irq_stack = .;
+   
+   . += _ABORT_STACK_SIZE;
+   __abort_stack = .;
+   
+   . += _SUPERVISOR_STACK_SIZE;
+   __supervisor_stack = .;
+   
+   . += _STACK_SIZE;
```
djava: sorry am i missing something? where are all of these stack sizes defined? and where is the program told to use these stack pointers?

lewisfm: https://github.com/speedskater1610/pros/blob/bda8671233caecdea81b6ac83cdf068a3e8835b9/firmware/v5.ld#L4-L10

tropicaaal: Worth noting that the CPU mode stacks are probably a reminant of [Xilinx's linkerscript](https://github.com/Xilinx/embeddedsw/blob/1bb19ac1ab06ab322ba4340bed372f93ca612a18/scripts/linker_files/lscript_a9.ld.in#L282). The PROS linkerscripts are based on the VEXCode ones, which are in turn based on the zc702 board linkerscript provided by Xilinx. Normally, these stacks would be loaded in the reset vector for each processor mode with a banked `sp`, but this is already done in `vexSystemBoot` by VEXos, although it probably doesn't hurt to keep these around.

tropicaaal:
I strongly feel that the `*(COMMON)` fallthrough at bss should be removed. PROS should not be automatically creating unresolved symbols at `.bss` with the hope of failing silently. Any `extern`-ed symbols that the linker fails to resolve should cause a hard linker error.

speedskater1610:
I took into consideration all of the feed back from the review and comment and added a commit that removed `*(COMMON)` and added a comment before the `.stack` section to help clarify why a `.stack` section is needed.

lewisfm:
__undef_stack, __fiq_stack, etc are not actually used as far as I know. I believe __stack is the only required one, since that's what the partner sdk sets on initialization.

djava:
If we're removing all the references to `__undef_stack, __fiq_stack`, etc, might as well remove _UNDEF_STACK_SIZE, _FIQ_STACK_SIZE, etc as well?

speedskater1610:
Nice catch, I removed the unused definitions.

speedskater1610:
I can test #783 repro on a physical V5 brain on monday.

Rocky14683:
Thanks @speedskater1610

firmware/v5.ld
```diff
@@ -5,9 +5,6 @@ _STACK_SIZE = DEFINED(_STACK_SIZE) ? _STACK_SIZE : 0x2000;
 
 _ABORT_STACK_SIZE = DEFINED(_ABORT_STACK_SIZE) ? _ABORT_STACK_SIZE : 1024;
 _SUPERVISOR_STACK_SIZE = DEFINED(_SUPERVISOR_STACK_SIZE) ? _SUPERVISOR_STACK_SIZE : 2048;
-_IRQ_STACK_SIZE = DEFINED(_IRQ_STACK_SIZE) ? _IRQ_STACK_SIZE : 1024;
-_FIQ_STACK_SIZE = DEFINED(_FIQ_STACK_SIZE) ? _FIQ_STACK_SIZE : 1024;
-_UNDEF_STACK_SIZE = DEFINED(_UNDEF_STACK_SIZE) ? _UNDEF_STACK_SIZE : 1024;
 
 _HEAP_SIZE = DEFINED(_HEAP_SIZE) ? _HEAP_SIZE : 0x02E00000;  /* ~48 MB */
```
djava, line 6: why remove some but not all of the unused stack sizes?

speedskater1610: Sorry I thought I did, I will right now.

speedskater1610:
Sorry to make everyone wait, but I didnt have time to test the repro but I definatly will on wendesday. once again very sorry for the delay.

speedskater1610:
Sorry for the delay guys, but I tested on a physical brain, today. #783s repro now works as expected

KdotDevelopment (approved):

